PR #8, opened 2016-07-28
Added CMake support for SVF

rboggild:
As requested, here is a PR for including CMake support for SVF for LLVM 3.8.


rockysui:
Hi Rasmus,

It seems that your patch does not support out-of-tree project cmake. Could you please help us make one based on your patch? It would be good that you could make some notes about the steps and command line options for the cmake.

Thanks

rboggild:
Hi Yulei,

sorry about that - our LLVM build is somewhat modified so we just have SVF in the projects folder - I'll look into it and see how to support out-of-source builds as well.

Best regards,

Rasmus

rboggild:
Hi Yulei,

I have added support for building SVF out of source (it is still possible to build in source by placing the SVF dir inside the projects/ dir).

All you have to do to build SVF with CMake now is:

1) Build LLVM 3.8 with Cmake  (e.g. cmake /path/to/llvm/source && make)
2) Set the LLVM_DIR environment variable to point at the build directory for LLVM (e.g. export LLVM_DIR=/path/to/llvm/build )
3) Build SVF with Cmake (e.g. cmake /path/to/svf && make)

This will produce both the wpa and saber binaries (placed in the /bin folder) and the shared libraries so that other projects can include the SVF pass.

I added an explicit check and warning if the LLVM_DIR is not set as the default message might not make a sense, if a user forgets to add the environmental variable.

Best,

Rasmus

rockysui:
Hi Rasmus,

Awesome! thanks a lot for the out-of-tree patch:)

Before I merge your pull request, could you please make the following small changes to your patch.

(1) Please separate the _cmake config files_ and _CUDD file deletions_ into two patches.
(2) There are lots of pointer casting warnings issued from CUDD package during compilation. Could you please add "-Wno-pointer-to-int-cast" and "-Wno-pointer-to-int-cast" flags in the CUDD/CMakeLists.txt to disable those warnings?

Best,
Yulei

rboggild:
Hi Yulei,

I added the flags to suppress the warnings of CUDD and I think I was able to update the git history to contain the commits we need:

1) The "cudd_tests" branch (other pull request) now contains only the move commands for the cudd files from the lib folder to the tests folder. This is required for the cmake files, because the LLVM cmake system requires all files to be included in the cmake files and we can only have one .c file with a main method.

2) The master branch contains the merged commits (including the "cudd_test" and the "cmake" branches). If you apply the other patch first, I think we should get the desired result.

Hope it works :-)

/Rasmus

rockysui:
Hi Rasmus,

Perfect! 

Thanks for your contribution.

Best,
Yulei

dtzWill:
Thank you both! :+1:

Issue #17, opened 2017-04-17
Error on wpa -ander -svfg -dump-svfg example.bc 


nix7965:
Hi, I am trying to understand and check SVF capabilities.

However, I found some error with a toy code.
I also attached error messages.
---------------------------  Code     ------------------------------------------------
#include <iostream>
#include <string>

using namespace std;

class Profile
{
    public:
        void printProfile()
        {
            cout << "Name : " << _name.c_str() << endl;
            cout << "Phone Number : " << _phoneNumber.c_str() <<endl;
        }
        void setName(string name)
        {
            _name = name;
        }
        void setPhoneNumber(string phoneNumber)
        {
            _phoneNumber = phoneNumber;
        }
    private:
        string _name;
        string _phoneNumber;
};

int main()
{
    Profile myProfile;
    myProfile.setName("Hong");
    myProfile.setPhoneNumber("012319562");
    myProfile.printProfile();
    return 0;
}
------------------------ Error message --------------------------------------

Writing 'ander_svfg.dot'...#0 0x0000000000e7870b llvm::sys::PrintStackTrace(llvm::raw_ostream&) (~/SVF/build/bin/wpa+0xe7870b)
#1 0x0000000000e78a20 PrintStackTraceSignalHandler(void*) (~/SVF/build/bin/wpa+0xe78a20)
#2 0x0000000000e7706d llvm::sys::RunSignalHandlers() (~/SVF/build/bin/wpa+0xe7706d)
#3 0x0000000000e78181 SignalHandler(int) (~/SVF/build/bin/wpa+0xe78181)
#4 0x00007fe76b13b330 __restore_rt (/lib/x86_64-linux-gnu/libpthread.so.0+0x10330)
#5 0x000000000045ed78 MRVer::getSSAVersion() const (~/SVF/build/bin/wpa+0x45ed78)
#6 0x0000000000463d4f llvm::DOTGraphTraits<SVFG*>::getCompleteNodeLabel(SVFGNode*, SVFG*) (~/SVF/build/bin/wpa+0x463d4f)
#7 0x0000000000462e08 llvm::DOTGraphTraits<SVFG*>::getNodeLabel(SVFGNode*, SVFG*) (~/SVF/build/bin/wpa+0x462e08)
#8 0x000000000047bec9 llvm::GraphWriter<SVFG*>::writeNode(SVFGNode*) (~/SVF/build/bin/wpa+0x47bec9)
[...]

PR #18, opened 2017-04-18
LLVM 4

jcarlson23:
This should be relatively easy to integrate - although I would put this into a separate (a staging or local) branch and then merge back into the GitHub master from there.  The WPAPass no longer inheriting from AliasAnalysis is probably the place I would suggest to be looked at the most.  The "SVF Alterations.rtf" gives a summary of the changes, etc...

Let me know what else I can do to help.  I've gotta move on to a few other things the rest of the week but next week I'm going to start outline a few pivots I'm hoping to expand on :-)

### DIFF
--- a/include/WPA/WPASolver.h
+++ b/include/WPA/WPASolver.h
@@ -6,7 +6,7 @@
 // Copyright (C) <2013-2016>  <Jingling Xue>
 
 // This program is free software: you can redistribute it and/or modify
-// it under the terms of the GNU General Public License as published by
+// it under the terms of the GNU Gg12eneral Public License as published by
 // the Free Software Foundation, either version 3 of the License, or
 // (at your option) any later version.
 
@@ -126,7 +126,7 @@ protected:
         child_iterator EE = GTraits::direct_child_end(*v);
         for (; EI != EE; ++EI) {
             if (propFromSrcToDst(*(EI.getCurrent())))
-                pushIntoWorklist(Node_Index(EI));
+                pushIntoWorklist(Node_Index(*EI));
         }
     }
     /// Propagate information from source to destination node, to be implemented in the child class
@@ -162,8 +162,8 @@ protected:
     }
 
     /// Get node ID
-    inline NodeID Node_Index(GNODE* node) {
-        return GTraits::getNodeID(*node);
+    inline NodeID Node_Index(GNODE node) {
+        return GTraits::getNodeID(node);
     }
 
 private:
--- a/include/WPA/WPASolver.h
+++ b/include/WPA/WPASolver.h
@@ -43,7 +43,7 @@ class WPASolver {
 public:
     ///Define the GTraits and node iterator for printing
     typedef llvm::GraphTraits<GraphType> GTraits;
-    typedef typename GTraits::NodeType          GNODE;
+    typedef typename GTraits::NodeRef           GNODE;
     typedef typename GTraits::EdgeType          GEDGE;
     typedef typename GTraits::ChildIteratorType child_iterator;
 
--- a/include/WPA/WPASolver.h
+++ b/include/WPA/WPASolver.h
@@ -122,11 +122,11 @@ protected:
 
     /// Propagation for the solving, to be implemented in the child class
     virtual void propagate(GNODE* v) {
-        child_iterator EI = GTraits::direct_child_begin(v);
-        child_iterator EE = GTraits::direct_child_end(v);
+        child_iterator EI = GTraits::direct_child_begin(*v);
+        child_iterator EE = GTraits::direct_child_end(*v);
         for (; EI != EE; ++EI) {
             if (propFromSrcToDst(*(EI.getCurrent())))
-                pushIntoWorklist(Node_Index(*EI));
+                pushIntoWorklist(Node_Index(EI));
         }
     }
     /// Propagate information from source to destination node, to be implemented in the child class
@@ -163,7 +163,7 @@ protected:
 
     /// Get node ID
     inline NodeID Node_Index(GNODE* node) {
-        return GTraits::getNodeID(node);
+        return GTraits::getNodeID(*node);
     }
 
 private:
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -50,7 +50,8 @@ set(SOURCES
 
 add_llvm_loadable_module(Svf ${SOURCES})
 add_llvm_Library(LLVMSvf ${SOURCES})
-
+llvm_map_components_to_libnames(llvm_libs bitwriter core ipo irreader instcombine instrumentation target linker analysis scalaropts support )
+target_link_libraries(LLVMSvf ${llvm_libs})
 
 if(DEFINED IN_SOURCE_BUILD)
     add_dependencies(Svf intrinsics_gen)
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -50,8 +50,13 @@ set(SOURCES
 
 add_llvm_loadable_module(Svf ${SOURCES})
 add_llvm_Library(LLVMSvf ${SOURCES})
+
+# It seems that the Svf library needs to explictly require the CUDD alloc/init/etc so LLVMCudd
+link_directories( ${CMAKE_BINARY_DIR}/lib/Cudd )
+
 llvm_map_components_to_libnames(llvm_libs bitwriter core ipo irreader instcombine instrumentation target linker analysis scalaropts support )
 target_link_libraries(LLVMSvf ${llvm_libs})
+target_link_libraries(Svf LLVMCudd ${llvm_libs})
 
 if(DEFINED IN_SOURCE_BUILD)
     add_dependencies(Svf intrinsics_gen)
--- a/lib/MemoryModel/MemModel.cpp
+++ b/lib/MemoryModel/MemModel.cpp
@@ -189,7 +189,7 @@ bool SymbolTableInfo::computeGepOffset(const llvm::User *V, LocationSet& ls) {
         // Handling array types, skipe array handling here
         // We treat whole array as one, then we can distinguish different field of an array of struct
         // e.g. s[1].f1 is differet from s[0].f2
-        if(isa<ArrayType>(*gi))
+        if(isa<ArrayType>(gi.getIndexedType()))
             continue;
 
 
@@ -209,7 +209,7 @@ bool SymbolTableInfo::computeGepOffset(const llvm::User *V, LocationSet& ls) {
         // These GEP instructions are simply making address computations from the base pointer address
         // e.g. idx1 = (char*) &MyVar + 4,  at this case gep only one offset index (idx)
 
-        if (isa<PointerType>(*gi)) {
+        if (isa<PointerType>(gi.getIndexedType())) {
             // If this is a pointer, we're likely accessing an array through this pointer.
             // idx gives the array index of which element is being accessed. But since this
             // is a field-index based memory model, we consider array as containing one
@@ -222,7 +222,7 @@ bool SymbolTableInfo::computeGepOffset(const llvm::User *V, LocationSet& ls) {
 
         // Handling struct here
 
-        if (const StructType *ST = dyn_cast<StructType>(*gi)) {
+        if (const StructType *ST = gi.getStructTypeOrNull()) {
             assert(op && "non-const struct index in GEP");
             const vector<u32_t> &so = SymbolTableInfo::Symbolnfo()->getStructOffsetVec(ST);
             if ((unsigned)idx >= so.size()) {
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 Release/
 Debug/
 build/
+build-master/
 html/
 Release+Asserts/
 Debug+Asserts/
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,6 +31,7 @@ endif()
 include_directories(${CMAKE_CURRENT_SOURCE_DIR}/include
                     ${CMAKE_CURRENT_BINARY_DIR}/include)
 
+
 add_subdirectory(lib)
 add_subdirectory(tools)
 
--- a/SVF Alterations.txt
+++ b/SVF Alterations.txt
@@ -1,0 +1,24 @@
+Diff between “llvm-4” and “master”
+
+Likely the most important difference is the requirement that a NodeRef (rather than NodeType) be declared for types that will utilize the GraphTraits as part of the class definition.
+
+I altered most of the getPassName() instances to return a llvm::StringRef rather than a const char * type.  This seems rather trivial and I didn’t see a reason not to use the preferred string type choice for LLVM.
+
+/include/Util/DataFlow.h
+The PostDominatorTree becomes PostDominatorTreeWrapperPass, which is pretty consistent in recent revisions where a pass is accessed via a wrapper.
+
+I altered the base class /include/Util/GraphUtil.h
+
+To have a typedef typename Traits::NodeRef NodeRef, per llvm 4.  I believe this is necessary, as I looked over the llvm-4.  This with the NodeRef alterations in each of the specific graph classes required some alterations in terms of dereferencing, etc.  Again, a quick pass and it seemed to work but this is a place where a simple “*” might have a subtle unintended consequence.
+
+include/WPA/WPAPass.h
+
+Here I had to remove the explicit inheritance from “AliasAnalysis”, and this likely needs a fix, as AliasAnalysis needs a TargetLibraryInfo to be passed as the argument now and I’m a little unsure as to where to get that from.  There is no more default constructor…  To get the TargetLibraryInfo it seems the best convention is:
+
+TLInfo = &getAnalysis<TargetLibraryInfoWrapperPass>().getTLI();
+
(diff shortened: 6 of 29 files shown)


rockysui:
Jared,

Thanks for your help!

I am pretty busy these few days, and will start looking into your pull request from next week :)

nix7965:
Hi, Jared.

In overall, I think your code works well.
However, one thing that I found is that 
When I execute "wpa -ander -svfg -dump-svfg c++.opt", it couldn't recognize "-svfg" which is the valid option. (https://github.com/unsw-corg/SVF/wiki/Analyze-a-Simple-C-Program)

wpa: Unknown command line argument '-svfg'.  Try: 'wpa -help'
wpa: Did you mean '-ci-svfg'?

Since I am not sure whether who will fix it, I leave my comment here. 

PS. nix7965 referenced this pull request 5 days ago -> Did I make a mistake something? Why that appears...

rockysui:
I believe you need to apply Jared's patch on the most recent version of SVF. Try "git pull" before merging his patch.

jcarlson23:
Let me know if you need anything.  I don't recall changing an argument for wpa, but it might need a change with the AliasAnalysis change (requiring TargetLibraryInfo in any constructor call).

nix7965:
From my experience, I applied "only" Jared's patch to the recent SVF. Then, it works. 

Also, there is some minor issue.  I don't know why but SVF with @jcarlson23 's patch can be compiled with "cmake -DCMAKE_POSITION_INDEPENDENT_CODE:BOOL=true".

include/WPA/WPASolver.h
```diff
@@ -6,7 +6,7 @@
 // Copyright (C) <2013-2016>  <Jingling Xue>
 
 // This program is free software: you can redistribute it and/or modify
-// it under the terms of the GNU General Public License as published by
+// it under the terms of the GNU Gg12eneral Public License as published by
```
dtzWill: looks like this snuck in!

dtzWill:
Merging this with current SVF master yields: https://github.com/dtzWill/SVF/commit/ee5aeada2f458222ec0857308ca259201e2ed47c

Which seems to work pretty well overall, haven't stressed it too much yet :).

I'll note that #17 happens 100% of the time for me with these changes, please see my comment on that issue for a suggest workaround (or a discussion of why a workaround is not needed and I'm doing something wrong! ;)).  Either way, probably worth being aware of.

Thanks for the LLVM 4 port and of course thanks for SVF! :+1:

include/WPA/WPASolver.h
jcarlson23: oh wow, sorry, I can update that if need be.. obviously just a typo... :(

dtzWill:
Haha yeah no problem at all, just thought I'd point it out so it can be
fixed.

jcarlson23:
I think #17 probably happens because the fix in master was after when I started the port to llvm-4, so I think that'll be solved with a proper merge...    At least I hope so, but I'm not surprised by it, fwiw...

dtzWill:
@jcarlson23

That was my initial guess as well but AFAICT the issue happens on latest master (without your changes) but happens to not crash.  As I mentioned on #17 running under valgrind shows it's doing the same invalid memory accesses but apparently they're not a problem for whatever reason.

We'll see what the SVF folks say though, and regardless it'd be good to have some of this confirmed by others :).

rockysui:
Hopefully, I will start testing Jared's patch from tomorrow:)

lib/CMakeLists.txt
```diff
@@ -51,6 +51,12 @@ set(SOURCES
 add_llvm_loadable_module(Svf ${SOURCES})
 add_llvm_Library(LLVMSvf ${SOURCES})
 
+# It seems that the Svf library needs to explictly require the CUDD alloc/init/etc so LLVMCudd
```
jcarlson23: Sure, this ought to be straightforward I think.  I just moved computers but let me create a VM to test in and put in a architecture check and I think this can be resolved... (I'll try to update tonight).

lib/MemoryModel/MemModel.cpp
```diff
@@ -189,7 +189,7 @@ bool SymbolTableInfo::computeGepOffset(const llvm::User *V, LocationSet& ls) {
         // Handling array types, skipe array handling here
         // We treat whole array as one, then we can distinguish different field of an array of struct
         // e.g. s[1].f1 is differet from s[0].f2
-        if(isa<ArrayType>(*gi))
+        if(isa<ArrayType>(gi.getIndexedType()))
```
jcarlson23: Hmm...  A bit trickier than I had hoped :(

So just to make sure I understand the issue is the Operand's type?  So in your code example 
`(*gi)->dump()` is equivalent to `gi.getIndexedType()->dump()` from 3.8 to 4.0, but in the iterator the `getOperand()` in 4.0 is now pointing to the container's type rather than directly to the operand's type?

From what I saw in the code, I agree that changing the memory model would be very low in my preferences... Let me dig in a bit later tonight...  See if we can't find a better solution.

rockysui: (*gi)->dump() are not equivalent to gi.getIndexedType()->dump() in 4.0 now.

Yes, it needs to be the operand's type rather than its container's

lib/CMakeLists.txt
rockysui: Thanks!

lib/MemoryModel/MemModel.cpp
dtzWill: Hmm, darn.  For what it's worth, here's a bit of discussion about the commit that caused this functionality change:

https://reviews.llvm.org/D26594

Maybe the information there will help understand the changes and how to accommodate them properly.

I've run into this while migrating another project to 4.0 but haven't had time to sort out the changes sorry :(.

lib/CMakeLists.txt
jcarlson23: I believe I have a fix but have been in VM purgatory today so but basically just used CMAKE_SYSTEM_NAME and a little if-else..   so just will test again, today got away from me but this is the easy part...

lib/MemoryModel/MemModel.cpp
jcarlson23: Oh cool, thanks @dtzWill !  I was offline for a while today so I didn't get a chance to look at the commit/discussion log.  

 I'll dig in a bit more here..  If I get stuck I'll ask around or post to the mailing list in a day or so...

lib/CMakeLists.txt
jcarlson23: Alright, this built for me on Ubuntu16.04 LT with the following changes:

`diff --git a/lib/CMakeLists.txt b/lib/CMakeLists.txt
index 3c42075..fe7891d 100644
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -51,12 +51,18 @@ set(SOURCES
 add_llvm_loadable_module(Svf ${SOURCES})
 add_llvm_Library(LLVMSvf ${SOURCES})
 
-# It seems that the Svf library needs to explictly require the CUDD alloc/init/etc so LLVMCudd
-link_directories( ${CMAKE_BINARY_DIR}/lib/Cudd )
+if ( CMAKE_SYSTEM_NAME MATCHES "Darwin")
+   # It seems that the Svf library needs to explictly require the CUDD alloc/init/etc so LLVMCudd
+   link_directories( ${CMAKE_BINARY_DIR}/lib/Cudd )
+   llvm_map_components_to_libnames(llvm_libs bitwriter core ipo irreader instcombine instrumentation target linker analysis scalaropts support )
+   target_link_libraries(LLVMSvf ${llvm_libs})
+   target_link_libraries(Svf LLVMCudd ${llvm_libs})
+else()
+   llvm_map_components_to_libnames(llvm_libs bitwriter core ipo irreader instcombine instrumentation target linker analysis scalaropts support )
+   target_link_libraries(LLVMSvf ${llvm_libs})
+   target_link_libraries(Svf ${llvm_libs})
+endif()
 
-llvm_map_components_to_libnames(llvm_libs bitwriter core ipo irreader instcombine instrumentation target linker analysis scalaropts support )
-target_link_libraries(LLVMSvf ${llvm_libs})
-target_link_libraries(Svf LLVMCudd ${llvm_libs})
 
 if(DEFINED IN_SOURCE_BUILD)
     add_dependencies(Svf intrinsics_gen)`

I can reissue the pull request if that's easier but this is easy, I'm going to focus on the iterator issue below.

lib/MemoryModel/MemModel.cpp
jcarlson23: Alright, sat down and I think I have a strategy to bridge this.  I just want to try this on a few small scale experiments considering the discussion @dtzWill pointed out as hopefully this bridge will provide functionality but can be removed with some ease at a later date...

jcarlson23: Afternoon (well, it's afternoon here :) ).  So I created a small test case for a bridging iterator that I hope solves this problem.  I created it separately both to allow SVF to move forward with 4.0 compatibility but also since @dtzWill mentioned that he had a project with a similar issue, I tried to make it somewhat generic.  

@rockysui Would you mind looking at:
https://github.com/jcarlson23/bridgedGEPiterator

And seeing if that helps out in the test cases you brought up?  @dtzWill Obviously, if you need changes for that other project feel free to let me know of alter as you see fit.

Let me know how it goes...

rockysui: Hi Jared,

Thanks for your efforts in writing a bridgeGepIterator.

I found an issue that the iterator seems not overloading "operator++"?

Could you please try the following case
```
struct MyStruct {
        int * f1; 
        int * f2; 
};

int main() {
        struct MyStruct s[2];
        int a,b;
        s[0].f1 = &a; 
        s[1].f1 = &b; 

        // Different fields of different elements in a 
        // certain array are treated as different objects.
        NOALIAS(s[a].f1, s[b].f2);

        return 0;
}
```

After you compile the above one into bc with "-mem2reg" turned on, please take a look at the following value, which corresponds "s[a]" in the source code  
```
%arrayidx3 = getelementptr inbounds [2 x %struct.MyStruct], [2 x %struct.MyStruct]* %s, i64 0, i64 %idxprom, !dbg !132
```
Ideally, the gepIterator should return two types after the two loop iterations:
[2 x %struct.MyStruct]* 
[2 x %struct.MyStruct]

However, your iterator returns the same type during two loop iterations
[2 x %struct.MyStruct]* 

Could you please double check it?

Thanks

rockysui: A follow-up on the bridgedGepIterator.
I had another look at the test case in your link "https://github.com/jcarlson23/bridgedGEPiterator"

I found the root cause. I believe the type of "*gi" (llvm-3.8.0) is one level higher than that of the iterator "gi.getIndexedType()" (llvm-4.0). 

Please see below.

$ ./bin/testgep ../simple.bc 
  %1 = alloca %struct.MyStruct*, align 8
  %2 = load %struct.MyStruct*, %struct.MyStruct** %1, align 8
  %3 = getelementptr inbounds %struct.MyStruct, %struct.MyStruct* %2, i32 0, i32 0

Use the 4.0 IndexedType()
%struct.MyStruct = type { i32, i8* }
i32 0
i32
i32 0

**Current bridgedGepIterator implementation:**
%struct.MyStruct*
i32 0
%struct.MyStruct*
i32 0


**The ideal implementation (same as that of llvm-3.8.0) should be:**
%struct.MyStruct*
i32 0
%struct.MyStruct
i32 0

Thanks

dtzWill: Couldn't sleep, so poking at this a bit.

Yes, in LLVM 4 the gep_type_iterator no longer exposes the "outer" type,
[see here for a bit of explanation](https://github.com/llvm-mirror/llvm/blob/release_40/include/llvm/IR/GetElementPtrTypeIterator.h#L90).

I've only looked at MemModel.cpp (haven't tried the LocMemModel, that looks more difficult) but at least as far as the "fi_tests" from PTAben are concerned the following change seems to "fix" things:

```diff
diff --git a/lib/MemoryModel/MemModel.cpp b/lib/MemoryModel/MemModel.cpp
index 9bd65f9..01c0d17 100644
--- a/lib/MemoryModel/MemModel.cpp
+++ b/lib/MemoryModel/MemModel.cpp
@@ -193,7 +193,7 @@ bool SymbolTableInfo::computeGepOffset(const llvm::User *V, LocationSet& ls) {
         // Handling array types, skipe array handling here
         // We treat whole array as one, then we can distinguish different field of an array of struct
         // e.g. s[1].f1 is differet from s[0].f2
-        if(isa<ArrayType>(gi.getIndexedType()))
+        if(gi.isBoundedSequential())
             continue;
```
Thought I'd share, haven't worked on this enough to endorse this but might help regardless :).
For example, using `isSequential()` instead of `isBoundedSequential()` did *not* work properly on the tests, so more investigation is required I think.

Basically looking at the changes made in the commit that changed `gep_type_iterator`s behavior show how the various LLVM components were updated to accommodate this change (including BasicAA).

BridgedGEPIterator looks great, haven't had a chance to look at it yet, was trying to understand the change in behavior first...

dtzWill: (And agreed it'd be best/nice if this could be solved without changing memmodel...)

jcarlson23: Ah, thanks @rockysui I just made a quick update to the bridgedGEPIterator repo, and looks better but I'll try to verify in more detail later this afternoon. But a preliminary look seems good:

```
[+] Compare with our bridged implementation
%struct.MyStruct*
i32 0
%struct.MyStruct = type { i32, i8* }
i32 1
```

I'll also try to look at the suggested change @dtzWill (and it seems like a good fit at first glance to me), then we might have two alternative solutions...

jcarlson23: Ok, I did a little more testing using fix from this morning in https://github.com/jcarlson23/bridgedGEPiterator

And now that seems to conform to the 3.8 behavior pretty well now.  

That said, using API, as @dtzWill did, is clearly the direction things are moving towards so that might be preferable...  If it were up to me, I think that's what I would do.

I'll leave it up to others...  

I noticed the PTABen reference and am going to start trying to adapt my setup towards using that (I just hadn't seen it before) and might play with some alterations in the test scripts, just having a very difficult time running them at all (this is after altering the setup.sh, still seeing permissions errors, etc).

dtzWill: @jcarlson23 just to double-check, I think you need to 'source' the setup.sh script (`. ./setup.sh`) in order for the exported variables to persist in your current shell.

That said I'm not actually using `setup.sh` and instead just set a variety of variables directly.  Here's my Nix expression for running PTABen on a particular SVF: http://dtz.so/desac.nix

(If you're interested I can share the Nix expressions I use for SVF as well... but even if you don't know Nix it's pretty readable what's going on there I think :))

If the bridged GEP iterator /could/ be used and trusted to be compatible with previous behavior, I think that's the best solution until someone explicitly sits down and reworks some of the memory model code to use the new-style interface (which may require more extensive changes to the way it models the code, although hopefully not).

Unfortunately with something like LLVM there are often too many changes to make "diff"-based comparisons on large applications practical.  Although maybe we could get some mileage out of (ab)using the IR upgrading:

* Build large applications as LLVM 3.8 bitcode
* Run SVF 3.8 on it, dumping results to file (or something)
* Run SVF 4.0 on /exactly/ the same bitcode file and dumping analysis results similarly

That might be effective for minimizing the countless changes that would change the analysis results due to changes in clang, optimizations, etc..

Personally I'm a fan of checking analysis results dynamically--while certainly insufficient for proving correctness I've yet to find an LLVM-based pointer analysis (beyond the traditionally simple ones in LLVM) that was able to drive callgraph-checking instrumentation that didn't fail at runtime :(.

jcarlson23: @dtzWill Yeah, thanks, I did that and still had some serious issues...  Some of it, I think, is just the mess of Xcode and multiple llvm builds.  Also I think some of the expected directories coming out of the build I had renamed... eh.. it's a new machine, I kind of dumped some of the old apps/tools in and added some new and haven't quite had time to sort the cruft, but I'll get there.

I'm curious what you mean regarding your last paragraph.  Are you referring to building a CFG based model to dynamically check pointers?  Something akin to CFI (Control Flow Integrity) but with more analytical depth?  Do you care how the source might be altered/appended to, something like Seahorn or Pagai?  Anyway, just genuinely curious as to what you're after there...

As this is starting to get off tangent, perhaps email is best, but whatever works...  Thanks again.

jcarlson23: I should have said, I'll give the NIX stuff a crack as that help me clean up what I've got going, thanks for sharing!

rockysui: Jared,

Alternatively, I've also attached my scripts to build and setup SVF (llvm-4.0). Hopefully, it will be helpful for your to test bridgedGepIterator against PTABen.

[buid.sh](https://github.com/unsw-corg/SVF/files/977877/buid.txt)
[setup.sh](https://github.com/unsw-corg/SVF/files/977878/setup.txt)

You PR is behind the master repo. For a clean merge, could you please give a new PR after passing all the micro-benchmarks using your bridgedGepIteraor?

jcarlson23: @rockysui BridgedGepIterator it is!  I'm going to close this PR out then and work locally for a few days and get you a new one.  I'll aim for Monday, using your scripts or the NIX setup from Will to right my testing setup...

rockysui: Thanks, Jared!

PR #22, opened 2017-05-04
BddCondManager: fix deallocation to use Cudd_Quit.

dtzWill:
Fixes crash near exit when using SABER.


rockysui:
Thanks for the fix!

PR #40, opened 2018-03-27
Make external API names/types sourced from a text file

mbarbar:
This introduces a new option: "read-extapi" to choose the location of the database file. The old DB was moved to lib/Util/extAPIDB.txt, and in it, the format is specified.


yuleisui:
Mohamad,

Could you please validate your PR by comparing its points-to results with that of the current version using cpu2000/2006?

lib/Util/ExtAPI.cpp
```diff
@@ -69,17 +72,25 @@ static void fillExtAPIStringMap() {
...
+
+    if (ReadExtApi.empty() && ptahome != NULL) {
+        ReadExtApi = std::string(ptahome) + "/" + DB_NAME;
     }
 
     ifstream db(ReadExtApi);
     if (!db.is_open()) {
-        cout << "could not open external API database!\n";
+        if (ptahome == NULL && ReadExtApi.empty()) {
+            cout << "Neither $" << PTAHOME
```
yuleisui: Can we use cerr to make sure we will print them out if either one of the two errors happens?

yuleisui: Please go ahead and merge it once the above is fixed.

PR #41, opened 2018-04-03
Fix C++ Analysis Error due to LLVM-6.0 changes of Attribute argument No.

yuleisui:
An error occurs due to the Argument number is changed in LLVM-6.0. 
```
cs.paramHasAttr(0, ...) 
```
"0" is the first argument now.
Previously, in LLVM-4.0, "1" is the first argument.

This needs to be confirmed if someone knows.


jackanth:
I can't find a review, but it looks like it's this commit:
https://github.com/llvm-mirror/llvm/commit/1f8f0490690b06270ff949d7da3d10a6921606f4#diff-1268f9473fb9732219d83a3855026bbd

yuleisui:
Thanks Jack.

It was a quite subtle change...

Luckily, we have the assertions to catch it.

PR #71, opened 2018-07-03
Fixing fetching getSourceLocOfFunction

Machiry:
The loop is spurious, we use `getSubProgram` directly from the `Function` ptr `F` to get the `DISubprogram`. The loop makes the same information appended multiple times making it unreadable.


yuleisui:
Thanks for the fix.

PR #74, opened 2018-07-06
Adding support for vector operands (Patch 2)

Machiry:
Added changes to `PAGBuilder` to correctly handle store instructions, in case of vector operands.

### DIFF
--- a/lib/MemoryModel/PAGBuilder.cpp
+++ b/lib/MemoryModel/PAGBuilder.cpp
@@ -414,6 +414,27 @@ void PAGBuilder::visitStoreInst(StoreInst &inst) {
         NodeID src = getValueNode(inst.getValueOperand());
 
         pag->addStoreEdge(src, dst);
+    } else {
+        // if this is a vector operand, process the individual operands.
+        if(isa<VectorType>(inst.getValueOperand()->getType()) && dyn_cast<ConstantVector>(inst.getValueOperand())) {
+            ConstantVector *CVExpr = dyn_cast<ConstantVector>(inst.getValueOperand());
+            bool dstCreated = false;
+            NodeID vdst;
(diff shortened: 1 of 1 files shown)


lib/MemoryModel/PAGBuilder.cpp
```diff
@@ -414,6 +414,27 @@ void PAGBuilder::visitStoreInst(StoreInst &inst) {
...
+            bool dstCreated = false;
+            NodeID vdst;
+            // iterate for each element in the vector
+            for(unsigned vindex = 0; vindex < CVExpr->getNumOperands(); ++vindex) {
+                if(CVExpr->getOperand(vindex)->getType()->isPointerTy()) {
+                    NodeID vsrc = getValueNode(CVExpr->getOperand(vindex));
+                    // create dst node only if its required. i.e., if any of the element
+                    // of the vector's value is a pointer type.
+                    if(!dstCreated) {
+                        vdst = getValueNode(inst.getPointerOperand());
```
yuleisui: Shall we also consider whether the `PointerOperand` is a ConstantVector or not?

yuleisui:
Machiry,

I had a look at your patch. The vector instruction is a special one to SVF when building constraints and graphs. SVF is not able to handle ConstantVector and a few Vector Instructions (e.g., https://llvm.org/docs/LangRef.html#vector-operations) at this stage.

1) Could we possibly break the ConstantVector into multiple instructions before our analysis, similar to what BreakConstantExpr does? This will be cleaner for backend analysis. Let me know your thoughts.

2) Apart from your handling of `store`, we also need to handle `load` instructions.

3) It seems to me you didn't consider the case that PointerOperand is also a ConstantVector (see code comments)

PR #85, opened 2018-08-10
Adding Intraedge sanity (Part 2)

Machiry:
Sanity function to verify that nodes of an Intraedge always refer to the same function.


yuleisui:
Thanks

PR #100, opened 2018-10-18
fixing use-after-free in moveOutEdgesToRepNode

davidtr1037:


dtzWill:
On Thu, 18 Oct 2018 15:44:07 -0700, Yulei Sui <notifications@github.com> wrote:
> Merged #100 into master.
> 
> -- 
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly or view it on GitHub:
> https://github.com/SVF-tools/SVF/pull/100#event-1913528889Non-text part: text/html

FWIW I've been using this patch (at end, [1]) to address this problem for some time.
(
I really thought I sent it upstream to you folks but apologies if not!
Perhaps the doubt expressed in the commit message/comments is why :innocent:.
)

Anyway I'm sharing now because it does the same fix in two places,
and while I don't remember anything of the context it may be worth
a quick look to be sure the other location of this pattern shouldn't
be fixed similarly?

Hope this helps, thanks for SVF!

[1] https://github.com/dtzWill/nur-packages/blob/6035493e453f6986e9f506439e2b64f9dedc2466/pkgs/svf/fix-consg-edge-del.patch


From 15d12fe980ba0b27054ddffd1e5fca11f1bf26fd Mon Sep 17 00:00:00 2001
From: Will Dietz <w@wdtz.org>
Date: Thu, 4 May 2017 16:11:41 -0500
Subject: [PATCH] ConsG: Delete edge after checking its properties.

As noted in comments, unsure if the check makes sense
before removing it but it crashes as-is so
here's hoping the check can be done before.
---
 lib/MemoryModel/ConsG.cpp | 8 ++++++--
 1 file changed, 6 insertions(+), 2 deletions(-)

diff --git a/lib/MemoryModel/ConsG.cpp b/lib/MemoryModel/ConsG.cpp
index a88d675..467c8fd 100644
--- a/lib/MemoryModel/ConsG.cpp
+++ b/lib/MemoryModel/ConsG.cpp
@@ -400,12 +400,14 @@ bool ConstraintGraph::moveInEdgesToRepNode(ConstraintNode* node, ConstraintNode*
         if(isa<CopyCGEdge>(edge))
             removeDirectEdge(edge);
         else if (isa<GepCGEdge>(edge)) {
-            removeDirectEdge(edge);
             // If the GEP is critical (i.e. may have a non-zero offset),
             // then it brings impact on field-sensitivity.
             if (!isZeroOffsettedGepCGEdge(edge)) {
                 criticalGepInsideSCC = true;
             }
+            // XXX: Not sure if above check makes sense before removing edge,
+            // however we can't check /after/ deleting it so here's hoping.
+            removeDirectEdge(edge);
         }
         else if(isa<LoadCGEdge>(edge) || isa<StoreCGEdge>(edge))
             reTargetDstOfEdge(edge,rep);
@@ -451,12 +453,14 @@ bool ConstraintGraph::moveOutEdgesToRepNode(ConstraintNode*node, ConstraintNode*
         if(isa<CopyCGEdge>(edge))
             removeDirectEdge(edge);
         else if (isa<GepCGEdge>(edge)) {
-            removeDirectEdge(edge);
             // If the GEP is critical (i.e. may have a non-zero offset),
             // then it brings impact on field-sensitivity.
             if (!isZeroOffsettedGepCGEdge(edge)) {
                 criticalGepInsideSCC = true;
             }
+            // XXX: Not sure if above check makes sense before removing edge,
+            // however we can't check /after/ deleting it so here's hoping.
+            removeDirectEdge(edge);
         }
         else if(isa<LoadCGEdge>(edge) || isa<StoreCGEdge>(edge))
             reTargetSrcOfEdge(edge,rep);
-- 
2.12.2

dtzWill:
Oh, I just saw that this PR fixed the one but not the other.  Looks like they're both fixed, hooray! Sorry for the noise :3.

PR #106, opened 2018-10-29
Add svf.main()

sheisc:
svf.main() is used to model the real ENTRY point of a C++ program,
  which initializes all global C++ objects and then call main().

### DIFF
--- a/lib/MemoryModel/MemModel.cpp
+++ b/lib/MemoryModel/MemModel.cpp
@@ -904,10 +904,12 @@ void SymbolTableInfo::handleCE(const Value *val) {
             collectVal(ce);
             collectVal(ce->getOperand(0));
             collectVal(ce->getOperand(1));
+            collectVal(ce->getOperand(2));
             // handle the recursive constant express case
             // like (gep (bitcast (gep X 1)) 1); the inner gep is ce->getOperand(0)
             handleCE(ce->getOperand(0));
             handleCE(ce->getOperand(1));
+            handleCE(ce->getOperand(2));
         }
         // remember to handle the constant bit cast opnd after stripping casts off
         else {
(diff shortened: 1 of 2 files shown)


lib/Util/SVFModule.cpp
```diff
@@ -36,11 +36,32 @@
...
+
+  svf.main() is used to model the real entry point of a C++ program,
+  which initializes all global C++ objects and then call main().
+
+      define void @svf.main() {
+        entry:
+          call void @_GLOBAL__sub_I_cast.cpp()
+          call void @_GLOBAL__sub_I_1.cpp()
+          call void @_GLOBAL__sub_I_2.cpp()
+          call void bitcast (i32 (i32, i8**)* @main to void ()*)()
```
yuleisui: It seems to me you are using main(int argc, char * argv[] , char * env[]) in addSVFMain method, not consistent?

sheisc: It can support the following functions now.
(1) int main()
(2) int main(int argc, char * argv[])
(3) int main(int argc, char * argv[] , char env[])

I forgot to update the comment in Line 50.
Bitcast is not needed any more.

lib/Util/SVFModule.cpp
```diff
@@ -36,11 +36,34 @@
...
+#define SVF_MAIN_FUNC_NAME           "svf.main"
+#define SVF_GLOBAL_SUB_I_XXX          "_GLOBAL__sub_I_"
+
 using namespace std;
 using namespace llvm;
 
 static cl::opt<std::string> Graphtxt("graphtxt", cl::value_desc("filename"),
                                      cl::desc("graph txt file to build PAG"));
+                                     
+static cl::opt<bool> SVFMain("svfmain", cl::init(false),
```
yuleisui: Shall we turn it on by default? please run few tests to make sure it is correct.

lib/Util/SVFModule.cpp
```diff
@@ -79,6 +102,9 @@ LLVMModuleSet::LLVMModuleSet(const vector<string> &moduleNameVec) {
 
 void LLVMModuleSet::build(const vector<string> &moduleNameVec) {
     loadModules(moduleNameVec);
+    if(SVFMain){
```
yuleisui: put the if branch inside addSVFmain?

PR #107, opened 2018-10-31
Avoid executable "registered more than once" error

HongxuChen:
SVF static/dyanmic libraries (e.g., LLVMSvf.a/Svf.so in Linux) no longer
link LLVM libraries; this requires SVF library users to manually add
LLVM library dependencies. For performance consideration, we only allow
LLVMSvf.a to be used by static library or executable since LLVMSvf.a is
built without -fPIC; shared libraries has to use Svf.so.


lib/CMakeLists.txt
```diff
@@ -55,12 +55,14 @@ add_llvm_loadable_module(Svf ${SOURCES})
...
 link_directories( ${CMAKE_BINARY_DIR}/lib/Cudd )
-llvm_map_components_to_libnames(llvm_libs bitwriter core ipo irreader instcombine instrumentation target linker analysis scalaropts support )
-target_link_libraries(LLVMSvf ${llvm_libs})
+
+## we don't link libs for static library since we rely on target linkage
+# llvm_map_components_to_libnames(llvm_libs bitwriter core ipo irreader instcombine instrumentation target linker analysis scalaropts support )
+# target_link_libraries(LLVMSvf ${llvm_libs})
+
+## we don't link shared library (module) since it will be resolved by opt
 if ( CMAKE_SYSTEM_NAME MATCHES "Darwin")
```
yuleisui: Did you handle both OSX and ubuntu system?

tools/WPA/wpa.cpp
```diff
@@ -59,7 +61,7 @@ int main(int argc, char ** argv) {
 
     SVFModule svfModule(moduleNameVec);
 
-    WPAPass *wpa = new WPAPass();
+    auto wpa = std::unique_ptr<WPAPass>(new WPAPass());
```
yuleisui: I would like to avoid using unique_ptr :) It is quite different from the overall coding style of SVF.

Could you please keep it for your own use and revert it back?

HongxuChen: No problem, will revert. What's SVF's style for this? I saw multiple occurrences of raw new?

lib/CMakeLists.txt
HongxuChen: Only Ubuntu 18.04. I've no OSX at hand right now.

lib/CMakeLists.txt
```diff
@@ -55,13 +55,18 @@ add_llvm_loadable_module(Svf ${SOURCES})
...
+## need further inspection on this
 if ( CMAKE_SYSTEM_NAME MATCHES "Darwin")
+    llvm_map_components_to_libnames(llvm_libs bitwriter core ipo irreader instcombine instrumentation target linker analysis scalaropts support )
+    target_link_libraries(LLVMSvf ${llvm_libs})
     target_link_libraries(Svf LLVMCudd ${llvm_libs})
 else()
-    target_link_libraries(Svf ${llvm_libs})
+    target_link_libraries(Svf)
 endif()
+set_property(TARGET LLVMSvf PROPERTY POSITION_INDEPENDENT_CODE ON)
```
yuleisui: pic will hurt performance. When compiling to SVF executables (e.g., wpa), can we add an option to use the non-pic one for our built-in exes, while exposing to external users for building their own project pic is enabled?

PR #108, opened 2018-10-31
expose getCallInstToCallGraphEdgesMap for analysis

HongxuChen:
Sometimes we need `callinstToCallGraphEdgesMap` information to get calling relations between callers and callees; therefore it may be worthy to expose this analysis result.

### DIFF
--- a/include/Util/PTACallGraph.h
+++ b/include/Util/PTACallGraph.h
@@ -244,7 +244,7 @@ public:
         return numOfResolvedIndCallEdge;
     }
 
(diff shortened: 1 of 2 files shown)


yuleisui:
The map exposed to other classes may lead to changes to that map.
I think the following two will be enough for your case:

https://github.com/SVF-tools/SVF/blob/1f612e26ed56d9f41fcfc18a94b905502dff6fe6/include/Util/PTACallGraph.h#L317

https://github.com/SVF-tools/SVF/blob/1f612e26ed56d9f41fcfc18a94b905502dff6fe6/include/Util/PTACallGraph.h#L323

Issue #156, opened 2019-10-18
Building SVF as a shared library


FrancisYL:
Hi, first thank you for sharing and maintaining such a great project!

I'm trying to write my own LLVM pass and declare WPAPass as a dependency so as to get alias results of pointers and memory locations. In order to declare WPAPass as a dependency, I need to build a shared library of SVF and load it using `opt`.

SVF is built as a static library by default. I tried to build SVF as a shared library for my project but I encountered the problem described in #53. Even after I removed `${llvm_libs}`, I got segmentation fault querying alias results, which makes me think that SVF is not designed to be used as a shared library at its design phase.

So my questions are, first, what's the initial design reasoning for SVF, why building it as a static library instead of a shared library? Second, is there any way I can build SVF as a shared library so that I can load and declare WPAPass as my dependency to use AliasAnalysis? Thank you!

yuleisui:
@HongxuChen would you mind taking a look at the issue, which is similar to the one you addressed before?

HongxuChen:
@FrancisYL Can you provide the segfault backtrace and the triggering example to describe your issues?
For the shared library issue, do you mean to use it as a plugin pass (which can be used by utilities such as `opt -load xxx.so`)? I remember SVF previously supported this but due to some clashes the support was dropped.

FrancisYL:
@HongxuChen Here are the changes I made.

First, for `lib/CMakeLists.txt`, I changed `add_llvm_library(Svf ${SOURCES})` to `add_llvm_library(Svf SHARED ${SOURCES})` and commented out all the occurance of `target_link_libraries(Svf ${llvm_libs})`. This successfully build SVF and generate `libSvf.so`.

I then wrote my own module pass (called CAT) calling WPAPass as my dependency by adding `analysisUsage.addRequired<WPAPass>()`. In runOnModule method I fetched the analysis result `WPAPass& wpa = getAnalysis<WPAPass>()` and provide two memory location to check whether they alias `wpa.alias(MemoryLocation::get(memInst), MemoryLocation::get(memInst2))`

I used `opt` to call my pass using
`opt -load ~/projects/SVF-9/Release-build/lib/libSvf.so -load ~/CAT/lib/CAT.so program.bc -disable-output -CAT`. I believed the library is successfully loaded and there's no 'bitcode-mdindex-threshold' is registered more than once issue mentioned in #53.

But this time I got a Segmentation fault when querying `WPAPass::alias`.
```
Stack dump:
[...]

PR #160, opened 2019-10-29
Add no Pointer Analysis is specified assertion

FrancisYL:
WPAPass is able to run different pointer analysis on the bitcode but no default pointer analysis is used. This leads to the segmentation fault mention in issue #156 when querying the alias result.

In order to fix this issue, it would be safe to check if any pointer analysis is chosen before doing any further work.

This commit adds an assertion to make sure that lacking pointer analysis is not allowed. It runs `wpa` as a command-line tool or invokes `WPAPass` through a shared library. In both cases, the program will abort if no pointer analysis is specified.


yuleisui:
Thanks. It looks good to me.

Issue #176, opened 2020-02-15
ref contains all the mod in Mod-Ref Analysis


FrancisYL:
I'm implementing interprocedural getModRefInfo APIs based on the mod-ref results generated by SVF and found one line of code in [MSSA/MemRegion.cpp](https://github.com/SVF-tools/SVF/blob/master/lib/MSSA/MemRegion.cpp#L520) decreases the accuracy.

Below is an example.
```C
int get_value(int* p, int a) {
  *p = a;
  return a;
}

int main(int argc, const char* argv[]) {
  int a = argc;
  int* p = &a;

  int value = get_value(p, a);

  printf("%d", value);
  return 0;
}
```

the memory location pointed by `p` is only modified in `get_value`, not referenced. A **Mod** result is expected when querying the mod-ref result between callsite `get_value(p, a)` and pointer `p` through two attributes `csToRefsMap` and `csToModsMap`. However, `ref |= mod` assigns all the mod to ref of this callsite, which leads to a less accurate result **ModRef**.

After removing `ref |= mod`, the accuracy increases and **Mod** result becomes possible. It also affects the MSSA generated afterward. Only **χ** will be annotated around callsite if there's only an indirect mod but no ref of the memory region. Given an **χ** represents both use and def of a memory region semantically, removing the redundant **μ** around the callsite seems reasonable.

I'm not sure is any upper client application gets affected so I raised this issue for further inspection. Thank you.

yuleisui:
In fact, you have answered your question. Any modification, e.g., x = \chi(x) represents both a mod and use of x.

For a callsite,  \mu(x) often annotated before a callsite denotes an "invisible" passing parameter for an address-taken variable x. Missing \mu will not be able to pass variable x into the callee function. x = \chi(x) annotated after the callsite denotes whether we will take the old value of x from the caller, or a new value of x modified from callee.

yuleisui:
Thanks!

PR #177, opened 2020-02-17
delay inclusion of mod to ref to preserve useful mod-ref analysis result

FrancisYL:
As raised in [issue 176](https://github.com/SVF-tools/SVF/issues/176). An early stage of inclusion of mod to ref in [MRGenerator::handleCallsiteModRef](https://github.com/SVF-tools/SVF/blob/master/lib/MSSA/MemRegion.cpp#L520) limits the accuracy of mod-ref analysis and makes two intermediate maps `csToRefsMap` and `csToModsMap` unuseable. Since `funToPointsToMap`, `callsiteToRefPointsToMap` and `callsiteToModPointsToMap` are the three maps that contain def-use information of memory objects and get propagated to generate the correct MSSA, a delay of inclusion of mod to ref to a later stage in [MRGenerator::collectModRefForCall()](https://github.com/SVF-tools/SVF/blob/master/lib/MSSA/MemRegion.cpp#L219) can both preserve the useful information of mod-ref analysis and maintain the correctness of MSSA.

### DIFF
--- a/include/MSSA/MemRegion.h
+++ b/include/MSSA/MemRegion.h
@@ -325,11 +325,11 @@ protected:
         addRefSideEffectOfFunction(fun,cpts);
     }
     inline void addCPtsToCallSiteRefs(PointsTo& cpts, CallSite cs) {
-        callsiteToRefPointsToMap[cs] = cpts;
+        callsiteToRefPointsToMap[cs] |= cpts;
         funToPointsToMap[cs.getCaller()].insert(cpts);
     }
     inline void addCPtsToCallSiteMods(PointsTo& cpts, CallSite cs) {
-        callsiteToModPointsToMap[cs] = cpts;
+        callsiteToModPointsToMap[cs] |= cpts;
         funToPointsToMap[cs.getCaller()].insert(cpts);
     }
     inline bool hasCPtsList(const Function* fun) const {
--- a/lib/MSSA/MemRegion.cpp
+++ b/lib/MSSA/MemRegion.cpp
@@ -218,7 +218,9 @@ void MRGenerator::collectModRefForCall() {
         }
         if(hasModSideEffectOfCallSite(*it)) {
             NodeBS mods = getModSideEffectOfCallSite(*it);
+            /// mods are treated as both def and use of memory objects
             addCPtsToCallSiteMods(mods,*it);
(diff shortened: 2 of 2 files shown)


yuleisui:
Hello,

This is a very good fix. Many thanks!

It would be good that you can have an example to show your improvement?

Yulei

FrancisYL:
```C
int get_value(int* p, int a) {
  *p = a;
  return a;
}

int main(int argc, const char* argv[]) {
  int a = argc;
  int* p = &a;

  int value = get_value(p, a);

  printf("%d", value);
  return 0;
}
```

When querying the interprocedural mod-ref result between callsite `get_value(p, a)` and pointer `p`, a **Mod** result can now be derived simply by finding that NodeID of object `a` is in `csToModsMap` but not `csToRefsMap`.

It will not affect the construction of MSSA since later `callsiteToRefPointsToMap` and `funToPointsToMap` include all objects from `csToRefsMap` and `csToModsMap`, which leads to the correct annotation of \mu & \chi operations around callsite `get_value(p, a)`.

```
==========FUNCTION: _Z9get_valuePii==========
2V_1 = ENCHI(MR_2V_0) 	pts{19 }
  store i32 %1, i32* %0, align 4
2V_2 = STCHI(MR_2V_1) 	pts{19 }

  ret i32 %1
RETMU(MR_2V_2) 	pts{19 }
==========FUNCTION: main==========
2V_1 = ENCHI(MR_2V_0) 	pts{19 }
  %3 = alloca i32, align 4
  store i32 %0, i32* %3, align 4
2V_2 = STCHI(MR_2V_1) 	pts{19 }

LDMU(MR_2V_2) 	pts{19 }
  %4 = load i32, i32* %3, align 4

CALMU(MR_2V_2) 	pts{19 }
  %5 = call i32 @_Z9get_valuePii(i32* %3, i32 %4)
2V_3 = CALCHI(MR_2V_2) 	pts{19 }

  %6 = getelementptr [3 x i8], [3 x i8]* @.str, i64 0, i64 0
  %7 = call i32 (i8*, ...) @printf(i8* %6, i32 %5)
  ret i32 0
RETMU(MR_2V_3) 	pts{19 }
```

Based on the changes in this commit, SVF will be able to handle mod-ref queries. I will pr my implementation of `getModRefInfo` APIs once this commit gets merged.

PR #178, opened 2020-02-18
add getModRefInfo APIs support for SVF

FrancisYL:
As in pr #177, SVF now naturally support interprocedural mod-ref queries by leveraging two maps `csToRefsMap` and `csToModsMap`. This commit adds three types of mod-ref query APIs to SVF.
* mod-ref query of whether a callInst instruction can mod or ref any memory location
* mod-ref query of whether a callInst instruction can mod or ref a specific memory location
* mod-ref info between two callInst instructions, which is in line with [LLVM getModRefInfo between two calls](https://llvm.org/docs/AliasAnalysis.html#the-getmodrefinfo-methods)

In order to make queries to those two maps above, a `std::vector<SVFG*> svfgVector` is added in `class WPA` to store all the svfg constructed based on Andersen's Pointer Analysis. BitVector stores the union of all query results. The query result rule is bold (less conservative), that is, `NoModRef` is favored the most than `Ref`, then `Mod` and finally `ModRef`. If `Ref` and `Mod` result appear at the same time, then `NoModRef` result gets returned.

Example 1: One ref and one mod
```C
int get_value(int* p, int* q, int a) {
  *p = a;
  return *q;
}

int main(int argc, const char* argv[]) {
  int a = argc;
  int b = argc;
  int* p = &a;
  int* q = &b;

  int value = get_value(p, q, a);

  printf("%d", value);
  return 0;
}
```
Memory pointed by `p` is modified and memory pointered by pointer `q` is referenced in get_value().
The mod-ref result of pointer `p` is **Mod** and of pointer `q` is **Ref**.
The mod-ref result of callInst `int value = get_value(p, q, a)` is **ModRef**.

Example 2: Mod inside another call
```C
int get_value2(int* p, int a) {
  *p = a;
  return a;
}

int get_value(int* p, int a) {
  return get_value2(p, a);
}

int main(int argc, const char* argv[]) {
  int a = argc;
  int* p = &a;

  int value = get_value(p, a);

  printf("%d", value);
  return 0;
}
```
The memory pointed by `p` is modified through another callInst inside `get_value()`.
The mod-ref result of callInst `int value = get_value(p, a)` and `p` is **Mod**.

Example 3: ModRef through indirect calls
```C
int get_value2(int* p, int a) {
  *p = a;
  return a;
}

int get_value1(int* p, int a) {
  return *p;
}

int main(int argc, const char* argv[]) {
  int a = argc;
  int* p = &a;

  int (*f)(int*, int*, int);
  if (a > 10)
    f = &get_value;
  else
    f = &get_value2;

  int value = f(p, a);

  printf("%d", value);
  return 0;
}
```
The memory pointed by `p` can either be referred through a function pointer that refers it or another function pointer that modifies it. The mod-ref result of pointer `p` is **ModRef**.

Example 4: ModRef between two calls
```C
int get_value1(int* p, int a) {
  *p = *p + 1;
  return a;
}

int get_value2(int* p, int a) {
  *p = a;
  return a;
}

int main(int argc, const char* argv[]) {
  int a = argc;
  int* p = &a;

  int value1 = get_value1(p, a);
  int value2 = get_value2(p, a);

  printf("%d %d", value1, value2);
  return 0;
}
```
The first call referred and modified the memory modified by the second call. The mod-ref result of `int value1 = get_value1(p, a)` and ` int value2 = get_value2(p, a)` is **ModRef**, and the result between ` int value2 = get_value2(p, a)` and ` int value1 = get_value1(p, a)` is only **Mod** since the second call mod the memory ref or mod by the first call.

Since its an extension by leveraging internal data of SVF. Please let me know are there anything wrong or suggestions. Thanks!

### DIFF
--- a/include/WPA/WPAPass.h
+++ b/include/WPA/WPAPass.h
@@ -40,6 +40,7 @@
 #include "MemoryModel/PointerAnalysis.h"
 
 class SVFModule;
+class SVFG;
 
 /*!
  * Whole program pointer analysis.
@@ -91,6 +92,20 @@ public:
     /// Print all alias pairs
     virtual void PrintAliasPairs(PointerAnalysis* pta);
 
+    /// Interface of mod-ref analysis to determine whether a CallSite instruction can mod or ref any memory location
+    virtual ModRefInfo getModRefInfo(const CallInst* callInst);
+
+    /// Interface of mod-ref analysis to determine whether a CallSite instruction can mod or ref a specific memory location, given Location infos
+    virtual inline ModRefInfo getModRefInfo(const CallInst* callInst, const MemoryLocation& Loc) {
+        return getModRefInfo(callInst, Loc.Ptr);
+    }
+
+    /// Interface of mod-ref analysis to determine whether a CallSite instruction can mod or ref a specific memory location, given Value infos
+    virtual ModRefInfo getModRefInfo(const CallInst* callInst, const Value* V);
+
+    /// Interface of mod-ref analysis between two CallSite instructions
+    virtual ModRefInfo getModRefInfo(const CallInst* callInst1, const CallInst* callInst2);
+
     /// We start from here
     virtual bool runOnModule(llvm::Module& module) {
         SVFModule svfModule(module);
@@ -112,6 +127,7 @@ private:
 
     PTAVector ptaVector;	///< all pointer analysis to be executed.
     PointerAnalysis* _pta;	///<  pointer analysis to be executed.
+    SVFG* _svfg;  ///< svfg generated through -ander pointer analysis
 };
 
 
--- a/lib/MSSA/MemRegion.cpp
+++ b/lib/MSSA/MemRegion.cpp
@@ -557,3 +557,83 @@ void MRGenerator::modRefAnalysis(PTACallGraphNode* callGraphNode, WorkList& work
         }
     }
 }
+
+/*!
+ * Determine whether a CallSite instruction can mod or ref
+ * any memory location
+ */
+ModRefInfo MRGenerator::getModRefInfo(CallSite cs) {
+    bool ref = hasRefSideEffectOfCallSite(cs);
+    bool mod = hasModSideEffectOfCallSite(cs);
+
+    if (mod && ref)
+        return ModRefInfo::ModRef;
+    else if (ref)
+        return ModRefInfo::Ref;
+    else if (mod)
+        return ModRefInfo::Mod;
+    else
+        return ModRefInfo::NoModRef;
+}
+
+/*!
+ * Determine whether a CallSite instruction can mod or ref
+ * a specific memory location pointed by V
+ */
+ModRefInfo MRGenerator::getModRefInfo(CallSite cs, const Value* V) {
+    bool ref = false;
+    bool mod = false;
+
+    if (pta->getPAG()->hasValueNode(V)) {
+        const PointsTo& pts(pta->getPts(pta->getPAG()->getValueNode(V)));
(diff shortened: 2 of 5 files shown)


FrancisYL:
Amendment has been added to this commit.

1. Remove mod-ref apis in SVFG.h and MemSSA.h to query mod-ref directly through `svfg->getMSSA()->getMRGenerator()->getModRefInfo(..)`.
2. Use one `SVFG* _svfg` instead of `svfgVector` to store the SVFG generated only through `-ander`. Assertions are also added at the beginning of all query functions.

One thing to mention is both `-ander` and `-wander` share the same enum of [AndersenWaveDiff_WPA](https://github.com/SVF-tools/SVF/blob/master/lib/WPA/WPAPass.cpp#L60). For now, the mod-ref query will be enabled when `-svfg` and either `-ander` or `-wander` turns on.

include/WPA/WPAPass.h
```diff
@@ -38,6 +38,7 @@
 #define WPA_H_
 
 #include "MemoryModel/PointerAnalysis.h"
+#include "MSSA/SVFG.h"
```
yuleisui: We do not need to include this header file.

`Class SVFG` would be enough.

lib/MSSA/MemRegion.cpp
```diff
@@ -557,3 +557,83 @@ void MRGenerator::modRefAnalysis(PTACallGraphNode* callGraphNode, WorkList& work
...
+ */
+ModRefInfo MRGenerator::getModRefInfo(CallSite cs1, CallSite cs2) {
+    bool ref = false;
+    bool mod = false;
+
+    /// return NoModRef neither two callsites ref or mod any memory
+    if (getModRefInfo(cs1) == ModRefInfo::NoModRef || getModRefInfo(cs2) == ModRefInfo::NoModRef)
+        return ModRefInfo::NoModRef;
+
+    PointsTo cs1Ref = getRefSideEffectOfCallSite(cs1);
```
yuleisui: If you are not modifying the Ref and Mod sets, you can consider using references to avoid copy construction.

`PointsTo cs1Ref = getRefSideEffectOfCallSite(cs1)`
or
`const PointsTo cs1Ref = getRefSideEffectOfCallSite(cs1)'

lib/MSSA/MemRegion.cpp
```diff
@@ -557,3 +557,83 @@ void MRGenerator::modRefAnalysis(PTACallGraphNode* callGraphNode, WorkList& work
...
+/*!
+ * Determine whether a CallSite instruction can mod or ref
+ * a specific memory location pointed by V
+ */
+ModRefInfo MRGenerator::getModRefInfo(CallSite cs, const Value* V) {
+    bool ref = false;
+    bool mod = false;
+
+    if (pta->getPAG()->hasValueNode(V)) {
+        PointsTo pts(pta->getPts(pta->getPAG()->getValueNode(V)));
```
yuleisui: avoid copy construction try using const reference
`const PointsTo& pts = pta->getPts(pta->getPAG()->getValueNode(V));`

FrancisYL:
Hi Yulei,

Thank you for the code review! I just amended the latest change to this commit.

Also, thank you for the suggestion for replacing `_svfg` with a manually constructed SVFG. However, I encountered a segmentation fault through that approach. I wondered it may caused by no svfModule is found.

I built SVF as a shared library and invoked it as a pass inside my LLVM pass. `WPAPass& wpa = getAnalysis<WPAPass>()`. My intent is to query both `alias` and `getModRefInfo` APIs and it would be easier for me to simply query it through `wpa.alias()/wpa.getModRefInfo()`. Given a manual construction of SVFG increases the difficulty of future maintenance. A temp `_svfg` really eases my job.

Best,

PR #185, opened 2020-02-26
Fix dockerfile

prashast:
Due to `LLVM_DIR` environment variable not being configured the SVF setup would fail. This pull request fixes that along with adding the compiled tools to `PATH`.


yuleisui:
Thanks~

PR #186, opened 2020-02-27
Add support to use SVF as a lib in a shared object.

cgiuffr:
See subject. This is useful to be able to build out-of-tree LLVM passes using SVF.


yuleisui:
Thanks for your patch.

cgiuffr:
> Thanks for your patch.

Thanks for the quick response!

Issue #188, opened 2020-03-10
MSSA misses to include struct object through function call


FrancisYL:
I have the following program which passes a Node pointer `n` and appends it with a new value to the Node pointed by `n`
```C++
struct Node {
  int value;
  Node *next;
};

void appendNode(Node *n, int newValue) {
  Node* newNode = (Node *)malloc(sizeof(Node));
  newNode->value = newValue;
  newNode->next = NULL;
  n->next = newNode;
}

int main(int argc, char **argv) {
  Node *n = (Node *)malloc(sizeof(Node));
  n->value = argc;
  n->next = NULL;
  appendNode(n, 10);

  return 0;
}
```

However, the MSSA generated fails to catch the pass of Node object after casting of pointer `n` in the bitcode.
```
==========FUNCTION: main==========
2V_1 = ENCHI(MR_2V_0) 	pts{30 }
4V_1 = ENCHI(MR_4V_0) 	pts{1030 }
6V_1 = ENCHI(MR_6V_0) 	pts{10 1010 }

CALMU(MR_2V_1) 	pts{30 }
  %3 = call noalias i8* @malloc(i64 16) #3
2V_2 = CALCHI(MR_2V_1) 	pts{30 }

  %4 = bitcast i8* %3 to %struct.Node*
  %5 = getelementptr inbounds %struct.Node, %struct.Node* %4, i32 0, i32 0
  store i32 %0, i32* %5, align 8
2V_3 = STCHI(MR_2V_2) 	pts{30 }

  %6 = getelementptr inbounds %struct.Node, %struct.Node* %4, i32 0, i32 1
  store %struct.Node* null, %struct.Node** %6, align 8
4V_2 = STCHI(MR_4V_1) 	pts{1030 }

CALMU(MR_4V_2) 	pts{1030 }
CALMU(MR_6V_1) 	pts{10 1010 }
  call void @_Z10appendNodeP4Nodei(%struct.Node* %4, i32 10)
4V_3 = CALCHI(MR_4V_2) 	pts{1030 }
6V_2 = CALCHI(MR_6V_1) 	pts{10 1010 }

  ret i32 0
RETMU(MR_2V_3) 	pts{30 }
RETMU(MR_4V_3) 	pts{1030 }
RETMU(MR_6V_2) 	pts{10 1010 }
```

Could you please take a look at this bug? many thanks!

yuleisui:
Can you elaborate a bit more about this, or you may wish to propose a solution?

[...]

PR #190, opened 2020-03-23
Refine getModRefInfo APIs to accomodate with field-sensitivity of SVF

FrancisYL:
SVF is field-sensitive when treating with class/struct objects. This causes the current getModRefInfo APIs to return NoModRef if a callisite modified/referred all fields except the base field of an object, as raised in issue #188.

The solution is to expand the mod/ref PointsTo set of callsite or the base pointer to include all fields of an object if the PointsTo set contains the base pointer.

Example
```C++
struct S {
  int v1;
  int v2;
  int v3;
};

void setV1(S *s, int v) {
  s->v1 = v;
}

void setV2(S *s, int v) {
  s->v2 = v;
}

void setV3(S *s, int v) {
  s->v3 = v;
}

  S *s = (S *)malloc(sizeof(S));
  s->v1 = argc;

  setV1(s, argc);
  setV2(s, argc);
  setV3(s, argc);

  return 0;
}
```

The mod-ref result between pointer `s` and callsite `setV2(s, argc)` or `setV3(s, argc)` now returns Mod. The mod-ref result between callsite `setV2(s, argc)` and `setV3(s, argc)` still returns NoModRef since they are not modifying the base field but modifying two different fields of the object.

### DIFF
--- a/lib/MSSA/MemRegion.cpp
+++ b/lib/MSSA/MemRegion.cpp
@@ -585,10 +585,17 @@ ModRefInfo MRGenerator::getModRefInfo(CallSite cs, const Value* V) {
     bool mod = false;
 
     if (pta->getPAG()->hasValueNode(V)) {
-        const PointsTo& pts(pta->getPts(pta->getPAG()->getValueNode(V)));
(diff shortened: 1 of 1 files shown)


yuleisui:
Thanks!

PR #195, opened 2020-04-15
Return conservative getModRefInfo result for external calls

FrancisYL:
The current getModRefInfo APIs returns `NoModRef` for external function calls, which is not correct (my bad). This commit fixes the issue and returns the conservative `ModRef` results.

External calls to `malloc`, `realloc`, etc. are not affected and will still return a more accurate `Mod` result as they are specially treated as [HeapAllocExtCall](https://github.com/SVF-tools/SVF/blob/master/lib/MSSA/MemRegion.cpp#L512).

Example:
```C
int main() {
  char* str = "Hello World";

  printf("%s\n", str);
  printf("%s\n", str);

  return 0;
}
```

The mod-ref result between the first and second `printf` external calls returns `ModRef`.

### DIFF
--- a/lib/MSSA/MemRegion.cpp
+++ b/lib/MSSA/MemRegion.cpp
@@ -571,7 +571,10 @@ ModRefInfo MRGenerator::getModRefInfo(const CallBlockNode* cs) {
     bool ref = hasRefSideEffectOfCallSite(cs);
     bool mod = hasModSideEffectOfCallSite(cs);
 
-    if (mod && ref)
+    /// return ModRef if tracks nothing of an external call
+    if (isExtCall(cs->getCallSite()) && !ref && !mod)
+        return ModRefInfo::ModRef;
+    else if (mod && ref)
         return ModRefInfo::ModRef;
     else if (ref)
         return ModRefInfo::Ref;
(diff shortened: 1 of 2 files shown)


yuleisui:
Why not reverting to the previous implementation if you think your previous patch is incorrect?

Would this fix be sound and precise?

lib/MSSA/MemRegion.cpp
```diff
@@ -571,7 +571,10 @@ ModRefInfo MRGenerator::getModRefInfo(const CallBlockNode* cs) {
     bool ref = hasRefSideEffectOfCallSite(cs);
     bool mod = hasModSideEffectOfCallSite(cs);
 
-    if (mod && ref)
+    /// return ModRef if tracks nothing of an external call
+    if (isExtCall(cs->getCallSite()) && !ref && !mod)
```
yuleisui, line 575: Can we put the `isExtCall` inside `hasRefSideEffectOfCallSite` or `hasSideEffectOfCallSite` to make the code cleaner? It looks a bit ad-hoc here.

FrancisYL:
I believe `printf` at least ref the first global constant string and the second string object in the above example.

A `Ref` or `ModRef` result is sound since two calls access the same object.

yuleisui:
In fact, we use ExternalAPI to define the side-effects of external functions and the parameter passing between application and the library code are modelled using extra PAG nodes and edges created during PAGBuilder. If you change the side-effect in `ExtAPI.cpp`, it will change the mod-ref analysis results naturally.

https://github.com/SVF-tools/SVF/blob/f7b3d0e3b51b25e3894d5645a0f5172284645f29/include/Util/ExtAPI.h

and 

https://github.com/SVF-tools/SVF/blob/f7b3d0e3b51b25e3894d5645a0f5172284645f29/lib/SVF-FE/PAGBuilder.cpp#L760

FrancisYL:
Thanks for the suggestion! Could you also guide me on which `extf_t` would be appropriate for the `printf`?

Since `printf` is a vararg function and loads it's first argument, for the other arguments, it could be a load or no side effect.
```C
printf("%s\n", s);        // load "%s\n" and s
printf("%d\n", intValue); // load "%s\n" but has no side effect on intValue
```

PR #196, opened 2020-04-15
Set moduleIdentifier of svfModule built from LLVM Module

FrancisYL:
A simple fix to pass the assertion
https://github.com/SVF-tools/SVF/blob/f7b3d0e3b51b25e3894d5645a0f5172284645f29/include/Util/SVFModule.h#L134


yuleisui:
It seems that there it was a broken built indicated by Travis. Can you double check it?

RafaeNoor:
I have the exact same error appearing

yuleisui:
Will merge it.

PR #197, opened 2020-04-15
Set mod ref sets for external callsite

FrancisYL:
Though SVF handles external functions' def-use behavior by defining a customized `extf_t`, it misses to include the results into the mod-ref analysis other than `HeapAllocExtCall`.

The reason is that in
https://github.com/SVF-tools/SVF/blob/f7b3d0e3b51b25e3894d5645a0f5172284645f29/lib/MSSA/MemRegion.cpp#L510
the `getModSideEffectOfFunction(callee)` and `getRefSideEffectOfFunction(callee)` returns nothing for callee functions who don't have definitions. To set it's mod ref sets, we have to retrieve it from the PAG edges constructed from
https://github.com/SVF-tools/SVF/blob/f7b3d0e3b51b25e3894d5645a0f5172284645f29/lib/SVF-FE/PAGBuilder.cpp#L760

An example like below
```C
  char* s1 = (char *)malloc(sizeof(argc));    // src
  char* s2 = (char *)malloc(sizeof(argc));    // dst
  memcpy(s2, s1, argc);

  return 0;
}
```
The mod ref behavior of external call `memcpy(s2, s1, argc)` is now returning the sound and correct `ModRef` result.

### DIFF
--- a/include/MSSA/MemRegion.h
+++ b/include/MSSA/MemRegion.h
@@ -306,6 +306,9 @@ protected:
         aliasMRs.insert(mr);
     }
 
+    /// Collect mod ref for external callsite other than heap alloc external call
+    virtual void collectModRefForExtCallSiteOtherThanHeapAlloc(const CallBlockNode* cs);
+
     /// Mod-Ref analysis for callsite invoking this callGraphNode
     virtual void modRefAnalysis(PTACallGraphNode* callGraphNode, WorkList& worklist);
 
--- a/lib/MSSA/MemRegion.cpp
+++ b/lib/MSSA/MemRegion.cpp
@@ -191,6 +191,7 @@ void MRGenerator::collectModRefForCall() {
     for(PAG::CallSiteSet::const_iterator it =  pta->getPAG()->getCallSiteSet().begin(),
             eit = pta->getPAG()->getCallSiteSet().end(); it!=eit; ++it){
         collectCallSitePts((*it));
+        collectModRefForExtCallSiteOtherThanHeapAlloc((*it));
     }
 
     DBOUT(DGENERAL, outs() << pasMsg("\t\tPerform Callsite Mod-Ref \n"));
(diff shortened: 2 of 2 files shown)


yuleisui:
It was good to dig it up! However, PAGBuilder has already `flattened` the external function by adding PAGEdges and Nodes. Therefore, it is unnecessary to collect side-effects again during the iterative mod-ref analysis. It will not contribute to any precision improvement. 

I understand that you need to query the mod-ref results for callsites. I would suggest you can modify your query API here:

https://github.com/SVF-tools/SVF/blob/f7b3d0e3b51b25e3894d5645a0f5172284645f29/lib/WPA/WPAPass.cpp#L230

FrancisYL:
Hi Yulei, thank you so much for the comment. I do am interested in querying the mod-ref results between callsite and pointer or between two callsites. And thanks for maintaining such a great project to make querying possible and easy to implement.

I think it's still worth setting the mod-ref set for external callsite. Though the external calls are `flatten` by adding dummy nodes or object nodes in PAGBuilder, it makes the value flow in the MSSA clearer with these sets available. An example would be
```C
  char *s = (char *)malloc(sizeof(char));
  memcpy(s, argv[0], argc);

  return 0;
}
```
and the MSSA dumped
```LLVM
==========FUNCTION: main==========
20V_1 = ENCHI(MR_20V_0) 	pts{24 }

CALMU(MR_20V_1) 	pts{24 }
  %3 = call noalias i8* @malloc(i64 1) #5
20V_2 = CALCHI(MR_20V_1) 	pts{24 }

  %4 = getelementptr inbounds i8*, i8** %1, i64 0
  %5 = load i8*, i8** %4, align 8
  %6 = sext i32 %0 to i64
  call void @llvm.memcpy.p0i8.p0i8.i64(i8* align 1 %3, i8* align 1 %5, i64 %6, i1 false)
  ret i32 0
RETMU(MR_20V_3) 	pts{24 }
```
It would be hard to track how `MR_20V_2` gets modified to be `MR_20V_3` if there're multiple external calls between them. With the sets available, the MSSA would be clearer
```LLVM
==========FUNCTION: main==========
26V_1 = ENCHI(MR_26V_0) 	pts{24 }

CALMU(MR_26V_1) 	pts{24 }
  %3 = call noalias i8* @malloc(i64 1) #5
26V_2 = CALCHI(MR_26V_1) 	pts{24 }

  %4 = getelementptr inbounds i8*, i8** %1, i64 0
  %5 = load i8*, i8** %4, align 8
  %6 = sext i32 %0 to i64

CALMU(MR_26V_3) 	pts{24 }
  call void @llvm.memcpy.p0i8.p0i8.i64(i8* align 1 %3, i8* align 1 %5, i64 %6, i1 false)
26V_4 = CALCHI(MR_26V_3) 	pts{24 }

  ret i32 0
RETMU(MR_26V_4) 	pts{24 }
```

I sometimes need to query the mod-ref result between two callsites and one or two of them may be external calls. It's easier to query with these sets available in the same scope.

In terms of the iterative modrefAnalysis algorithm, since external calls that have side effects would contribute at most once to the overall execution. I can lift this computation outside the loop to reduce complexity.

yuleisui:
Thanks for your explanation and it makes sense to me. Will merge it.

yuleisui:
Just had a look and tested several benchmarks at this patch.  The number of callsite mu/chi increased around 6% on average on the benchmarks I tested. These mus/chis would not increase precision since the external calls have been flattened. I have made the changes on top of your patch by (1) dumping the correct mod-ref for external calls (`MemSSA::dumpMSSA`), and (2) add the new helper functions `getModInfoForCall` and `getRefInfoForCall`. Could you please help check whether they are ok?

https://github.com/SVF-tools/SVF/pull/201/commits/c1a205eddee8e369e153c2bd0e2b28e5c3a4898f

FrancisYL:
Thanks for doing that and I really appreciate it! I've taken a look and found two small issues. I will push a request soon.

Issue #199, opened 2020-04-17
Modeling IO Buffer Object with SVF


FrancisYL:
Hello,

I was wondering if it's a good idea to extend the current SVF memory model by adding an extra IO Buffer Object (NodeID = 4 with Type = Int8Ty) and Value (NodeID = 5).

std functions like `printf()`, `scanf()` do have side-effect on the IO buffer. There will be a race when multiple `printf()` trying to print to any output stream and the mod-ref results between any two `printf()` call pairs should be `ModRef`.
```C
printf("This prints first");
printf("This prints second");
```

One or more `extf_t` might be needed and external functions that perform IO operations defined in the https://github.com/SVF-tools/SVF/blob/f06eba337a17954981c6a470bcd273607c992f06/lib/Util/ExtAPI.cpp#L24 will get changed. This IO buffer modeling would benefit my project a lot and I'm glad to help with implementing that.

yuleisui:
Yes, I am ok if you want to add the READ side-effects for the existing IO functions or others.

The following steps might be helpful:

1. add one or more side-effect enum in ExtAPI to indicate which parameter or its references will have a READ side-effect.
https://github.com/SVF-tools/SVF/blob/f06eba337a17954981c6a470bcd273607c992f06/include/Util/ExtAPI.h#L79

2. handle your case by adding load PAGEdges
https://github.com/SVF-tools/SVF/blob/f06eba337a17954981c6a470bcd273607c992f06/lib/SVF-FE/PAGBuilder.cpp#L986

3. add or modify the existing functions' side-effects in ExtAPI.cpp
https://github.com/SVF-tools/SVF/blob/f06eba337a17954981c6a470bcd273607c992f06/lib/Util/ExtAPI.cpp#L310

yuleisui:
Once you add the READ side-effect, I am happy to have a look at your pull request. Thanks

FrancisYL:
Thanks!

PR #200, opened 2020-04-18
Modeling I/O Buffer within SVF and add necessary methods

FrancisYL:
Link to issue #199, the data flow of any function who performs I/O operation can be more accurately modeled with an I/O buffer object.

This commit adds `IOBufferObj`(NodeID = 4) and `IOBufferPtr`(NodeID = 5) to the SymbolTable of SVF  combined with necessary methods.

I assume this wouldn't break any other assertion checks in SVF since this new object is only for modeling purpose and no pointer can point to it.

Please feel free to let me know if anything needs to be changed. Thanks!

### DIFF
--- a/include/Graphs/PAG.h
+++ b/include/Graphs/PAG.h
@@ -522,6 +522,25 @@ public:
     }
     //@}
 
+    /// Get I/O buffer
+    //@{
+    inline NodeID getIOBufferNode() const {
+        return symInfo->IOBufferSymID();
+    }
+    inline NodeID getIOBufferPtr() const {
+        return symInfo->IOBufferPtrSymID();
+    }
+    inline bool isIOBufferObj(NodeID id) const {
+        return (SymbolTableInfo::isIOBufferObj(id));
+    }
+    inline bool isIOBufferPtr(NodeID id) const {
+        return (SymbolTableInfo::isIOBufferPtr(id));
+    }
+    inline const MemObj* getIOBufferObj() const {
+        return symInfo->getIOBufferObj();
+    }
+    //@}
+
     inline u32_t getNodeNumAfterPAGBuild() const {
         return nodeNumAfterPAGBuild;
     }
@@ -615,6 +634,12 @@ public:
     }
     inline NodeID addBlackholePtrNode() {
         return addDummyValNode(getBlkPtr());
+    }
+    inline NodeID addIOBufferObjNode() {
+        return addObjNode(NULL, new DummyObjPN(getIOBufferNode(),getIOBufferObj()), getIOBufferNode());
+    }
+    inline NodeID addIOBufferPtrNode() {
+        return addDummyValNode(getIOBufferPtr());
     }
     //@}
 
--- a/include/MemoryModel/MemModel.h
+++ b/include/MemoryModel/MemModel.h
@@ -40,6 +40,8 @@ enum SYMTYPE {
     ConstantObj,
     BlkPtr,
     NullPtr,
+    IOBufferObj,
+    IOBufferPtr,
     ValSym,
     ObjSym,
     RetSym,
@@ -259,7 +261,7 @@ public:
     /// Constructor
     MemObj(const Value *val, SymID id);
 
-    /// Constructor for black hole and constant obj
+    /// Constructor for black hole, constant obj and I/O buffer obj
     MemObj(SymID id, const Type* type = NULL);
 
     /// Destructor
@@ -270,7 +272,7 @@ public:
     /// Initialize the object
     void init(const Value *val);
 
-    /// Initialize black hole and constant object
+    /// Initialize black hole, constant object and I/O buffer obj
     void init(const Type* type);
 
     /// Get obj type
@@ -306,6 +308,9 @@ public:
 
     /// Whether it is a black hole object
     bool isBlackHoleObj() const;
+
+    /// Whether it is an IO buffer object
+    bool isIOBufferObj() const;
 
     /// object attributes methods
     //@{
--- a/include/SVF-FE/SymbolTableInfo.h
+++ b/include/SVF-FE/SymbolTableInfo.h
@@ -179,6 +179,9 @@ public:
     static inline bool isNullPtr(NodeID id) {
         return (id == NullPtr);
     }
+    static inline bool isIOBufferPtr(NodeID id) {
+        return (id == IOBufferPtr);
+    }
     static inline bool isBlkObj(NodeID id) {
         return (id == BlackHole);
     }
@@ -188,6 +191,9 @@ public:
     static inline bool isBlkObjOrConstantObj(NodeID id) {
         return (isBlkObj(id) || isConstantObj(id));
     }
+    static inline bool isIOBufferObj(NodeID id) {
+        return (id == IOBufferObj);
+    }
 
     inline void createBlkOrConstantObj(SymID symId) {
         assert(isBlkObjOrConstantObj(symId));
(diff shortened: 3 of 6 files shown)


yuleisui:
If you take a look at my previous three steps, I do not think we need an additional IObuffer node for mod-ref analysis?

FrancisYL:
Yeah thanks for the suggestion in the previous three steps. I think the main purpose is that I need one central object that could model all the I/O operations. For instance, a `printf()` would read its arguments and write to this I/O buffer for output. A `scanf()` would write to this I/O buffer for input and read from it.

yuleisui:
`printf` only has READ side effect and the modeling for central io involves modification which is not what we want?

FrancisYL:
My project, which is a Parallelizer, uses SVF's getModRefInfo to determine the dependence between callsites, basically, if two callsites don't have any write dependence between them then they could be parallelized. SVF returns sound results for most cases except for calls who involve I/O operations, either external I/O calls like `printf` or calls with I/O operations in its callee function. For instance
```C
void printWrapper1() {
  printf("Print first");
}

void printWrapper2() {
  printf("Print second");
}

int main() {
  printWrapper1();
  printWrapper2();
  return 0;
}
```
Based on the current result, my Parallelizer would tell me that `printWrapper1()` and `printWrapper2()` don't have write dependence between them so it just parallelizes it, which leads to an undeterministics results state since the printing order might get changed.

The way to not parallelize them is by having an I/O buffer object, that is, any calls who have I/O operations will have write effect on this object. It is more like a state object instead of a global since there's no LLVM pointer that can point to it but can only be tracked by a pointer defined inside SVF.

I've tried to model this object with blackhole but it fails because blackhole doesn't have any Type associated with it.

Given external call `puts`, a use case for this I/O buffer object would be
```C++
case ExtAPI::EFT_IO_A0R: {
  NodeID vnS = pag->getValueNode(cs.getArgument(0));
  NodeID dummy = pag->addDummyValueNode();
  addLoadEdge(vnS, dummy);            // arg0 gets referenced
  
  NodeID outputPtr = pag->getIOBufferPtr();
  addStoreEdge(dummy, outputPtr);     // the I/O buffer object gets modified
  
  break;
}
```

I personally think this I/O buffer object could be helpful since it also tracks the value flow of I/O operations. Please correct me if I'm wrong. Thanks!

yuleisui:
According to your description, your code parallelization is a client-specific problem. Adding an IO object will break the underlying semantic of MSSA and SVFG of the application code.

I think one way to determine your parallelization do is to collect the interested IO functions and inspect their READ side-effect. Then, you analyze the order of the objects being READ/REFERRED in these IO functions, to decide whether you will parallelize them or not.

yuleisui:
Another concern is that storing objects in an IO object on PAG may also cause more extra solving time during Andersen if we have many IO functions. Do you have a way to achieve your parallization using READ side effect only?

FrancisYL:
Unfortunately not. My current implementation uses both LLVM and SVF's AliasAnalysis (alias() and getModRefInfo() methods) and it will take whoever's results which are bolder (SVF's result is way bolder and better than LLVM, which leads to better parallelization and performance improvement).

I only determine which result to take but don't have extra check on the correctness of results because they are assumed to be correct or conservatively correct (more dependence is better than missing dependence).

Since I now found SVF gives a bolder but not correct result between callsites who involves I/O operations. The only way I can think of, for now, before querying, is to go through the callgraph to make sure the callsites do not invole any I/O operation, which is not a favored method and create extra burden on the cliend side.

yuleisui:
I guess LLVM's mod-ref will not give modification side effect for IO functions?

FrancisYL:
It returns `ModRef` results between callsites.

I attached a sample program and the output of my analysis pass.
```C
  int value;

  printf("Print something");
  scanf("Read value", value);
  puts("Print someing again");

  return 0;
}
```
```LLVM
  %4 = call i32 (i8*, ...) @printf(i8* %3)	@ "main"
	  %6 = call i32 (i8*, ...) @scanf(i8* %5, i32 undef)	@ "main"
		NoModRef  @ SVF
		ModRef  @ LLVM
	  %8 = call i32 @puts(i8* %7)	@ "main"
		NoModRef  @ SVF
		ModRef  @ LLVM
  %6 = call i32 (i8*, ...) @scanf(i8* %5, i32 undef)	@ "main"
	  %4 = call i32 (i8*, ...) @printf(i8* %3)	@ "main"
		NoModRef  @ SVF
		ModRef  @ LLVM
	  %8 = call i32 @puts(i8* %7)	@ "main"
		NoModRef  @ SVF
		ModRef  @ LLVM
  %8 = call i32 @puts(i8* %7)	@ "main"
	  %4 = call i32 (i8*, ...) @printf(i8* %3)	@ "main"
		NoModRef  @ SVF
		ModRef  @ LLVM
	  %6 = call i32 (i8*, ...) @scanf(i8* %5, i32 undef)	@ "main"
		NoModRef  @ SVF
		ModRef  @ LLVM
```

yuleisui:
There should be no modification side effects but only read side effects

FrancisYL:
I haven't inspected that deep into the LLVM codebase, but I think LLVM might leverage the behavior of external calls by function attributes. And I find the following pairs (non-I/O external function versus I/O external function) of function interesting
```LLVM
; Function Attrs: nounwind readonly
declare dso_local i64 @atol(i8*) #1

declare dso_local i32 @printf(i8*, ...) #4
```
It make sense to me that `atol` and `printf` will read its pointer argument, but only `atol` is attributed with `readonly` but I/O function `printf` is not.

```LLVM
; Function Attrs: nounwind readnone speculatable
declare float @llvm.floor.f32(float) #6

declare dso_local i32 @putchar(i32) #4
```
Neither `floor` or `putchar` read their argument because they are all non-pointer types, but only `floor` is attributed with `readnone` but I/O function `putchar` is not.

It must be some reason that these I/O functions are not attributed with `readonly` or `readnone` and I think it's because they all modify the underlying I/O buffer in memory.

FrancisYL:
In terms of the problem that add IO object in PAG would increase cause extra time solving Andersen. Would it be possible to add an extra check for external callsites who performs I/O operations in
https://github.com/SVF-tools/SVF/blob/f06eba337a17954981c6a470bcd273607c992f06/lib/MSSA/MemRegion.cpp#L186
So that the I/O buffer object will be added to the mod set instead of be added in the PAG?

yuleisui:
Adding IO objects will introduce many false dependencies since the underlying IO objects are not supposed to be in the application code. The side effect of an external call in SVF is determined by `ExtAPI.cpp`. For your `printf`,  there is only READ side effect but no MODIFICATION side effect.

Adding IO objects is similar to our previous experience in constant objects. But it becomes more heavyweight since the object can also be modified and not immutable according to your assumption, which adding more unnecessary overhead to the points-to analysis, both flow-insensitive and flow-sensitive (using SVFG). You can have a try for `-ander` and `-fspta` for a  larger size program, say 500KLOC benchmark and let me know whether it will be significantly slower.

yuleisui:
I am not trying to demotivate you, instead, I think you might wish to consider making changes to the query function or modify you query in your client:
https://github.com/SVF-tools/SVF/blob/f06eba337a17954981c6a470bcd273607c992f06/lib/WPA/WPAPass.cpp#L230

Adding IO objects will change the SVFG, and thus affecting later bug detections on top of the value-flows. I would suggest making a change to the READ side effect only as per my previous 3 steps. Let me know if you can think of a better solution.

FrancisYL:
Thanks for the clarification and suggestions! I now get more insight into SVF's design through this conversation. I'm now closing this pull request.

PR #202, opened 2020-04-20
Refine getRefInfoForCall()

FrancisYL:
Thanks for adding these APIs, which are really helpful. I don't consider writes (mod) to a memory location should be added to the ref set so I'm removing it.


yuleisui:
Hi @FrancisYL, would you mind taking a look at the travis for SVF's continuous integration build. The travis (https://github.com/SVF-tools/SVF/blob/master/.travis.yml) was contributed by someone else and I am not an expert on travis. If you have time, would you mind help fixing this broken build?

FrancisYL:
I'm not an expert on that either, unfortunately. However, I'm glad to help and dig that sometimes in the future.

PR #203, opened 2020-04-22
Add reachability check between SVFFunctions through PTACallGraph

FrancisYL:
A generic reachability check between SVFFunctions is useful for my project and I assume it's worth adding to the public API for PTACallGraph and PTACallGraphNode.

### DIFF
--- a/lib/Graphs/PTACallGraph.cpp
+++ b/lib/Graphs/PTACallGraph.cpp
@@ -235,6 +235,35 @@ void PTACallGraph::verifyCallGraph()
 }
 
 /*!
+ * Whether its reachable between two functions
+ */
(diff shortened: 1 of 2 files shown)


lib/Graphs/PTACallGraph.cpp
```diff
@@ -56,6 +56,29 @@ void PTACallGraphEdge::addInDirectCallSite(const CallBlockNode* call) {
 }
 //@}
 
+bool PTACallGraphNode::isReachableFromFunction(const SVFFunction *fun) const {
```
yuleisui, line 59: Would you be able to merge this code into`isReachableBetweenFunctions `? It seems redundant to make two functions.

FrancisYL:
Pushed, please take a look and let me know if there's any other comment.

PR #206, opened 2020-04-28
Implement Flow-Sensitive Type-Based Heap Cloning

mbarbar:
A large PR, however most changes should be self-contained to FSTBHC, and few affect other infrastructure (CHG, FSPTA, ...).

### DIFF
--- a/include/Graphs/PAG.h
+++ b/include/Graphs/PAG.h
@@ -46,7 +46,7 @@ public:
     typedef std::set<const CallBlockNode*> CallSiteSet;
     typedef std::map<const CallBlockNode*,NodeID> CallSiteToFunPtrMap;
     typedef std::map<NodeID,CallSiteSet> FunPtrToCallSitesMap;
-    typedef llvm::DenseMap<NodeID,NodeBS> MemObjToFieldsMap;
+    typedef DenseMap<NodeID,NodeBS> MemObjToFieldsMap;
     typedef std::set<const PAGEdge*> PAGEdgeSet;
     typedef std::list<const PAGEdge*> PAGEdgeList;
     typedef std::list<const PAGNode*> PAGNodeList;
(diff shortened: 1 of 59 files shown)


include/Graphs/PAG.h
```diff
@@ -607,6 +607,18 @@ class PAG : public GenericGraph<PAGNode,PAGEdge> {
     inline const MemObj* addDummyMemObj(NodeID i, const Type* type) {
         return SymbolTableInfo::Symbolnfo()->createDummyObj(i,type);
     }
+    inline NodeID addCloneDummyObjNode(const MemObj *mem) {
```
yuleisui, line 610: Could you please put these three functions in your TBHC.h/cpp, so that it will not be misused by other users?

yuleisui:
An error occurs when trying to use Andersen for a C++ program without meta info:

`wpa -ander example.bc`

`Assertion failed: (md != nullptr && "Missing type metadata at virtual callsite"), function getCSStaticType, file /Users/yulei/Downloads/svf-public/include/SVF-FE/DCHG.h, line 365.`

mbarbar:
Could you share the C++ code and how it was compiled? I cannot seem to reproduce this. I suspect I know what could possibly be happening but I can't confirm it.

lib/MemoryModel/PointerAnalysisImpl.cpp
```diff
@@ -219,7 +220,12 @@ void BVDataPTAImpl::dumpAllPts() {
 void BVDataPTAImpl::onTheFlyCallGraphSolve(const CallSiteToFunPtrMap& callsites, CallEdgeMap& newEdges) {
     for(CallSiteToFunPtrMap::const_iterator iter = callsites.begin(), eiter = callsites.end(); iter!=eiter; ++iter) {
         const CallBlockNode* cs = iter->first;
-        if (isVirtualCallSite(cs->getCallSite())) {
+
+        // If we have a DCHG (i.e., ctir available), use that to check for virtualness.
+        bool isVirtual = LLVMModuleSet::getLLVMModuleSet()->allCTir() ?
```
yuleisui, line 225: Can we move this into method 'isVirtualCallSite'?

mbarbar: isVirtualCallsite would need access to the chg, which would require another argument. Is this desired?

mbarbar: > Maybe we can put `isVirtualCallSite` in the `CommonCHGraph` and move CPPUtil's `isVirtualCallSite` to `CHA`

There is a call to `cppUtil::isVirtualCallSite` in `DDAClient.cpp`. It seems to me there is no `CHGraph`. Is that right or am I missing something?

yuleisui: There are few places invoking `isVirtualCallSite`, that is why we need to make a single wrapper for it, otherwise, it may cause crashes in other places.

Or we can move DCHG's `isVirtualCallSite` into `cppUtil::isVirtualCallSite`?

mbarbar: > Or we can move DCHG's `isVirtualCallSite` into `cppUtil::isVirtualCallSite`?

This seems to be the best way.

Issue #212, opened 2020-05-19
SVFG & strcpy


swthorn:
How are the SVFG and PAG graphs structured differently for memcpy or strcpy functions? From my testing, it appears as though values flowing through a strcpy function are not correctly propagated to the correct destination in the SVFG. If this is a known limitation already, sorry for bringing it up!

Check this gist: https://gist.github.com/swthorn/d53978bcf1e6180dd3b06bde76141aa2

The graph for test1.cpp is (some parts cut out for brevity):
![image](https://user-images.githubusercontent.com/62764299/82370248-41947e80-99e6-11ea-9aa5-341d8a5ef5a9.png)

We can see that the getelementptr instruction is propagated through by an llvm memcpy call all the way back to the main function.

For test2.cpp, there is no connection between the getelementptr of the global string variable and variables in the main function:
![image](https://user-images.githubusercontent.com/62764299/82370428-86201a00-99e6-11ea-8901-242b100aa908.png)

I'm wondering if you could possibly enlighten me if I'm just interpreting the graphs incorrectly? I was testing with a program propagating strings using strcpy and was really confused as to why things were not working as I had expected them to. When I took my test program and swapped out the strcpy's to directly write the string, everything worked.

Otherwise thanks! The framework is pretty fun to play around with!

yuleisui:
Thanks for reporting this. SVF handles external APIs using a lightweight summary.

The summary for 'strcpy' (also strcat and stpcpy) are incorrect (https://github.com/SVF-tools/SVF/blob/master/lib/Util/ExtAPI.cpp#L702). They should be placed next to https://github.com/SVF-tools/SVF/blob/master/lib/Util/ExtAPI.cpp#L736 with the side-effect of EFT_L_A0__A0R_A1R.

Would you mind making this change and let me know whether the graphs are the same?

swthorn:
Yes, that did appear to fix the issue, great!

Would there be anymore of these string functions that need changing, perhaps memset, strncpy and strncat?

Or perhaps with {"_ZNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEaSERKS4_", ExtAPI::CPP_EFT_A0R_A1R}, // c++11 

These functions may have changed their mangled names in C++14 (which SVF is now using)?

I can turn this into a pull request if I know which functions to change.

yuleisui:
[...]

PR #218, opened 2020-05-22
Update some string function's pass-through rules.

swthorn:
#212


yuleisui:
Can you also change the summary of strncpy and strncat?

swthorn:
Done!

PR #231, opened 2020-05-29
Runtime exception on clean up

swthorn:
I'm not sure what caused this exception to occur in some cases, but it always occurs during the cleanup of SVF for me now. I think it's related to the nodes in the SVFG that I consider "sources" during a source-sink analysis. I haven't dug that deep into it but this is a fix regardless.

image


yuleisui:
Is this the fix for error when running SABER?

swthorn:
Nope, I'm using SVF as an external library.

yuleisui:
We recently would like to promote this type of usage of SVF.  May I know how did you use SVF as an external library? Did you follow this?
https://github.com/SVF-tools/SVF/wiki/Using-SVF-as-a-lib-in-your-own-tool

PR #234, opened 2020-06-08
Account for ctir in build.sh

mbarbar:
As title.

I have also made a few changes:
* Fewer if-else (found it easier to modify/extend this way). There's still some room for improvement perhaps.
* Test against `$XYZ_DIR`, not `$XYZHome`. This way the user can choose a different directory (my `$LLVM_DIR` is set but it still downloaded and set the Ubuntu LLVM build which does not work on my machine, for example).
* Have not tested for Mac, as I am unable to (including the macOS ctir build). Perhaps CI will catch any such errors once the Test-Suite PR is accepted.


mbarbar:
I've also pushed a bug fix. I hadn't noticed that this wasn't merged but I think it'll be fine if CI passes. The CI seems to fail to compile some tests, so I'll take a look at that.

mbarbar:
The test failure is not to do with the test itself but something in the construction.
```
wpa: /home/mbarbar/clone/svf-public/include/Graphs/VFG.h:258: void VFG::checkIntraEdgeParents(const VFGNode*, const VFGNode*): Assertion `(srcfun == dstfun) && "src and dst nodes of an intra VFG edge are not in the same function?"' failed.
```

The commit that introduces the problem is: `815f09cca48691d39e949bef6a59abafdd8e46bd`.

I'm not quite sure of the problem. This example reproduces it:
```
int *global = nullptr;

class A {
public:
    int x;

    A() {
        foo();
    }

    virtual void foo() {
    }
};

class B : public A {
public:
    B() { }

    // Never called.
    virtual void foo() override {
        global = nullptr;
    }
};

int main(void) {
    B *b = new B();
}
```

yuleisui:
Just fixed the assertion. This is because the commit does not distinguish the two stores (global = nullptr;) in the PAG. It is now back to normal to distinguish these two store PAGEdges for flow-sensitive analysis.

https://github.com/SVF-tools/SVF/commit/cd8a169c2952846d99ed515a5a8c3fe77c7e80a1

mbarbar:
Great, thanks, will give it a go.

PR #256, opened 2020-07-05
Move some std::map/set to DenseMap/Set.

mbarbar:
This affects performance. Some basic examples on my local machine (right column is with `std::*`, left is with `Dense*`):
```
mruby
 Ander: 15.70  26.10
 SVFG : 00.94  00.97
 FS   : 31.05  39.89

du
 Ander: 02.36  03.53
 SVFG : 00.13  00.14
 FS   : 15.07  16.59

```

Some things of note:
* I didn't change everything, just most of what can be changed in WPA and related graphs without too much effort (so no template specialisation as necessary by `Dense*`).
* `NodeToPTSSMap` is specifically unchanged because (I believe) it is being used in such a way that would invalidate a `DenseMap` iterator and cause seg faults (in `MRGenerator::CollectPtsChain`).
* `std::map::[]` still takes a lot of time. I believe the main culprit is `PTSMap`. It would make a big difference to move `PTSMap` to be a `DenseMap`, but it would take some effort; needs some template specialisation.
* Getting the clock takes **a lot** more time than it should. Perhaps an option to only time entire analyses (i.e. `*::analyze`) rather than each operation could be useful.
* Memory doesn't seem too affected (looks better, in fact) but the testing I have done wasn't too extensive.

### DIFF
--- a/include/DDA/DDAPass.h
+++ b/include/DDA/DDAPass.h
@@ -25,7 +25,7 @@ public:
     /// Pass ID
     static char ID;
     typedef SCCDetection<SVFG*> SVFGSCC;
-    typedef std::set<const SVFGEdge*> SVFGEdgeSet;
+    typedef DenseSet<const SVFGEdge*> SVFGEdgeSet;
     typedef std::vector<PointerAnalysis*> PTAVector;
 
     DDAPass() : ModulePass(ID), _pta(NULL), _client(NULL) {}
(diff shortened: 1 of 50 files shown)


yuleisui:
I have just made the input of each pointer analysis using PAG instead of LLVMModule. The change is not much but might affect several files in WPA. I will update that first, could you make a pull request after that?

mbarbar:
Yes sure, I'll take a look tomorrow (if it's ready) and merge.

yuleisui:
Just done. Please check whether there is a conflict

mbarbar:
Merged fine and CI looks happy.

PR #261, opened 2020-07-13
Various fixes

gerion0:
This are some unrelated changes grouped in a single pull request. I can split it, if you want.

The changes are:
- Build with debug options: I'm not sure if this is the correct way. I'm sure CMake offers a better way but I'm have no clue of CMake.
- Enable analyses in 32 bit: The code in MemModel currently creates one APInt instance with hardcoded 64 bit. In our analyses this has failed for some 32 bit programs.
- Doxygen file: There is some online documentation. However, I have not found a way to generate it locally, so this commit adds a simple doxygen file for generating html documentation.
- Compiler warnings in headers: When using SVF as a library, it produces a lot of compiler warning, when using `-Wall -Wextra -Wpedantic`. This commits fixes a few of them.


yuleisui:
Thanks and looks good to me

PR #263, opened 2020-07-14
Fixed NPM install failed binding

AjaniBilby:
When installing via NPM the path bindings were volatile and disappeared after one power cycle.
Hence they are now bound via NPM for NPM installs.

The creation of the 'bind' folder is purely due to the fact that the NPM bindings need to exist when the package is initially downloaded, and thus also before the library is built locally. The NodeJS files purely just act as proxies for the real executables, and completely mimic the behaviour of executing them directly.

**Testing:**
Only done minor testing with WSL, and no full installs of Linux or other operating systems

**Side-note:**
Using the NPM install methods requires the flag ``--unsafe-perm`` due to the scripts required to be executed post-install.


AjaniBilby:
For some reason when installing via NPM it doesn't bind correctly. I can't work out why because it runs the correct shell scripts, with sudo perms, but the commands do not appear in /usr/bin/ after installation. Hence why I implemented NPM's binding method for NPM installations.

yuleisui:
You will need to export the path of SVF's executables, see below:

https://github.com/SVF-tools/SVF/blob/master/setup.sh#L39-L40

AjaniBilby:
The problem is that NPM tries to set up the path bindings before it runs the post-install scripts. So the files do not yet exist and it stops at that point with an error message.
Upon much further exploration seems like the core problem is actually that your build/setup scripts presume certain applications are already installed an configured within a certain way for them to execute correctly.

I will wipe my Linux install and start again from fresh to test if I can solve these issues via pre-install setups and if so I will close this merge and update the setup guide accordingly.

AjaniBilby:
I am now very confused because during the post-install process I am now getting a hang, where nothing appears to happen and it just sits at the post-test screen not making any progress, and it also has not registered the ``wpa`` command in the system path...

Below is a cut down version of the log, and below that, there is a link to the full output log
```
> ./build.sh & ./setup.sh

Setting up environment for SVF
LLVM_DIR = /mnt/c/Users/aj/Documents/GitHub/SVF/llvm-10.0.0.obj
Z3_DIR = /mnt/c/Users/aj/Documents/GitHub/SVF/z3.obj
./setup.sh: 30: [[: not found
+ svf-tools@1.0.1
added 1 package in 0.149s

...

Setting up environment for SVF
LLVM_DIR = /mnt/c/Users/aj/Documents/GitHub/SVF/llvm-10.0.0.obj
Z3_DIR = /mnt/c/Users/aj/Documents/GitHub/SVF/z3.obj
Test project /mnt/c/Users/aj/Documents/GitHub/SVF/Release-build

...

100% tests passed, 0 tests failed out of 114

Total Test time (real) =   3.49 sec
```
[Full log](https://gist.github.com/AjaniBilby/1a9782b4cad99f026a7dcb0bc64702c1)

yuleisui:
Binding is done by `source ./setup.sh`. Your pull request should not add any more *.js files. Can you do it via only `package.json`

AjaniBilby:
> Binding is done by `source ./setup.sh`. Your pull request should not add any more *.js files. Can you do it via only `package.json`

``source`` was the keyword that was missing in the original package.json as well. I have now added that and it works.
Also, note that my changes to ``setup.sh``. Without them, Ubuntu 20.04 LTS would always hit an error at the last if statement

yuleisui:
Your changes to `./setup.sh` looks strange to me. It does not make any difference. Can you try the original one again using `source`? In fact, you only need `source ./build.sh`, no need to `source ./setup.sh` since it will be invoked by `./build.sh`.

Try only `source ./build.sh` and see how it goes.

AjaniBilby:
I was too eager with my statement that it works with the ``source`` prefix...
When I have the source prefix, it will work for that exact terminal instances. If you open another instance ``wpa`` will not be bound, and if you restart the instance the bound it, it also becomes unbound.

When I revert my chances to ``setup.sh`` and run it with the ``source`` prefix, it does appear to execute, however it does not bind ``wpa`` even for the current terminal instance.

yuleisui:
Can you make a single-commit pull request with only a minimal change to `package.json` and submit it again?

It is now very messy.

AjaniBilby:
I'd need to delete my fork, then refork it to do that. And I'm not sure of the potential side effects.
Will this entire pull-request be deleted if I do that?

The only other option I see is to revert every single one of my commits on my fork except for the package change.

Issue #269, opened 2020-07-16
Segfault in PAGBuilder::build()


gerion0:
I'am experiencing a segfault within the construction of the PAG.

The relevant code snippet, that triggers the error is:
```
		SVFModule* svfModule = LLVMModuleSet::getLLVMModuleSet()->buildSVFModule(get_module());
		assert(svfModule != nullptr && "SVF Module is null");

		PAGBuilder builder;
		PAG* pag = builder.build(svfModule);
		assert(pag != nullptr && "SVF Module is null");
```
The error itself is:
```
Thread 1 "python3.7" received signal SIGSEGV, Segmentation fault.
ICFGBuilder::build (this=this@entry=0x7fffffffb8a0, svfModule=0x5555561a4a00) at ../subprojects/svf/lib/SVF-FE/ICFGBuilder.cpp:44
44              const SVFFunction *fun = *iter;
```
The backtrace is:
```
#0  ICFGBuilder::build(SVFModule*) (this=this@entry=0x7fffffffb8a0, svfModule=0x5555561a4a00) at ../subprojects/svf/lib/SVF-FE/ICFGBuilder.cpp:44
#1  0x00007fffee1e1e47 in PAG::PAG(bool) (this=0x55555772a6e0, buildFromFile=<optimized out>) at ../subprojects/svf/include/SVF-FE/SymbolTableInfo.h:144
#2  0x00007fffee19b62f in PAG::getPAG(bool) (buildFromFile=false) at ../subprojects/svf/include/Graphs/PAG.h:145
#3  PAGBuilder::PAGBuilder() (this=0x7fffffffb930) at ../subprojects/svf/include/SVF-FE/PAGBuilder.h:53
```
Here are the relevant code lines (it could be that some line numbers are different due to local asserts etc.):
`SymbolTableInfo.h:144` is the `return mod;` line in
```
    /// Module
    inline SVFModule* getModule()
    {
        return mod;
    }
```

`PAG.h:145` is the `new PAG(...)` line in:
```
    static inline PAG* getPAG(bool buildFromFile = false)
    {
        if (pag == NULL)
        {
            pag = new PAG(buildFromFile);
        }
        return pag;
    }
```

`PAGBuilder.h:53` is the constructor:
```
    PAGBuilder(): pag(PAG::getPAG()), svfMod(NULL), curBB(NULL),curVal(NULL)
    {
    }
```

Further debugging of frame 0 leads to:
```
(gdb) print svfModule
[...]

PR #273, opened 2020-07-24
Various Fixes

### LINKED ISSUE #260: Possible use after free in LLVMModuleSet, unwanted deconstruction of llvm::Module
Maybe I am wrong, but I think there is an invalid free of the LLVM module. I'm referring to [this](https://github.com/SVF-tools/SVF/blob/master/lib/SVF-FE/LLVMModule.cpp#L69) line:
```
modules[0] = std::unique_ptr<Module>(&mod);
```
Here the reference (i.e. pointer) to the LLVM module is given an unique_ptr. As soon as the unique_ptr goes out of scope it deconstructs the LLVM module. Since LLVMModuleSet is a singleton this has to be triggered by the developer, but e.g. this code should result in a memleak:
```
void foo(llvm::Module& my_mod) {
    	SVFModule* svfModule = LLVMModuleSet::getLLVMModuleSet()->buildSVFModule(my_mod);
        do_stuff_with_svfModule(svfModule);
        LLVMModuleSet::releaseLLVMModuleSet;
}
```
Maybe it is meaningful to use something like `std::vector<std::variant<std::unique_ptr<llvm::Module>, llvm::Module*>>` as internal data structure.

gerion0:
Some unrelated changes:
- Some fixes of compiler warnings
- Include LLVM with `-isystem` to suppress header warnings
- ICFG: ignore all intrinsics like `lllvm.donothing` not only debug ones
- LLVMModule: change memory handling and unify calls to buildSVFModule, fixes #260 and workaround #269 
- allow building of shared libraries
- ICFG: allow unimplemented functions (that don't have basic blocks)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,7 +24,7 @@ if(NOT COMMAND add_llvm_library)
         include(AddLLVM)
 
         add_definitions(${LLVM_DEFINITIONS})
-        include_directories(${LLVM_INCLUDE_DIRS})
+        include_directories(SYSTEM ${LLVM_INCLUDE_DIRS})
 
     else()
         message(FATAL_ERROR "\
--- a/include/DDA/DDAClient.h
+++ b/include/DDA/DDAClient.h
@@ -27,7 +27,7 @@ public:
 
     virtual ~DDAClient() {}
 
-    virtual inline void initialise(SVFModule* module) {}
+    virtual inline void initialise(SVFModule*) {}
 
     /// Collect candidate pointers for query.
     virtual inline NodeSet& collectCandidateQueries(PAG* p)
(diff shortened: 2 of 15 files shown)


yuleisui:
CI failed and segment fault. Could you have a look?

gerion0:
The tests passes for me with LLVM 9 on my local machine. I will investigate whether this is different with LLVM 10.

gerion0:
Sorry, I can't reproduce it. I have tested also with LLVM 10 from my distribution and all tests passed. I have set `SVF_CTIR=1` and `SVF_Z3=1` like in the travis script and run `build.sh` and all tests passed.

Do you have any suggestions, how to find the cause of the segfault?

yuleisui:
I would suggest break the pull request into smaller ones, 1. remove header file warnings, 2. Cmakelist, and 3. LLVMModule and LLVMUtil.  I guess it might caused by 3. You can check the CI later to see this
https://travis-ci.org/github/SVF-tools/SVF/builds/711625645

gerion0:
Closed in advance of #275 #276 #277 and #278.

PR #277, opened 2020-07-26
LLVMModule: fix memory handling

gerion0:
It does the following:
- module references are not converted to unique_ptr anymore, thus fixing
  a possible use after free
- the cxts pointer is freed, fixing a memleak
- simplify handling of for loops
- unify handling of buildSVFModule(namevector) and buildSVFModule(module)
  As a side effect the SymbolTableInfo is always generated, which works
  around a segfault, see #269

Fixes: #260
Workaround: #269

### DIFF
--- a/include/SVF-FE/LLVMModule.h
+++ b/include/SVF-FE/LLVMModule.h
@@ -60,6 +60,8 @@ private:
     /// Constructor
     LLVMModuleSet(): svfModule(nullptr), moduleNum(0), cxts(nullptr), modules(NULL) {}
 
+    void build(const std::vector<std::string> &moduleNameVec);
+
 public:
     static inline LLVMModuleSet *getLLVMModuleSet()
     {
@@ -77,8 +79,6 @@ public:
 
     SVFModule* buildSVFModule(Module &mod);
     SVFModule* buildSVFModule(const std::vector<std::string> &moduleNameVec);
-
-    void build(const std::vector<std::string> &moduleNameVec);
 
     u32_t getModuleNum() const
     {
--- a/include/SVF-FE/LLVMModule.h
+++ b/include/SVF-FE/LLVMModule.h
@@ -47,7 +47,8 @@ private:
     static LLVMModuleSet *llvmModuleSet;
     SVFModule* svfModule;
     std::unique_ptr<LLVMContext> cxts;
-    std::vector<std::unique_ptr<Module>> modules;
+    std::vector<std::unique_ptr<Module>> owned_modules;
+    std::vector<std::reference_wrapper<Module>> modules;
 
     /// Function declaration to function definition map
     FunDeclToDefMapTy FunDeclToDefMap;
@@ -92,7 +93,7 @@ public:
     Module &getModuleRef(u32_t idx) const
     {
         assert(idx < getModuleNum() && "Out of range.");
-        return *(modules[idx].get());
+        return modules[idx];
     }
 
     // Dump modules to files
--- a/include/SVF-FE/LLVMModule.h
+++ b/include/SVF-FE/LLVMModule.h
@@ -86,8 +86,7 @@ public:
 
     Module *getModule(u32_t idx) const
     {
-        assert(idx < getModuleNum() && "Out of range.");
-        return modules[idx].get();
+        return &getModuleRef(idx);
     }
 
     Module &getModuleRef(u32_t idx) const
--- a/include/SVF-FE/LLVMModule.h
+++ b/include/SVF-FE/LLVMModule.h
@@ -48,7 +48,7 @@ private:
     SVFModule* svfModule;
     std::unique_ptr<LLVMContext> cxts;
     u32_t moduleNum;
-    std::unique_ptr<Module> *modules;
+    std::vector<std::unique_ptr<Module>> modules;
 
     /// Function declaration to function definition map
     FunDeclToDefMapTy FunDeclToDefMap;
@@ -58,7 +58,7 @@ private:
     GlobalDefToRepMapTy GlobalDefToRepMap;
 
     /// Constructor
-    LLVMModuleSet(): svfModule(nullptr), moduleNum(0), cxts(nullptr), modules(NULL) {}
+    LLVMModuleSet(): svfModule(nullptr), moduleNum(0), cxts(nullptr) {}
 
     void build(const std::vector<std::string> &moduleNameVec);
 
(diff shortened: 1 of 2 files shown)


yuleisui:
Your changes seem significant. I am unsure why would like to modify LLVMModule. The previous implementation works well and I do not think the memory leak is a big issue. Would you like to break down this patch a bit for me to understand: (1) pure refactoring, and (2) your fix for bugs

gerion0:
Splitting the commit is really hard, since the refactoring is based on the changed datastructures, but I will try to explain my motivations.

## Issues, that I want to fix

### Memory handling
Let me create a minimal example:

```
std::vector<std::string> get_module_names();
llvm::Module& get_module();

int main1() {
    auto module_names = get_module_names();
    SVFModule* svfModule = LLVMModuleSet::getLLVMModuleSet()->buildSVFModule(module_names);
    LLVMModuleSet::releaseLLVMModuleSet;
}

int main2() {
    llvm::Module& module = get_module();
    SVFModule* svfModule = LLVMModuleSet::getLLVMModuleSet()->buildSVFModule(module);
    LLVMModuleSet::releaseLLVMModuleSet;
}
```

The current memory handling for `main1` works as following:
1. `getLLVMModuleSet` [allocates](https://github.com/SVF-tools/SVF/blob/master/include/SVF-FE/LLVMModule.h#L67) memory for an `LLVMModuleSet` (`llvmModuleSet = new LLVMModuleSet();`).
2. `buildSVFModule` then leads to the IR parsing and `llvm::Module` construction. LLVM hands back the `llvm::Module`s as `unique_ptr` and therefore also handing back the control about the memory of the modules. SVF then [allocates](https://github.com/SVF-tools/SVF/blob/master/lib/SVF-FE/LLVMModule.cpp#L139) an `LLVMContext` (`cxts = new LLVMContext[1];`) and an module array (`modules = new unique_ptr<Module>[moduleNum];`) where it stores the pointers in.
3. `releaseLLVMModuleSet` is then called, causing the `llvmModuleSet` object to be descructed and the memory freed. Since a descructor of `LLVMModuleSet` is missing the memory for `cxts` and `modules` is never freed, leaving all LLVM modules behind.

This is somewhat different for `main2`:
1. Same as above
2. `buildSVFModule` assigns an already existing `LLVMContext` pointer to `ctxs` and then [constructs](https://github.com/SVF-tools/SVF/blob/master/lib/SVF-FE/LLVMModule.cpp#L68) an `unique_ptr` array (`modules = new unique_ptr<Module>[moduleNum];`). After that, it overgives the `llvm::Module&` to an `unique_ptr` and moves this `unique_ptr` into the array (`modules[0] = std::unique_ptr<Module>(&mod);`). Deconstructing of the array would therefore also lead to the deconstruction of the module reference, which is probably unwanted.
3. Same as above. Does not free all memory, therefore not causing any harm with an unwanted destruction.

The quick fix would be to define a constructor:
```
~LLVMModuleSet() {
    delete cxts[0];
    for (unsigned i = 0; i < moduleNum; ++i) {
        delete modules[i];
    delete modules;
}
```
However, this does not work for `cxts` since it is allocated in `main1` but not in `main2` and lead to the unwanted descruction of the `llvm::Module` in `main2`.

This is partly also explained in #260.

### Missing SVFModule on PAG construction, if `buildSVFModule(llvm::Module&)` is used

This is explained in #269. Here is a minimal example again:
```
llvm::Module& get_module();

int main() {
    llvm::Module& module = get_module();
    SVFModule* svfModule = LLVMModuleSet::getLLVMModuleSet()->buildSVFModule(module);
    PAGBuilder builder;
    builder->build(svfModule);
}
```

The short explanation is: The construction of `PAGBuilder` already constructs a `PAG`, _before_ `PAGBuilder::build()` is invoked.
Since `buildSVFModule(Module&)` is used, a `SymbolTableInfo` is _not_ initialized.
However, the construction of the `PAG` with `getPag()` in the constructor of `PAGBuilder` needs an `SVFModule`.
Normally, this `SVFModule` is taken from the `SymbolTableInfo`.
Since the `SymbolTableInfo` is not initialized, an `SVFModule` cannot be found, resulting in a segfault at the PAG construction.

## How I try to fix it.

### Fixing the memory handling
There are two paths to an `SVFModule`: 1. `buildSVFModule(Module&)` and 2. `buildSVFModule(std::vector<std::string>)`.
In the first case all memory is already allocated. The SVF only needs to store the references.
In the second case the memory is allocated and given the SVF to manage.

Therefore I have introduced two new datastructures:
1. `modules`: a `std::vector<std::reference_wrapper<llvm::Module>>` to store the references.
2. `owned_modules`: a std::vector<std::unique_ptr<llvm:Module>> to store the owned unique_ptrs.

All operations are done on the reference vector, since it is existent in both paths.
If `buildSVFModule(std::vector<std::string>)` is used, the unique_ptr vector is filled.
It then manages the life times of `llvm::Modules`.

Since the vectors are member variables they are destructed automatically, when `LLVMModuleSet` is deleted.
The vectors take care of there inner data and deconstruct them as well and therefore fixing the above memleak.
The above problem of deleting the reference also does not occur, since the module reference is never converted to a `unique_ptr`.

### Fixing the `SymbolTableInfo` generation

I also tried to unify both paths so the `SymbolTableInfo` is always generated. Therefore I changed `buildSVFModule(std::vector<std::string>)` to only load the modules in the `modules` vector and then calling `build()`, which take care of the rest. `buildSVFModule(llvm::Module&)` does the same: It loads the one module into the `modules` vector and then calls `build()`.

yuleisui:
It is still quite complicated for me to understand. It would be good to break down this pull request further. I guess you can focus on functional requirement first (e.g., supporting LLVM Module) and then non-functional one (e.g., memory leaks)

gerion0:
The commit is now splitted. I hope, this helps,

yuleisui:
CI build failed. It seems `LLVM_VERSION_MAJOR >= 7` did not work.

gerion0:
I have tested locally with every single commit again with `.build.sh` and all checks passed. I will create some pull request to trigger the CI. With that I can simulate a binary search until I find the faulty commit.

gerion0:
Ok, commit 18c7785 fails. This was kind of expected, since it is the biggest change (data structure wise). However, I cannot reproduce it. Do you know a way, how to find out what is wrong?

yuleisui:
It is a bit hard to tell. You may wish to further split the patch to see. One possibility is the wrong use of `Module` or `unique pointer`?

gerion0:
I have found the error. The lines
```
owned_modules.emplace_back(std::move(mod));
modules.emplace_back(*mod);
```
was in the wrong order, since you must not dereference a moved unique_ptr.

Issue #272, opened 2020-07-23
Inconsistent flow-sensitive results: SVFG vs. optimised SVFG


mbarbar:
There are cases where the flow-sensitive results differ depending on whether the SVFG is optimised.

To test and see the differences, this shell script can help:
```
#!/bin/sh

echo -n "start OPT..."
./bin/wpa -dump-svfg -stat=false -fspta -print-pts -optSVFG=true $1 | ag 'nodeid' | sort > opt.txt
echo " done"
echo -n "start NOOPT..."
./bin/wpa -stat=false -fspta -print-pts -optSVFG=false $1 | ag 'nodeid' | sort > noopt.txt
echo " done"
diff opt.txt noopt.txt
```

Example 1:
```
; ModuleID = 'cons.cpp'
source_filename = "cons.cpp"
target datalayout = "e-m:e-p270:32:32-p271:32:32-p272:64:64-i64:64-f80:128-n8:16:32:64-S128"
target triple = "x86_64-unknown-linux-gnu"
%class.B = type { %class.A.base, [4 x i8] }
%class.A.base = type <{ i32 (...)**, i32 }>
%class.A = type <{ i32 (...)**, i32, [4 x i8] }>
$_ZN1BC2Ev = comdat any
$_ZN1AC2Ev = comdat any
$_ZN1B3fooEv = comdat any
$_ZN1A3fooEv = comdat any
$_ZTV1B = comdat any
$_ZTS1B = comdat any
$_ZTS1A = comdat any
$_ZTI1A = comdat any
$_ZTI1B = comdat any
$_ZTV1A = comdat any

@_ZTI1A = linkonce_odr dso_local constant { i8*, i8* } { i8* bitcast (i8** getelementptr inbounds (i8*, i8** null, i64 2) to i8*), i8* getelementptr inbounds ([3 x i8], [3 x i8]* null, i32 0, i32 0) }, comdat, align 8
@_ZTV1A = linkonce_odr dso_local unnamed_addr constant { [3 x i8*] } { [3 x i8*] [i8* null, i8* bitcast ({ i8*, i8* }* @_ZTI1A to i8*), i8* bitcast (void (%class.A*)* @_ZN1A3fooEv to i8*)] }, comdat, align 8

define dso_local i32 @main() #0 personality i8* bitcast (i32 (...)* @__gxx_personality_v0 to i8*) {
  %a4 = call i8* @_Znwm(i64 16) #5
[...]

PR #281, opened 2020-07-28
Connecting Formal/Actual Parameters/Returns

mbarbar:
This is for #272. For my test cases, it appears that optimised vs unoptimised are now giving the same results with these changes.

The functions which propagate aparm->fparm/fret->aret slipped my mind whilst making #272, so these nodes were definitely handled by FSPTA but they were being connected a bit weird. Before, for example, some formal parameters had incoming edges from non-actual parameters, I assume for optimisation's sake.

`mruby` also runs as expected, taking 20s vs. 34s for optimised vs. unoptimised.

### DIFF
--- a/include/Graphs/VFG.h
+++ b/include/Graphs/VFG.h
@@ -357,10 +357,10 @@ protected:
     /// Connect VFG nodes between caller and callee for indirect call site
     //@{
     /// Connect actual-param and formal param
-    virtual inline void connectAParamAndFParam(const PAGNode* cs_arg, const PAGNode* fun_arg, const CallBlockNode*, CallSiteID csId, VFGEdgeSetTy& edges)
+    virtual inline void connectAParamAndFParam(const PAGNode* cs_arg, const PAGNode* fun_arg, const CallBlockNode* cbn, CallSiteID csId, VFGEdgeSetTy& edges)
     {
-        NodeID actualParam = getDef(cs_arg);
-        NodeID formalParam = getDef(fun_arg);
+        NodeID actualParam = getActualParmVFGNode(cs_arg, cbn)->getId();
+        NodeID formalParam = getFormalParmVFGNode(fun_arg)->getId();
         VFGEdge* edge = addInterEdgeFromAPToFP(actualParam, formalParam,csId);
         if (edge != NULL)
             edges.insert(edge);
--- a/include/Graphs/VFG.h
+++ b/include/Graphs/VFG.h
@@ -357,19 +357,19 @@ protected:
     /// Connect VFG nodes between caller and callee for indirect call site
     //@{
     /// Connect actual-param and formal param
-    virtual inline void connectAParamAndFParam(const PAGNode* cs_arg, const PAGNode* fun_arg, const CallBlockNode* cbn, CallSiteID csId, VFGEdgeSetTy& edges)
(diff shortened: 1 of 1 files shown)


include/Graphs/VFG.h
```diff
@@ -357,19 +357,19 @@ class VFG : public GenericVFGTy
     /// Connect VFG nodes between caller and callee for indirect call site
     //@{
     /// Connect actual-param and formal param
-    virtual inline void connectAParamAndFParam(const PAGNode* cs_arg, const PAGNode* fun_arg, const CallBlockNode*, CallSiteID csId, VFGEdgeSetTy& edges)
+    virtual inline void connectAParamAndFParam(const PAGNode* csArg, const PAGNode* funArg, const CallBlockNode* cbn, CallSiteID csId, VFGEdgeSetTy& edges)
     {
-        NodeID actualParam = getDef(cs_arg);
-        NodeID formalParam = getDef(fun_arg);
+        NodeID actualParam = getActualParmVFGNode(csArg, cbn)->getId();
```
yuleisui, line 362: How about if the AcutalParmVFGNode was opt out (removed) during SVFGOPT?

yuleisui: I think this change will only work for indirect callsites but not normally direct callsites. We will need to add a conditional check.

mbarbar: > How about if the AcutalParmVFGNode was opt out (removed) during SVFGOPT?

For the first comment, this shouldn't be run by `SVFGOPT` because it overrides this function, so `SVFGOPT` will keep doing whatever it was doing before.

> I think this change will only work for indirect callsites but not normally direct callsites. We will need to add a conditional check.

Isn't this only called for indirect calls anyway on the fly?

The VFG construction looks okay to me, in that `getActualParmVFGNode` is used to make actual param -> formal param edges.
```
        else if(FormalParmVFGNode* formalParm = SVFUtil::dyn_cast<FormalParmVFGNode>(node))
        {
            for(CallPESet::const_iterator it = formalParm->callPEBegin(), eit = formalParm->callPEEnd();
                    it!=eit; ++it)
            {
                const CallBlockNode* cs = (*it)->getCallSite();
                ActualParmVFGNode* acutalParm = getActualParmVFGNode((*it)->getSrcNode(),cs);
                addInterEdgeFromAPToFP(acutalParm,formalParm,getCallSiteID(cs, formalParm->getFun()));
            }
        }
```

yuleisui: True. Only used when on the fly build. Could you add an assertion in `connectAParamAndFParam` to make sure the Formal and actual SVFGNodes do exist and have not been removed from SVFG? 

Though it fixed the problem, It still not quite clear why `getDef` won't work. (1) The formal parameter once removed will be replaced with a Phi node and its definition site will be redefined to point to that Phi (https://github.com/SVF-tools/SVF/blob/master/include/Graphs/SVFGOPT.h#L238) and (2) the actual parameter once removed will have the same value as its definition site too.

mbarbar: I'm not sure if such an assertion would be meaningful because SVFG node removal only removes the edges, not the actual node itself (assuming it even happens in `SVFG`).

>Though it fixed the problem ...

The changes should only apply to `SVFG` since `SVFGOPT` overrides this function with one which considers phi.

mbarbar: Actually, `getActualParmVFGNode` and the other 3 similar functions have an assert in them, so I think it would be unnecessary to assert.

Issue #268, opened 2020-07-16
Enhancement: Introduce an SVF namespace


gerion0:
Especially when using SVF as library, it would be nice if all symbols and preprocessor constants are prefixed with SVF, i.e. to use a namespace SVF for C++ itself and the prefix "SVF_" for preprocessor constansts.

For example, I currently include the SVF with the following lines:
```
#define VERSION_BKP VERSION
#undef VERSION
#include <Graphs/VFGNode.h>
#include <SVF-FE/PAGBuilder.h>
#include <Util/BasicTypes.h>
#include <WPA/Andersen.h>
#undef VERSION
#define VERSION VERSION_BKP
#undef VERSION_BKP
```

SVF-tools:
Good suggestion! Would you be able to help us with adding `SVF` as the namespace?

pzieris:
Hi guys, I'd be willing to help you with that. Especially in adding a C++ namespace. Please let me know how you'd like to proceed on this!

yuleisui:
Thanks, you could add a namespace called "SVF" for all the classes and methods. It would be good to do the minimum changes to the project.

pzieris:
Hi Yulei, by the way, while adding the namespace I noticed that at lot of cpp files have `using namespace SVFUtil` but then either don't require the namespace at all or still call methods through `SVFUtil::foo()`. If it's okay with you I would remove all those unnecessary `using namespace SVFUtil` to clean up the code a bit. What do you think?

yuleisui:
Sounds good to me. Please go ahead and thanks.

pzieris:
Hi again, I just started removing `using namespace SVFUtil` and adding `SVFUtil::` to functions where necessary, when I realized how many calls to functions from the SVFUtil namespace there actually are. So, to be clear, when I continue with this, there will be *a lot* of changes.

As an alternative, I thought about keeping `using namespace SVFUtil` but then removing the `SVFUtil::` everywhere to be consistent. However, after doing this for some files, compiling SVF gave me ambiguous calls to casting functions, e.g., to `SVFUtil::dyn_cast()` and `llvm::dyn_cast()`. As far as I see, the only option to continue with this alternative would be to remove `include/Util/Casting.h` altogether and instead rely on the LLVM casting functions. However, I cannot tell whether there is a reason why these casting functions are re-implemented in SVF?

Do you have any preference on how to proceed?

yuleisui:
[...]

PR #282, opened 2020-07-29
Add SVF namespace

pzieris:
Add SVF namespace for #268


yuleisui:
Thanks for your contribution. To validate the namespace, would you mind changing the SVF-example a bit to see whether it will pass its CI?

PR #298, opened 2020-08-17
Warning fixes

canlockkeyin:
This collection of commits fixes a lot of the warnings that arise when using SVF as a library with `-Wall -Wextra -Wpedantic`.

### DIFF
--- a/lib/Util/TypeBasedHeapCloning.cpp
+++ b/lib/Util/TypeBasedHeapCloning.cpp
@@ -231,7 +231,7 @@ const NodeBS TypeBasedHeapCloning::getGepObjClones(NodeID base, unsigned offset)
         }
 
         addGepToObj(newGep, base, totalOffset);
-        const DIType *newGepType;
+        const DIType *newGepType = nullptr;
         if (baseType->getTag() == dwarf::DW_TAG_array_type || baseType->getTag() == dwarf::DW_TAG_pointer_type)
         {
             if (const DICompositeType *arrayType = SVFUtil::dyn_cast<DICompositeType>(baseType))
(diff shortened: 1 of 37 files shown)


yuleisui:
I saw you had make quite a few changes on Cudd which is a third party library and officially not a part of SVF. If possible, could you submit a patch only fixing the warnings of the svf code but leave cudd code unchanged? I would prefer to change the compilation option in CMakelist.txt under cudd folder.

canlockkeyin:
I've removed the CUDD changes from the pull request, I hope this worked as intended.

lib/Util/TypeBasedHeapCloning.cpp
```diff
@@ -454,8 +455,9 @@ NodeID TypeBasedHeapCloning::cloneObject(NodeID o, const DIType *type, bool reus
         {
             clone = addCloneFIObjNode(fiObj->getMemObj());
         }
-        else if (const DummyObjPN *dummyObj = SVFUtil::dyn_cast<DummyObjPN>(obj))
+        else
```
yuleisui, line 458: This changes the underlying semantics.

yuleisui: May trigger a null pointer at line 451

canlockkeyin: > This changes the underlying semantics.

Can `isa<type>(obj)` be true, and `dyn_cast<type>(obj)` still fail? Otherwise I don't see how this would change anything, as we have only two possible cases, which should be covered by both `if {} else if {}` and `if {} else {}`.

yuleisui: make sense.

yuleisui:
Thanks and I will merge now

PR #303, opened 2020-08-25
build.sh: use ninja

gerion0:
This has the main advantage that ninja uses all cores per default resulting in a faster compilation on machines with more or less than 4 cores.

See #302 for more discussion.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,8 @@ env:
 
 before_script:
   - if [ "$TRAVIS_OS_NAME" = "linux" ]; then sudo apt-get update       ; fi
-  - if [ "$TRAVIS_OS_NAME" = "linux" ]; then sudo apt-get install cmake gcc g++ nodejs; fi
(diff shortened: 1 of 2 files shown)


mbarbar:
One concern with ninja for C++ projects is the high memory usage. For example, I have 8 GB of memory and 4 cores/8 threads on one of my machines and last I tried ninja, it using all available CPU caused the machine to halt from OOM. For make, I limit myself to `j4` on that machine. Not sure if there is a way to limit meory given to ninja.

gerion0:
`-j4` reduces the memory only as a side effect. The option actually limits the number of parallel jobs to 4 at maximum. Ninja has the exact same option, so `ninja -j4` should behave similar to `make -j4`. Maybe, it is a good idea to propagate this argument to the user? So `build.sh` e.g. uses all 32 cores by default (much faster given that enough RAM is available) but can be limited with `build.sh -jX` to use less cores.

mbarbar:
I wasn't aware controlling jobs was possible in `ninja`. I guess I never checked? Thanks haha. I like the idea of giving `build.sh` an optional `-j` argument.

PR #305, opened 2020-08-25
Cmake fixes

gerion0:
This pull request enables building of SVF with LLVM 10 on Gentoo (and probably on other systems). Gentoo changed its packaging of LLVM so that it only provides a big libLLVM-10.so. With this commits SVF can link to it again.

I have tested this PR against LLVM 9 and 10 with static and shared libraries.


gerion0:
Please do not merge yet. There is some problem when using this as meson subproject and linking against LLVM 9.

gerion0:
I think, I have fixed all errors, can be merged, if you agree.

Issue #306, opened 2020-08-27
wpa will Abort in VFG.h:409


tiedaoxiaotubie:
When I tried to execute `wpa -ander -svfg -dump-svfg libpng16.a.op`, an error will raise:
```
wpa: /home/SVF/include/Graphs/VFG.h:409: SVF::NodeID SVF::VFG::getDef(const SVF::PAGNode*) const: Assertion `it!=PAGNodeToDefMap.end() && "PAG node does not have a definition??"' failed.
Aborted (core dumped)
```
This error is completely same with https://github.com/SVF-tools/SVF/issues/296#issue-677353986

Compared with showcase in issue#296, the bc file is much more smaller this time:
[libpng.zip](https://github.com/SVF-tools/SVF/files/5136425/libpng.zip)

yuleisui:
@tiedaoxiaotubie can you try the latest version of SVF to see whether the issue is gone?

rainoftime:
At the latest commit, the bc triggers an issue
~~~~
Writing 'svfg_final.dot'...wpa: /home/legend/SVF/lib/Graphs/SVFG.cpp:1040: static std::string llvm::DOTGraphTraits<SVF::SVFG*>::getNodeAttributes(llvm::DOTGraphTraits<SVF::SVFG*>::NodeType*, SVF::SVFG*): Assertion `false && "no such kind of node!!"' failed.
~~~~

yuleisui:
Fixed.

Issue #296, opened 2020-08-12
fstbhc raise error


tiedaoxiaotubie:
I am testing different pointer analysis mode now, when trying `wpa --fstbhc --print-all-pts libxml.a.bc`, an error was raised:
```
wpa: /home/vultest/SVF/include/Graphs/VFG.h:409:  SVF::NodeID  SVF::VFG::getDef(const SVF::PAGNode*) const: Assertion `it!=PAGNodeToDefMap.end() && "PAG node does not have a definition??"` failed.
Aborted
```
The bc file was also attached in this issue :)

BTW: all pointer analysis mode were tested with this bc file, only `fstbhc` failed, which means the other modes worked well 👍 
[libxml2.a.bc.zip](https://github.com/SVF-tools/SVF/files/5060448/libxml2.a.bc.zip)

PR #312, opened 2020-09-04
Introduce handling of unary operations

jumaho:
Hi there,

while trying to analyze larger projects, including the SPEC CPU 2017 suite, we realized that some WPA analyses failed because SVF does not handle the `fneg` LLVM instruction or, more generally, does not handle unary operations at all, in contrast to binary operations. `fneg` is currently the only unary operation in LLVM, but the category might be extended in the future.

This patch adds support for unary operations to SVF. The support is implemented analogous to binary ops, in PAG, VFG, SVFG, etc. We tested the patch with our previously failing analyses on the SPEC CPU suite, which are now successfully running through.

The failing behavior we observed without this patch seems quite similar to #306 (and hence also #296), so this patch might also fix those.

### DIFF
--- a/include/Graphs/PAG.h
+++ b/include/Graphs/PAG.h
@@ -56,9 +56,11 @@ public:
     typedef std::vector<const PAGNode*> PAGNodeList;
     typedef std::vector<const CopyPE*> CopyPEList;
     typedef std::vector<const BinaryOPPE*> BinaryOPList;
+    typedef std::vector<const UnaryOPPE*> UnaryOPList;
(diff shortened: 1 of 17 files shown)


yuleisui:
Thanks for your contribution and it is really a good quality patch! In fact, the current SVF also did not handle LLVM's  `BranchInst` at which a conditional register value is used. It is similar to your unary operation and it does not modify or create any new value. I just mark it down here in case It might cause a similar issue. Of course, we would appreciate if you can also add that in if you have time.

Issue #330, opened 2020-09-30
DFPTData LLVM RTTI


mbarbar:
In refactoring the PTData classes, I was looking over the LLVM RTTI for them.

https://github.com/SVF-tools/SVF/blob/e5e780a46f62b53f1be611a260055fdf49db122f/include/MemoryModel/PointsToDFDS.h#L425-L428

This looks good.

https://github.com/SVF-tools/SVF/blob/e5e780a46f62b53f1be611a260055fdf49db122f/include/MemoryModel/PointsToDFDS.h#L429-L432

This looks correct but unnecessary and can be deleted.

https://github.com/SVF-tools/SVF/blob/e5e780a46f62b53f1be611a260055fdf49db122f/include/MemoryModel/PointsToDFDS.h#L433-L437

I believe this should be `return ptd->getPTDTY() == PTData<Key,Data>::IncDFPTD`.

The above 3 snippets are in `DFPTData`.

The code in the above snippet should replace the following (in `DFPTData`):
https://github.com/SVF-tools/SVF/blob/e5e780a46f62b53f1be611a260055fdf49db122f/include/MemoryModel/PointsToDFDS.h#L195-L198

Does all this sound correct?

yuleisui:
Sounds correct to me. Please submit a patch for the fix.

> In refactoring the PTData classes, I was looking over the LLVM RTTI for them.
> 
> https://github.com/SVF-tools/SVF/blob/e5e780a46f62b53f1be611a260055fdf49db122f/include/MemoryModel/PointsToDFDS.h#L425-L428
> 
> This looks good.
> 
> https://github.com/SVF-tools/SVF/blob/e5e780a46f62b53f1be611a260055fdf49db122f/include/MemoryModel/PointsToDFDS.h#L429-L432
> 
> This looks correct but unnecessary and can be deleted.
>
Not quite sure. Would this be necessary?

> https://github.com/SVF-tools/SVF/blob/e5e780a46f62b53f1be611a260055fdf49db122f/include/MemoryModel/PointsToDFDS.h#L433-L437
> 
> I believe this should be `return ptd->getPTDTY() == PTData<Key,Data>::IncDFPTD`.
>
Yes, please fix this.

> The above 3 snippets are in `DFPTData`.
> 
> The code in the above snippet should replace the following (in `DFPTData`):
> 
[...]

PR #331, opened 2020-10-03
Refactor PTData

mbarbar:
This refactors PTData to (almost) seamlessly allow for different "backends". `include/MemoryModel/AbstractPointsToDS.h` contains some abstract classes defining the interface most analyses will use (occasionally, they may need to access implementation-defined details), and `include/MemoryModel/MutablePointsToDS.h` implements those classes as what we had before.

Some stuff was converted to `const` and I introduced a few interfaces to get around returned points-to sets being `const`. In essence we can no longer do things like `getPts(x) |= otherPts` in analyses. Otherwise, interfaces are mostly unchanged.

It would be good I think to consider #330 as part of the review too.


yuleisui:
Most of the code looks good to me. Any particular reason why we need to return const pts from `getPts`?

mbarbar:
A few reasons:

1. It better separates workload (modifying points-to sets is `PTData`'s job *only*.
2. It allows for more implementation freedom in `PTData` if we can expect points-to sets to only be changed through specific interfaces.
3. Internally, real points-to sets may not be used (e.g. BDDs) and `getPts` is only providing a "view".

include/DDA/DDAVFSolver.h
```diff
@@ -525,11 +525,12 @@ class DDAVFSolver
     virtual inline CPtSet& getCachedPointsTo(const DPIm& dpm)
     {
         if (isTopLevelPtrStmt(dpm.getLoc()))
-            return getCachedTLPointsTo(dpm);
```
yuleisui, line 528: It seems to me that we can use `getCachedTLPointsTo ` here?

yuleisui (commented):
Only one line is confusing, otherwise, all good

include/DDA/DDAVFSolver.h
mbarbar: `getCachedPointsTo` is used non-`const`ly (to union points-to sets, etc.), and getCachedTLPointsTo in `FlowDDA` relies on `getPts`, so `getCachedTLPointsTo` needs to be `const`.

Another way to overcome this is to add functions into these classes which perform the union and make `getCachedPointsTo` `const`. For example, instead of:
```
    PointsTo &pt = getCachedPointsTo(dpm);
    pt |= someOtherPt;  // Or unionDDAPts(pt, someOtherPt), etc.
```

We would do:
```
    unionDDAPts(dpm, someOtherPt);  // Overload similar to unionPts in PTData
```

The only issue with this is that we would then internally need a non-`const` version of `getCachedPointsTo` for the overload to use. I'd have to study the code closer to think of a cleaner way unless you have some suggestions, I don't think I've really looked at DDA before.

mbarbar: In fact, it seems like we *need* a cleaner way, on second thought. I'll think about how to go about fixing it.

yuleisui:
I think the following should be fine by adding const to the return pts:
```
    virtual inline const CPtSet& getCachedPointsTo(const DPIm& dpm)
    {
        if (isTopLevelPtrStmt(dpm.getLoc()))
            return getCachedTLPointsTo(dpm);
        else
            return getCachedADPointsTo(dpm);
    }
```

yuleisui:
It seems the only problem is here when using a non-const pts:
```
    virtual inline void updateCachedPointsTo(const DPIm& dpm, CPtSet& pts)
    {
        CPtSet& dpmPts = getCachedPointsTo(dpm);
        if (unionDDAPts(dpmPts, pts){
            .....
        }
    }
```

yuleisui:
The following should work.
```
    virtual inline CPtSet& getCachedPointsTo(const DPIm& dpm)
    {
        if (isTopLevelPtrStmt(dpm.getLoc()))
            return dpmToTLCPtSetMap[dpm];
        else
            return dpmToADCPtSetMap[dpm];
    }
```

mbarbar:
Yes, this would work. The concern however is in derived classes: what if they override `getCachedTLPointsTo` (to use `getPts` for example), and presumably similar functions like `addPts` and whatnot, and they stop using `dpmtoTLCPtSetMap` altogether but don't override `getCachedPointsTo`. In that case `getCachedPointsTo` will return empty sets.

yuleisui:
I think it is ok for now. Can we remove `getCachedTLPointsTo` and `getCachedADPointsTo` and make a comment suggesting that `getCachedPointsTo` should not be overriding? Any better idea?

mbarbar:
I think it is the other way around: `getCachedPointsTo` (if we accessed the maps directly) would need to be updated if the underlying maps used are changed, otherwise it's all good to keep.

I believe the clean long term solution is to abstract away points-to set manipulation (like the union, clear, etc. methods in the `PTData` family).

For now, I decided to make things `const` so it's clear that these methods are used for reading and not modifying points-to sets. I've added another `unionDDAPts` method which relies on implementation detail of the class. Now overriding `getCachedTLPointsTo` should update `getCachedPointsTo` and any method which relies on it without a problem (for reading points-to sets), and updating `unionDDAPts(DPIm, ...)` should allow us to correctly write points-to sets.

yuleisui:
`UnionDDAPts` Looks good to me even it is a work around. Let me know if you are done, I will merge it.

mbarbar:
That should be it, ready to merge, perhaps after Travis is done building.

yuleisui:
There are quite a few warnings due to `override` keyword required for virtual functions: https://stackoverflow.com/questions/29145476/requiring-virtual-function-overrides-to-use-override-keyword

https://travis-ci.org/github/SVF-tools/SVF/jobs/732520117

mbarbar:
Good catch, I should have looked at the macOS build too. Made a PR now.

PR #333, opened 2020-10-05
Use DenseMap for PtsMap.

mbarbar:
As title. Required specifying CondVar for DenseMapInfo. The hash could be improved, but it shouldn't be too bad.

Some Andersen's stats, in seconds:
mutt  39.4 -> 32.3
mruby 9.77 -> 6.194

Edit: Also, I decided to implement the empty and tombstone keys like LLVM does, with ~0 and ~1. It think with such a NodeID, no analysis can scale...
https://llvm.org/doxygen/DenseMapInfo_8h_source.html#118

### DIFF
--- a/include/DDA/DDAVFSolver.h
+++ b/include/DDA/DDAVFSolver.h
@@ -33,10 +33,10 @@ public:
     typedef OrderedMap<DPIm, CPtSet> DPImToCPtSetMap;
     typedef OrderedMap<DPIm,CVar> DPMToCVarMap;
     typedef OrderedMap<DPIm,DPIm> DPMToDPMMap;
-    typedef Map<NodeID, DPTItemSet> LocToDPMVecMap;
-    typedef Set<const SVFGEdge* > ConstSVFGEdgeSet;
+    typedef OrderedMap<NodeID, DPTItemSet> LocToDPMVecMap;
+    typedef OrderedSet<const SVFGEdge* > ConstSVFGEdgeSet;
     typedef SVFGEdge::SVFGEdgeSetTy SVFGEdgeSet;
-    typedef Map<const SVFGNode*, DPTItemSet> StoreToPMSetMap;
+    typedef OrderedMap<const SVFGNode*, DPTItemSet> StoreToPMSetMap;
 
     ///Constructor
     DDAVFSolver(): outOfBudgetQuery(false),_pag(NULL),_svfg(NULL),_ander(NULL),_callGraph(NULL), _callGraphSCC(NULL), _svfgSCC(NULL), ddaStat(NULL)
--- a/include/DDA/DDAVFSolver.h
+++ b/include/DDA/DDAVFSolver.h
@@ -33,8 +33,8 @@ public:
     typedef std::map<DPIm, CPtSet> DPImToCPtSetMap;
     typedef std::map<DPIm,CVar> DPMToCVarMap;
     typedef std::map<DPIm,DPIm> DPMToDPMMap;
-    typedef DenseMap<NodeID, DPTItemSet> LocToDPMVecMap;
-    typedef DenseSet<const SVFGEdge* > ConstSVFGEdgeSet;
+    typedef SVFMap<NodeID, DPTItemSet> LocToDPMVecMap;
+    typedef SVFSet<const SVFGEdge* > ConstSVFGEdgeSet;
     typedef SVFGEdge::SVFGEdgeSetTy SVFGEdgeSet;
     typedef std::map<const SVFGNode*, DPTItemSet> StoreToPMSetMap;
 
--- a/include/DDA/DDAVFSolver.h
+++ b/include/DDA/DDAVFSolver.h
@@ -29,14 +29,14 @@ public:
     typedef SCCDetection<PTACallGraph*> CallGraphSCC;
     typedef PTACallGraphEdge::CallInstSet CallInstSet;
     typedef PAG::CallSiteSet CallSiteSet;
-    typedef std::set<DPIm> DPTItemSet;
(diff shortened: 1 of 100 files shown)


yuleisui:
One flow-sensitive test case failed.

https://travis-ci.org/github/SVF-tools/SVF/jobs/732846128#L1496

mbarbar:
`-ander` results with...
`std::map`:
mruby: 19s
mutt: 49.8s

`std::unordered_map`:
mruby: 16.8s
mutt: 41.67s

Like we discussed, the STL maps are simpler and `DenseMap` is faster, thoughn that would need more tests to determine. Regardless, if we're moving to STL anyway, it doesn't matter. And like we discussed, perhaps someday a map tailored to SVF could be written.

Though I'm not 100% sure about the new `operator==` for `LocationSet` (required by `unordered_map`), perhaps @kisslune can take a look.

include/DDA/DDAVFSolver.h
```diff
@@ -33,8 +33,8 @@ class DDAVFSolver
     typedef std::map<DPIm, CPtSet> DPImToCPtSetMap;
```
yuleisui, line 33: can we also replace all the std::map with SVFMap?

yuleisui (commented):
I can see there are still quite a few places using std::map. Shall we change all of them?

mbarbar:
Good catch. I had changed `Dense*` but not `std::*`.

Currently `SVFSet` and `Map` are set to `std::map` and `std::set`. Their unordered counterparts are faster. Should I switch to those?

Another issue... what if someone needs an ordered set? For example, in `BVDataPTAImpl::dumpAllPts`, the `NodeSet` is used to order the PAG nodes before dumping (the PAG iterator was on a `DenseMap` before, if my memory serves me well). Should we have...

`SVFSet = std::set or std::unordered_set`
`SVFMap = std::map or std::unordered_map`
`SVFOrderedMap = std::map`
`SVFOrderedSet = std::set`

So then, `SVFSet` and `SVFMap` can be set to anything (ordered or unordered: the user can't rely on that knowledge except that they are maps or sets), but `SVFOrderedSet` and `SVFOrderedMap` will *always* be ordered, no matter what implementation they use (LLVM, Boost, STL, ...).

Finally, should we drop the `SVF` prefix and just call them `Map` and `Set`? May be simpler. They are under the `SVF` namespace afterall... `SVF::SVFMap` vs `SVF::Map`.

yuleisui:
> Good catch. I had changed `Dense*` but not `std::*`.
> 
> Currently `SVFSet` and `Map` are set to `std::map` and `std::set`. Their unordered counterparts are faster. Should I switch to those?
> 
> Another issue... what if someone needs an ordered set? For example, in `BVDataPTAImpl::dumpAllPts`, the `NodeSet` is used to order the PAG nodes before dumping (the PAG iterator was on a `DenseMap` before, if my memory serves me well). Should we have...
> 
> `SVFSet = std::set or std::unordered_set`
> `SVFMap = std::map or std::unordered_map`
> `SVFOrderedMap = std::map`
> `SVFOrderedSet = std::set`
> 
Yes, this is a good idea!

> So then, `SVFSet` and `SVFMap` can be set to anything (ordered or unordered: the user can't rely on that knowledge except that they are maps or sets), but `SVFOrderedSet` and `SVFOrderedMap` will _always_ be ordered, no matter what implementation they use (LLVM, Boost, STL, ...).
> 
> Finally, should we drop the `SVF` prefix and just call them `Map` and `Set`? May be simpler. They are under the `SVF` namespace afterall... `SVF::SVFMap` vs `SVF::Map`.
> 
Sounds good to me.
> Edit: let me fix the CI failure.

mbarbar:
Fixed and added, the commit messages should explain rationale.

Still, I am unsure of the `operator==` I implemented for `LocationSet`; there is some commented code in `operator<` and I do not know why it is commented.

yuleisui:
This change is significant. Just in case, can you test a few bcs to see whether this pull request will produce the same points-to results as the old one?

mbarbar:
For `-fspta` and `-ander`, the following produce the same average top level points-to set sizes as the old implementation (tested against a commit just before the `PTData` refactor):

du.bc
i3.bc
mruby.bcc
dpkg.bcc
ninja.bcc
janet.bc
tmux.bc
nvim.bc
bake.bc

yuleisui:
Any benchmarks/programs produce different `average top-level points-to set sizes`, if not I will merge it now.

mbarbar:
I ran tests on those only (nvim and bake were the problematic ones from the renaming perspective we discussed) and all them were successful. If you'd like me to run on some larger programs, I can but it will take a while to complete.

yuleisui:
Yes, please run some larger ones.

mbarbar:
Sure. I'm running on bash, mutt, and hyriseConsole now.

mbarbar:
Tested against bash, mutt, and hyriseConsole and all is good. I also changed some unordered maps to ordered maps (which object creation may rely on the iteration of) and the issues with nvim and bake appear to be fixed.

yuleisui:
Very Good! 
For DDA, can we change in order map or set to ordered ones because dynamic recursive  traversal in DDA analyses may require deterministic order. I think you can still keep the hash functions you have written, just in case if we are interested in using unordered set later.

mbarbar:
Done. Went through all instances of `Set` and `Map` in `include/DDA` and `lib/DDA` (except I barely touched Stat) and changed them to ordered. There are some instances of types defined outside DDA which I didn't change, like `typedef PAG::CallSiteSet CallSiteSet;`. I also didn't change two variables, `ander_vfns` and `dda_vfns`.

Not sure about testing. Maybe in the future DDA tests in Test-Suite could make the process quicker.

PR #335, opened 2020-10-09
Addendum to PTData refactor

mbarbar:
Reverse points-to sets need to account for the fact that we separated key and datum. A reverse set is saying... given this **Datum**, these are **Keys** (i.e. a set of Keys, not a Data), that contain that datum in their associated **Data**. Thus it is `Map<Datum, KeySet>`. I used `Set` for the `KeySet` and suprisingly for large Andersen runs... it is faster.

For example, on my machine, (avg. TL PTS size in parentheses),
```
lynx: 74s to 57s (150.492)
bash: 29s to 20s (239.593)
astyle: 24.7s to 19.4s (211.37)
mruby: 6.466s to 6.614s (20.2511)
```

Overall, perhaps `SparseBitVector` can be reconsidered to `unordered_set` or even a plain `BitVector`*. Memory usage is obviously also an important factor to consider, though, and I didn't test that.

\* The BCs I've worked on have always had less than 10000 objects giving a maximum BC memory usage of less then 1.5 kilobytes (discounting BitVector's internal maintenance). With a dynamically sized bit vector, this is obviously much smaller. Also with a dynamically sized bit vector, a cheap pre-analysis can predict which objects are likely to be stored together and hence give them similar values so we don't have bit vectors with huge gaps (e.g. instead of a bit vector storing node IDs 5 and 2560 and 7000, we can rename them to, 8, 9, 10, for example). The difficulty is considering all relations when renaming but I can't help but think something analogous can be found in another field. Whatever... maybe that's another paper. :)

### DIFF
--- a/include/MemoryModel/AbstractPointsToDS.h
+++ b/include/MemoryModel/AbstractPointsToDS.h
@@ -37,6 +37,8 @@ template <typename Key, typename Datum, typename Data>
 class PTData
 {
 public:
+    typedef Set<Key> KeySet;
+
     /// Types of a points-to data structures.
     enum PTDataTy
     {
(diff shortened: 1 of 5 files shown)


yuleisui:
Good! Let me merge it now.

Issue #328, opened 2020-09-28
WPA tool crashes with -locMM switch on string handling


pietroborrello:
I'm trying to build an analysis where pointers stored in an array do not share the same points-to set, to be able to distinguish the objects that may be pointed by each of the pointers in the array. However, enabling the `-locMM` flag (for which I'm still not sure it would solve my problem but saw #11 ), segfaults the `wpa` tool during PointerAnalysis initialization, while parsing string accesses.

To reproduce (a minimal poc):
`main.c`
```
#include <stdio.h>
int main(int argc, char **argv) {
	printf("hello\n");
	return 0;
}
``` 
then execute:
```
$ clang -c -fno-discard-value-names -emit-llvm main.c -o main.bc
$ opt -mem2reg main.bc -o main.opt
$ wpa -ander -debug --dwarn=0 --print-all-pts -locMM -stats=0 main.opt
```

The tool segfaults with the following output:
```
> $ wpa -ander -debug --dwarn=0 --print-all-pts -locMM -stats=0 main.opt                                                            
Args: wpa -ander -debug --dwarn=0 --print-all-pts -locMM -stats=0 main.opt 
Building PAG ...
Building Symbol table ...
collect sym from ##@.str = private unnamed_addr constant [7 x i8] c"hello\0A\00", align 1 
create a new value sym 4
handle constant expression i8* getelementptr inbounds ([7 x i8], [7 x i8]* @.str, i64 0, i64 0)
create a new value sym 5
[1]    24137 segmentation fault (core dumped)  wpa -ander -debug --dwarn=0 --print-all-pts -locMM -stats=0 main.opt
```

At a quick analysis in a debugger the following stack trace is generated:
```
[#0] 0x555555636122 → SymbolTableInfo::getStructInfo(this=0x5555566cfbf0, T=0x5555566d1150)
[#1] 0x555555632be1 → SymbolTableInfo::isConstantObjSym(this=0x5555566cfbf0, val=0x5555566d2858)
[#2] 0x5555556323e8 → SymbolTableInfo::collectVal(this=0x5555566cfbf0, val=0x5555566d2858)
[#3] 0x555555632db7 → SymbolTableInfo::handleCE(this=0x5555566cfbf0, val=0x5555566d90c8)
[#4] 0x55555563319b → SymbolTableInfo::handleGlobalCE(this=0x5555566cfbf0, G=0x5555566d2858)
[...]

PR #336, opened 2020-10-12
Add back LocMemModel support

pietroborrello:
Pull request for issue #328 
Basically add a virtual method wrapper in `SymbolTableInfo`, i.e. `SymbolTableInfo::getLocationSetOffset`, which is overriden by `LocSymTableInfo` to select either `ls.getOffset()` or `ls.getByteOffset()`. 
I have added calls to such method in `PAG::getGepObjNode` and `PAG::addGepObjNode`: this should not break the compatibility between `locMM` and non `locMM` memory models.
Btw I'm not sure if I have to use `SymbolTableInfo::Symbolnfo()->getLocationSetOffset(ls)` in place of every invocation of `ls.getOffset()` in the entire code base, or just in the PAG methods.

### DIFF
--- a/lib/Graphs/PAG.cpp
+++ b/lib/Graphs/PAG.cpp
@@ -608,7 +608,7 @@ NodeID PAG::getGepObjNode(const MemObj* obj, const LocationSet& ls)
     NodeID base = getObjectNode(obj);
 
     // Base and first field are the same memory location.
-    if (FirstFieldEqBase && ls.getOffset() == 0) return base;
+    if (FirstFieldEqBase && SymbolTableInfo::Symbolnfo()->getLocationSetOffset(ls) == 0) return base;
 
     /// if this obj is field-insensitive, just return the field-insensitive node.
     if (obj->isFieldInsensitive())
@@ -643,7 +643,7 @@ NodeID PAG::addGepObjNode(const MemObj* obj, const LocationSet& ls)
                                             getNodeNumAfterPAGBuild() > StInfo::getMaxFieldLimit() ?
                                             getNodeNumAfterPAGBuild() : StInfo::getMaxFieldLimit()
                                         )));
-    NodeID gepId = (ls.getOffset() + 1) * gepMultiplier + base;
+    NodeID gepId = (SymbolTableInfo::Symbolnfo()->getLocationSetOffset(ls) + 1) * gepMultiplier + base;
     GepObjNodeMap[std::make_pair(base, ls)] = gepId;
     GepObjPN *node = new GepObjPN(obj, gepId, ls);
     memToFieldsMap[base].set(gepId);
--- a/lib/MemoryModel/MemModel.cpp
+++ b/lib/MemoryModel/MemModel.cpp
@@ -306,21 +306,13 @@ bool LocSymTableInfo::computeGepOffset(const User *V, LocationSet& ls)
 {
 
     assert(V);
-    int baseIndex = -1;
     int index = 0;
+
     for (bridge_gep_iterator gi = bridge_gep_begin(*V), ge = bridge_gep_end(*V);
             gi != ge; ++gi, ++index)
     {
+
         if(SVFUtil::isa<ConstantInt>(gi.getOperand()) == false)
-            baseIndex = index;
-    }
-
-    index = 0;
-    for (bridge_gep_iterator gi = bridge_gep_begin(*V), ge = bridge_gep_end(*V);
-            gi != ge; ++gi, ++index)
-    {
-
-        if (index <= baseIndex)
         {
             /// variant offset
             // Handling pointer types
--- a/lib/MemoryModel/MemModel.cpp
+++ b/lib/MemoryModel/MemModel.cpp
@@ -345,6 +345,12 @@ bool LocSymTableInfo::computeGepOffset(const User *V, LocationSet& ls)
                 Size_t num = at->getNumElements();
                 ls.addElemNumStridePair(std::make_pair(num, sz));
             }
+            else if (const StructType *ST = SVFUtil::dyn_cast<StructType>(*gi) )
+            {
+                // Handling struct here
+                Size_t sz = getTypeSizeInBytes(ST);
+                ls.addElemNumStridePair(std::make_pair(1, sz));
+            }
             else
                 assert(false && "what other types?");
         }
--- a/lib/MemoryModel/MemModel.cpp
+++ b/lib/MemoryModel/MemModel.cpp
@@ -395,98 +395,6 @@ bool LocSymTableInfo::computeGepOffset(const User *V, LocationSet& ls)
 }
 
 /*!
- * Collect array information
- */
-void LocSymTableInfo::collectArrayInfo(const llvm::ArrayType*)
-{
-    /*
-    StInfo *stinfo = new StInfo();
-    typeToFieldInfo[ty] = stinfo;
-
-    /// If this is an array type, calculate the outmost array
-    /// information and append them to the inner elements' type
-    /// information later.
-    u64_t out_num = ty->getNumElements();
-    const Type* elemTy = ty->getElementType();
-    u32_t out_stride = getTypeSizeInBytes(elemTy);
-
-    /// Array itself only has one field which is the inner most element
-    stinfo->addOffsetWithType(0, elemTy);
-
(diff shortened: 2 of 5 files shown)


pietroborrello:
Ah `LocSymTableInfo::computeGepOffse` is missing the check for the `StructType` as has been added to `SymbolTableInfo::computeGepOffse`

pietroborrello:
I added the case for the struct type in `LocSymTableInfo::computeGepOffset` for non constant offsets. Btw it is not completely clear to me the logic here. Why should we `addElemNumStridePair` for each `index <= baseIndex` and not only for `index == baseIndex`, just incrementing the byte offset in the case the initial args of the GEP are constants?

lib/MemoryModel/MemModel.cpp
```diff
@@ -345,6 +345,12 @@ bool LocSymTableInfo::computeGepOffset(const User *V, LocationSet& ls)
                 Size_t num = at->getNumElements();
                 ls.addElemNumStridePair(std::make_pair(num, sz));
             }
+            else if (const StructType *ST = SVFUtil::dyn_cast<StructType>(*gi) )
```
yuleisui, line 348: Will this ever be executed if this is a non-constant offset? Not quite sure, since it is bit a long time ago writing this.

yuleisui (commented):
Your fix looks good and clean to me. Would you mind trying to test the original non-LocMM implementation using some large size bcs to make sure their points-to results are identical before and after your changes?

For LocMM setting, you can always tweak the code in `LocSymTableInfo` to improve soundness and correctness...

lib/MemoryModel/MemModel.cpp
pietroborrello: Yes in the case a gep indexes a structure that contains an array with a variable index like
```c
typedef struct A_s {
	int a;
	int arr[128];
} A_t;

printf("%d\n", s->arr[i]);
```
which translates to:
```c
%struct.A_s = type { i32, [128 x i32] }

%arrayidx = getelementptr inbounds %struct.A_s, %struct.A_s* %s, i64 0, i32 1, i64 %i
```

pietroborrello: The problem is that the branch is executed for every `index <= baseIndex`, is it necessary?

pietroborrello:
The original non-locMM implementation should not have changed, and I'm testing it. My main concern is if there is any other point in the code where I should use `SymbolTableInfo::getLocationSetOffset` instead of `LocationSet::getOffset`.
Can I substitute all the occurrences, so that `LocationSet::getByteOffset` would be used everywhere when locMM is enabled? 
Or is LocationSet used as a structure also independently from SymbolTableInfo, so we should keep `LocationSet::getOffset`?

yuleisui:
`LocationSet` is used as a structure also independently from SymbolTableInfo. However, you plan to use `SymbolTableInfo::Symbolnfo()->getLocationSetOffset(ls)` to implement polymorphism.

It might be messy to replace them all since there are quite a few places using `getOffset()`: https://github.com/SVF-tools/SVF/search?q=getOffset%28%29

Another way is to just return the value of `getByteOffset()` when invoking `getOffset()` if you flag is turned on?

pietroborrello:
Yes, checking the `SymbolTableInfo::Symbolnfo()` state inside of `getOffset` to return the value of `getByteOffset` might be an option too, but this would effectively replace all the occurrences with my method, this is why I was asking. Would it be correct?

yuleisui:
At least the original option (field-based offset) is correct. You can try to see all the places (around 5-6 places except printing) using `getOffset()` to see whether it is correct when your flag is turned on.

pietroborrello:
For example `LocationSet::computeAllLocations` which is called in the `LocationSet::checkRelation` and `LocationSet::intersect` method use `getOffset`, and these should be changed to make locMM work properly I think. These methods are not used outside of the MemModel classes, but I don't know if in future will be the case, and using the byte offset would cause bugs. Can I change them?

yuleisui:
Yes, you can change any code as long as the current field-based `getOffset` is correct. Please keep the patch to be minimum too.

pietroborrello:
Ok I changed the minimum to make locMM work, and should improved precision and scalability managing variant offsets only when needed. Moreover I confirm the non locMM mode works the same in a 50Mb bc file I'm using in my project, as the behaviour of the wrapper should be transparent

lib/Graphs/PAG.cpp
```diff
@@ -608,7 +608,7 @@ NodeID PAG::getGepObjNode(const MemObj* obj, const LocationSet& ls)
     NodeID base = getObjectNode(obj);
 
     // Base and first field are the same memory location.
-    if (FirstFieldEqBase && ls.getOffset() == 0) return base;
```
yuleisui, line 611: To be safe, I would suggest making `getOffset()` as a protected method in `SymbolTableInfo` to prevent from future usage?

Or maybe you can just use your flag inside `getOffset()`.

yuleisui (commented):
It would be good to handle `getOffset` consistently across the repo.

pietroborrello:
I'm not sure I'm understanding, `getOffset()` is a method in `LocationSet`. Adding the check on `SymbolTableInfo::Symbolnfo()` inside that method would mean to tie `LocationSet` to `SymbolTableInfo`, and if I understood correctly `LocationSet` may be used outside the `SymbolTableInfo` use case.
I'm realizing that the previous model where LocationSet had only a single field which was filled with either byte offset or field offset was easier to deal with without breaking the consistency of the code. 
Moreover, I just noticed that `SymbolTableInfo::Symbolnfo()->getFlattenFieldInfoVec` is also used in different methods (as in [`ObjTypeInfo::isNonPtrFieldObj`](https://github.com/SVF-tools/SVF/blob/f64b9fae14ed66b2f25bde3c4d37aad6ee5037ca/lib/MemoryModel/MemModel.cpp#L193)) and may be broken with `locMM` enabled, so that (and similar methods) should be substituted as well.

Given the overall code is in a shape which favours non `locMM` usage, I'm starting to consider closing this pull request and adding instead a flag to enable array idx sensitivity to the field-based analysis. Do you think it would be better?
I think `locMM` would be primarily used to obtain idx sensitivity instead of really needing byte granularity, and that would match my use case.
Is there any reason array idx sensitivity (at least constant ones) was never added to the field sensitive analysis?

yuleisui:
> I'm not sure I'm understanding, `getOffset()` is a method in `LocationSet`. Adding the check on `SymbolTableInfo::Symbolnfo()` inside that method would mean to tie `LocationSet` to `SymbolTableInfo`, and if I understood correctly `LocationSet` may be used outside the `SymbolTableInfo` use case.
> I'm realizing that the previous model where LocationSet had only a single field which was filled with either byte offset or field offset was easier to deal with without breaking the consistency of the code.
> Moreover, I just noticed that `SymbolTableInfo::Symbolnfo()->getFlattenFieldInfoVec` is also used in different methods (as in [`ObjTypeInfo::isNonPtrFieldObj`](https://github.com/SVF-tools/SVF/blob/f64b9fae14ed66b2f25bde3c4d37aad6ee5037ca/lib/MemoryModel/MemModel.cpp#L193)) and may be broken with `locMM` enabled, so that (and similar methods) should be substituted as well.
> 
> Given the overall code is in a shape which favours non `locMM` usage, I'm starting to consider closing this pull request and adding instead a flag to enable array idx sensitivity to the field-based analysis. Do you think it would be better?
Yes, I think so.
> I think `locMM` would be primarily used to obtain idx sensitivity instead of really needing byte granularity, and that would match my use case.
> Is there any reason array idx sensitivity (at least constant ones) was never added to the field sensitive analysis?
SVF's pointer analysis starts with standard Andersen (array-insensitive). The array-sensitive one was an experimental version. With the continuous evolvement, that one was not maintained and tested. I believe we can bring it back with more robust implementation and testing.

pietroborrello:
> SVF's pointer analysis starts with standard Andersen (array-insensitive). The array-sensitive one was an experimental version. With the continuous evolvement, that one was not maintained and tested. I believe we can bring it back with more robust implementation and testing.

Great, do you have any pointer to the array-sensitive implementation code?

yuleisui:
I think you can find the old implementation of locMM in previous svf releases like SVF-1.6 or 1.7 https://github.com/SVF-tools/SVF/releases/tag/SVF-1.7

PR #356, opened 2021-01-04
Add versioned FS points-to analysis

mbarbar:
As title. Largely self-contained. The most major non-self-contained change is adding an option to add reverse PTS in the data structure.


lib/Util/SVFUtil.cpp
```diff
@@ -135,6 +135,24 @@ void SVFUtil::dumpSet(NodeBS bs, raw_ostream & O)
     }
 }
 
+void SVFUtil::dumpMeldVersion(MeldVersion &v)
```
yuleisui, line 138: why this method here, shall we put it into your versioned pts?

include/Util/SVFBasicTypes.h
```diff
@@ -58,8 +58,10 @@ typedef u32_t EdgeID;
 typedef unsigned SymID;
 typedef unsigned CallSiteID;
 typedef unsigned ThreadID;
+typedef unsigned Version;
 
 typedef llvm::SparseBitVector<> NodeBS;
+typedef llvm::SparseBitVector<> MeldVersion;
```
yuleisui, line 64: Anywhere else Meldversion, if not we can put it under VFS?

mbarbar:
Done. Moved to VFS and made private.

PR #360, opened 2021-01-24
Introduce dense node ID allocation

mbarbar:
This PR introduces dense node ID allocation, keeping the old allocation as an option.

Dense allocation:
Allocates all objects in a cluster, and all values in a cluster. This makes points-to sets more compact. This is the extent of the heuristic, however you should see good memory and time improvements, even in Andersen's.

Debug allocation:
The old allocation: objects and values are interleaved and GEPs use the multiplier.

New option is `-node-alloc-strat` and takes either `debug` or `dense` and defaults to `dense`. Node ID allocation is separated from the symbol table and used in multiple places.

### DIFF
--- a/lib/SVF-FE/SymbolTableInfo.cpp
+++ b/lib/SVF-FE/SymbolTableInfo.cpp
@@ -50,6 +50,8 @@ SymID SymbolTableInfo::totalValSymNum = 0;
 
 const std::string SymbolTableInfo::userNodeAllocationStrategyDense = "dense";
 const std::string SymbolTableInfo::userNodeAllocationStrategyDebug = "debug";
+enum SymbolTableInfo::NodeAllocationStrategy SymbolTableInfo::allocStrat =
+    SymbolTableInfo::NodeAllocationStrategy::DENSE;
 
 static llvm::cl::opt<unsigned> maxFieldNumLimit("fieldlimit",  llvm::cl::init(512),
         llvm::cl::desc("Maximum field number for field sensitive analysis"));
--- a/lib/SVF-FE/SymbolTableInfo.cpp
+++ b/lib/SVF-FE/SymbolTableInfo.cpp
@@ -90,10 +90,6 @@ void MemObj::init(const Value *val)
     }
 }
 
-SymbolTableInfo::SymbolTableInfo(void)
-    : modelConstants(false), maxStruct(NULL), maxStSize(0)
-{ }
-
 /*!
  * Get the symbol table instance
  */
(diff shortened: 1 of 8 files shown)


mbarbar:
I should point out. I tested with a bunch of bitcode, `-ander` and `-fspta`, and average top level points-to set sizes are equivalent.

Average points-to set sizes are not a good measure because they are currently non-deterministic, as I found out a few weeks ago. The number of objects is also not a good measure because it is non-deterministic (with and without this PR), and my suspicion is that according to some node allocations + the non-determinism of unordered maps means some PWCs get collapsed faster/slower preventing/allowing creation of some fields, but I haven't tested that hypothesis.

yuleisui:
Apart from vsfs, all other pointer analyses seems crashed when doing testing. https://travis-ci.org/github/SVF-tools/SVF/jobs/755922657#L1380-L1540

mbarbar:
I had initialised some constants in the wrong order, and didn't test the default argument for the new flag... fixed now.

VSFS should have failed too, so those tests must not be running properly. I'll investigate when I have time, maybe later next week.

lib/SVF-FE/SymbolTableInfo.cpp
```diff
@@ -92,6 +90,10 @@ void MemObj::init(const Value *val)
     }
 }
 
+SymbolTableInfo::SymbolTableInfo(void)
```
yuleisui, line 93: Fix this based on our discussion

lib/Util/NodeIDAllocator.cpp
```diff
@@ -0,0 +1,119 @@
...
+    const NodeID NodeIDAllocator::blackHolePointerId = 2;
+    const NodeID NodeIDAllocator::nullPointerId = 3;
+
+    static llvm::cl::opt<std::string> nodeAllocStrat(
+        "node-alloc-strat", llvm::cl::init(SVF::NodeIDAllocator::userStrategyDense),
+        llvm::cl::desc("Method of allocating (LLVM) values to node IDs [dense, debug]"));
+
+    NodeID NodeIDAllocator::allocateObjectId(void)
+    {
+        if (strategy == Strategy::NONE) setStrategy(nodeAllocStrat);
```
yuleisui, line 28: make it singleton

lib/Util/NodeIDAllocator.cpp
```diff
@@ -0,0 +1,119 @@
+//===- NodeIDAllocator.cpp -- Allocates node IDs on request ------------------------//
+
+#include "Util/NodeIDAllocator.h"
+
+namespace SVF
+{
+    NodeID NodeIDAllocator::numObjects = 3;
+    NodeID NodeIDAllocator::numValues = 3;
+    NodeID NodeIDAllocator::numSymbols = 3;
+    NodeID NodeIDAllocator::numNodes = 4;
```
yuleisui, line 10: make sure two strategies both start from ID 4 for val and obj nodes.

mbarbar:
Implemented and tested on my end. If you're happy with the changes and CI doesn't complain, could probably merge.

PR #365, opened 2021-02-11
Add singleton VFSPTA

mbarbar:
Copied from `FlowSensitive`. I think this should solve what you're after.

### DIFF
--- a/include/WPA/VersionedFlowSensitive.h
+++ b/include/WPA/VersionedFlowSensitive.h
@@ -79,6 +79,25 @@ public:
     }
     //@}
 
+    /// Create single instance of versioned flow-sensitive points-to analysis.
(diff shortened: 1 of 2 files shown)


include/WPA/VersionedFlowSensitive.h
```diff
@@ -167,6 +186,8 @@ class VersionedFlowSensitive : public FlowSensitive
     double meldMappingTime;  ///< Time to map MeldVersions to Versions.
     double versionPropTime;  ///< Time to propagate versions to versions which rely on them.
     //@}
+
+    static VersionedFlowSensitive *vfspta;
```
yuleisui, line 190: Do we need a null initialization in VersionedFlowSensitive.cpp?

mbarbar: Yes, my apologies.

PR #367, opened 2021-02-14
Removed redundant destroy()

charlesxsh:
In ICFG's parent class GenericGraph, destroy() has been called once. In child's virtual destructor, we don't need to call it again since parent destructor will do that.


charlesxsh:
#368

PR #373, opened 2021-02-20
Make options global

mbarbar:
The Options class gathers all command line options (or at least, aims to). For example, option `X` would be accessed as `Options::X` after including `Util/Options.h`.

Why not singleton? Command line options are set once for the entire execution, it makes no sense to change these (unless we might want to internally mess with an option depending on other options, perhaps?).

**Rationale**
I wanted to add an option used in many files. It is possible to declare an `llvm::cl::opt` as `extern` in some header, then initialise it in some `.cpp`, but I thought this would be cleaner as it puts all options in one place rather than interspersed across files and allows everyone to view *any* options from one place.

**Status**
I haven't moved every option to `Options` as the testing effort would be too large. I've moved `marked-clocks-only` and `node-alloc-strat` as examples. I think `Options.h`/`.cpp` are very readable and it clears some clutter from other files. My suggestion is we add new options here, and move existing options slowly over time.

### DIFF
--- a/include/MemoryModel/PTAStat.h
+++ b/include/MemoryModel/PTAStat.h
@@ -31,6 +31,7 @@
 #define ANDERSENSTAT_H_
 
 #include "Util/BasicTypes.h"
+#include "Util/Options.h"
 #include <iostream>
 #include <map>
 #include <string>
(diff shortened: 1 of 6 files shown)


yuleisui:
The patch looks good to me. Would it be possible to simply move all the options to Options.h/cpp? I think you could do a search for `static llvm::cl::opt` to find them all.

mbarbar:
I'd like to... it's just with deadlines it might eat too much time due to the testing involved to make sure nothing breaks during the move. There are about 100 `cl::opt`s.

yuleisui:
Are you putting more code for global options? Can I merge it now?

mbarbar:
If you're okay with it being a gradual thing, it is ready to merge. I just added the field limit because I wanted to use that option somewhere.

PR #376, opened 2021-02-25
simplify solve() in Andersen and FlowSensitive analysis

yuleisui:
Move the body of `solve()` in WPASolver.h to `Andersen.cpp` and `FlowSensitive.cpp` to simplify the code.

### DIFF
--- a/lib/WPA/FlowSensitive.cpp
+++ b/lib/WPA/FlowSensitive.cpp
@@ -78,8 +78,8 @@ void FlowSensitive::analyze()
 
         callGraphSCC->find();
 
-        solve();
-
+        initWorklist();
+        solveWorklist();
     }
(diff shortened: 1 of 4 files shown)


lib/WPA/FlowSensitive.cpp
```diff
@@ -78,8 +78,8 @@ void FlowSensitive::analyze()
 
         callGraphSCC->find();
 
-        solve();
-
+        initWorklist();
+        solveWorklist();
     }
```
mbarbar, line 83: Which `init`/`solveWorklist` functions are called here, being that `FlowSensitive` inherits from `WPAFSSolver`, instead of `WPASolver`.

mbarbar:
Looks good to me.

PR #379, opened 2021-02-28
Code improvement

benquike:
1. Remove unneeded Makefiles
2. Factor out llvm_libs setup
3. cleanup CXX_FLAGS setup


lib/CUDD/CMakeLists.txt
```diff
@@ -39,7 +39,7 @@ set(SOURCES
 
 add_llvm_library(Cudd ${SOURCES} LINK_LIBS m)
 
-set_target_properties(Cudd PROPERTIES COMPILE_FLAGS "-Wno-format -Wno-int-to-pointer-cast -Wno-pointer-to-int-cast -O3 -w -DHAVE_IEEE_754 -DSIZEOF_VOID_P=8 -DSIZEOF_LONG=8")
```
yuleisui, line 42: We will need to keep these options to properly use cudd on a 64 bit machine. The default macros are used for 32-bit ones.

CMakeLists.txt
```diff
@@ -13,21 +13,40 @@ if(NOT COMMAND add_llvm_library)
...
-		set(CMAKE_CXX_FLAGS "-fPIC -std=gnu++14 -O3 -fno-rtti -Wno-deprecated -DHAVE_IEEE_754 -DSIZEOF_VOID_P=8 -DSIZEOF_LONG=8")
-	endif()
-	set(CMAKE_C_FLAGS "-fPIC")
-
-	find_package(LLVM REQUIRED CONFIG)
+        set(CMAKE_CXX_STANDARD_REQUIRED ON)
+        # add -std=gnu++14
+        set(CMAKE_CXX_EXTENSIONS ON)
+
+        add_compile_definitions(HAVE_IEEE_754)
```
yuleisui, line 20: Thanks for your contribution:) Could this be moved to the CMakeList under CUDD since these marcos are only used there?

CMakeLists.txt
```diff
@@ -13,21 +13,40 @@ if(NOT COMMAND add_llvm_library)
         # We need to match the build environment for LLVM:
         # In particular, we need C++11 and the -fno-rtti flag
         set(CMAKE_CXX_STANDARD 14)
-	if(CMAKE_BUILD_TYPE MATCHES "Debug")
-		set(CMAKE_CXX_FLAGS "-fPIC -std=gnu++14 -g -O0 -fno-rtti -Wno-deprecated -DHAVE_IEEE_754 -DSIZEOF_VOID_P=8 -DSIZEOF_LONG=8")
-	else()
-		set(CMAKE_CXX_FLAGS "-fPIC -std=gnu++14 -O3 -fno-rtti -Wno-deprecated -DHAVE_IEEE_754 -DSIZEOF_VOID_P=8 -DSIZEOF_LONG=8")
-	endif()
-	set(CMAKE_C_FLAGS "-fPIC")
```
yuleisui, line 21: Can we still keep -fPIC or did I miss anything in your patch?

benquike: PIC is enabled by default on recent Linux distributions, to support ASLR.
So it is duplicated.

CMakeLists.txt
```diff
@@ -13,21 +13,40 @@ if(NOT COMMAND add_llvm_library)
...
+        set(CMAKE_CXX_FLAGS  "${CMAKE_CXX_FLAGS} ${COMMON_FLAGS} -fno-rtti")
+        set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${COMMON_FLAGS}")
+
+        if(CMAKE_BUILD_TYPE MATCHES "Debug")
+		  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -O0")
+          set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -O0")
+	    else()
+		  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -O3")
+          set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -O3")
+	    endif()
```
yuleisui, line 34: would be nice to adjust these indents.

CMakeLists.txt
yuleisui: Just in case, someone is built on an old version of Linux.

CMakeLists.txt
benquike: Updated, please have a look.

CMakeLists.txt
benquike: Yes. Update.
Please have a look.
Also, I removed `-Wno-deprecated`, as the current code can be built now without it.

CMakeLists.txt
yuleisui: It is ready to merge once you fix these indents. It looks not properly aligned. Thanks

benquike: Updated.

CMakeLists.txt
```diff
@@ -11,23 +11,38 @@ if(NOT COMMAND add_llvm_library)
...
-	endif()
-	set(CMAKE_C_FLAGS "-fPIC")
-
-	find_package(LLVM REQUIRED CONFIG)
+        set(CMAKE_CXX_STANDARD_REQUIRED ON)
+        # add -std=gnu++14
+        set(CMAKE_CXX_EXTENSIONS ON)
+
+        set(COMMON_FLAGS "-fPIC -Werror")
+        set(CMAKE_CXX_FLAGS  "${CMAKE_CXX_FLAGS} ${COMMON_FLAGS} -fno-rtti")
```
yuleisui, line 21: Sorry for one more thing here. Lines 20-21 seems redundant if we have Lines 25-29. Could you merge them to make it compact?

benquike: No. They are not redundant.
25-29 are only for setting up `-Ox`, while 21-22 are for setup options for C and CXX sources in common.

Issue #382, opened 2021-03-03
A Question about vtable parsing


benquike:
I am looking into the code for parsing vtable [here](https://github.com/SVF-tools/SVF/blob/master/lib/SVF-FE/CHG.cpp#L456) and have some questions.

Here, it seems that it is handling null pointers in the vtable after the typeinfo field.
First of all, I am not sure whether there are cases null pointers exist in the vtable.
If possible could you please give me a concrete example?

Secondly, later on in the same function [here](https://github.com/SVF-tools/SVF/blob/master/lib/SVF-FE/CHG.cpp#L570),
some other functions will be inserted into the places where null pointers are detected. Can you explain what is the reason behind this implementation?

Thanks!

yuleisui:
This part of the code is not written by me. You feel free to improve it and would be good to give examples for comparison purposes. Thanks,

PR #384, opened 2021-03-04
3 udpates

benquike:
1. Improve the code for detecting thunk functions
2. When a thunk function is detected in vtable, instead of thunk function itself, its target function is  inserted into the parsed vtable
3. When parsing passible targets from the vtable, a argument type based filtering mechanism is added

### DIFF
--- a/include/SVF-FE/CHG.h
+++ b/include/SVF-FE/CHG.h
@@ -168,7 +168,7 @@ private:
      *
      * virtualFunctionVectors = {{Af1, Af2, ...}, {Bg1, Bg2, ...}}
      */
-    std::vector<std::vector<const SVFFunction*>> virtualFunctionVectors;
+    std::vector<FuncVector> virtualFunctionVectors;
 };
 
 /// class hierarchy graph
@@ -284,6 +284,9 @@ public:
     {
         return chg->getKind() == Standard;
     }
+
+protected:
+    void addFuncToFuncVector(CHNode::FuncVector &v, const SVFFunction *f);
 
 private:
     SVFModule* svfMod;
--- a/include/SVF-FE/CPPUtil.h
+++ b/include/SVF-FE/CPPUtil.h
@@ -47,6 +47,7 @@ struct DemangledName
 {
     std::string className;
     std::string funcName;
+    bool isThunkFunc;
 };
 
 struct DemangledName demangle(const std::string &name);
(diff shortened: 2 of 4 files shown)


yuleisui:
Thanks for the contribution. Would you be able to provide some data or statistics for this change regarding your bcs tested, e.g., the number of CHA edges and nodes, etc.?

benquike:
Consider this test:

```
class Top {
    public:
        int x1;
        int x2;
        Top() {
            x1 = 1;
            x2 = 2;
        }
        virtual void foo(int a) {}
        virtual void bar(bool b) {}
};

class Left: virtual Top {
    public:
        int y1;
        int y2;
        Left() {
            y1 = 3;
            y2 = 4;
        }
        void foo(int a) {}
        virtual void left_foo(char x) {}
        virtual void left_bar(short y) {}
};

class Middle: virtual Top {
    public:
        int y1;
        int y2;
        Middle() {
            y1 = 5;
            y2 = 6;
        }

        void foo(int x) {}
        void bar(bool b) {}
        virtual void middle_foo(unsigned char x) {}
        virtual void middle_bar(unsigned short y) {}
};

class Right: virtual Top {
    public:
        int z1;
        int z2;
        Right() {
            z1 = 7;
            z2 = 8;
        }
        void foo(int x) {}
        virtual void right_foo(long x) {}
        virtual void middle_bar(unsigned long y) {}
};
class Botton: public Left, public Right {
    public:
        int w1;
        int w2;
        Botton() {
            w1 = 9;
            w2 = 10;
        }
        void foo(int x) {}
        void left_foo(char x) {}
        void middle_foo(unsigned char x) {}
        void right_foo(long x) {}
        virtual void bar(char *cp) {}
};
void foo_dispatch(Top *t) {
    // CS1
    t->foo(0);
}
void right_foo_dispatch(Right *r) {
     // CS2
     r->foo(0);
}

void left_dispatch(Left *l) {
    // CS3
    l->left_bar(0);
    // CS4
    l->left_foo('a');
}
int main(int argc, char *argv[]) {
    Right r;
    Middle m;
    Left l;

    Top t;
    Botton b;

    return 0;
}
```

yuleisui:
Could you elaborate more about this case? What were the previous result and that after your fixe?

benquike:
Updated. 
BTW, is it possible to add some support for unittest in the project?

benquike:
Also, I have a question about the vtable parsing code and have pasted my question here: https://github.com/SVF-tools/SVF/issues/382

Can you spare some time to look at it?

yuleisui:
> Updated.
> BTW, is it possible to add some support for unittest in the project?

We plan to add some support for CI/CD.

yuleisui:
> Updated.
> BTW, is it possible to add some support for unittest in the project?

Thanks.

PR #385, opened 2021-03-07
A few commits to improve the code

benquike:
* convert the  line ending of several sources from Windows to Unix
* Turn on more warnings as error in CMakefile and remove some duplicated code
* A lot of refactoring of the code base following the warnings of clangd


CMakeLists.txt
```diff
@@ -17,13 +17,23 @@ if(NOT COMMAND add_llvm_library)
...
+        set(COMMON_FLAGS "-fPIC -Werror -Wall -Wextra -Wno-unused-parameter -Wno-unused-variable")
+
+        if (CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
+            # SVF/lib/Util/TypeBasedHeapCloning.cpp:441:14: error: variable 'clone' is used
+            # uninitialized whenever 'if' condition is false
+            set(COMMON_FLAGS "${COMMON_FLAGS} -Wno-sometimes-uninitialized")
+        elseif (CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
+            set(COMMON_FLAGS "${COMMON_FLAGS} -Wno-maybe-uninitialized -Wno-sign-compare -Wno-unused-but-set-variable -Wno-cast-function-type -Wno-implicit-fallthrough")
+        endif()
+
```
yuleisui, line 29: Why should we add these options for compiling the code base. The ideal one should be `-Werror`?

benquike: Actually, there are a lot of issues in the code if we turn on `-Wextra`.

I  think these reports from the compiler help identifying possible issues a lot.
Those `-Wno-xxx` issues are ones we can have a look at and try to solve.

benquike: Especially those related reported by `Werror=maybe-uninitialized`. These are potential sources of bugs very hard to debug.

yuleisui:
Thanks for your contribution, however, this patch is too huge. We can not merge it like this. Please split patch into several sub patches and explain why those are necessary. 

I would suggest:
(1) There is no need to refactoring the code of CUDD, which we will remove from the mainstream in the future.
(2) For `virtual` and `override`, in some code we would like to keep both even in the derived class in order for polymorphism for external APIs. We may need to take a close look using a separate patch.
(3) I am not quite sure the difference between `using` and `default`. Are those changes necessary?
(4) For `auto`, sometimes it is a bit hard to understand the code. Do we need to use `auto` to replace all types?

PR #386, opened 2021-03-07
Add KeySet template parameter

mbarbar:
As discussed, the reverse set data structure is now controllable.

I also disabled the reverse set when the analysis is field-insensitive. Steensgaard is always field-insensitive of course, so it's never enabled.

### DIFF
--- a/include/MemoryModel/AbstractPointsToDS.h
+++ b/include/MemoryModel/AbstractPointsToDS.h
@@ -35,12 +35,10 @@ namespace SVF
 /// Key:   "owning" variable of a points-to set.
 /// Datum: elements in points-to sets.
 /// Data:  the points-to set; a collection of Datums.
-template <typename Key, typename Datum, typename Data>
+template <typename Key, typename KeySet, typename Datum, typename Data>
 class PTData
 {
 public:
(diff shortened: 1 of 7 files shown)


yuleisui:
If we have `Datum` as a parameter of `PTData`, do we still need `Data`. Can we somehow infer `Data` based on `Datum` to save one parameter?

mbarbar:
Not quite. Same issue with `Key`/`KeySet`. If we have `NodeID` as `Datum`... is the `Data` a `Set<NodeID>` or a `std::bitset` or an `llvm::SparseBitVector` or an `OrderedSet<NodeID>`, etc. In my personal branch, I'm even sometimes doing both `Data = SparseBitVector` and `Data = NormalBitVector` in the same execution despite `Datum` being the same.

For future, it gets hairier when we start considering more fancy objects (though I don't think we should since bit-vectors are nice to deal with).

yuleisui:
I see. Can we use consistent names `Data` and `DataSet`?

mbarbar:
Yes sure. I'll make an update to this PR tomorrow so I make sure to rename variables appropriately too.

yuleisui:
Thank you!

PR #389, opened 2021-03-16
Add --print-symbol-table option to print SymbolTableInfo after it is built

richardlford:
The purpose of dumping the SymbolTableInfo is to be able to correlate the SymID numbers with the LLVM IR. The --print-symbol-table option will print the information for each symbol id in order.


richardlford:
I need to clean up an extraneous change. Will close this and open another.

PR #390, opened 2021-03-16
Add --print-symbol-table option

richardlford:
It is helpful to be able to correlate the SymID numbers in the symbol table with the LLVM IR. This pull request adds a 

--print-symbol-table 

option which will print the symbol table after it has been created.


yuleisui:
Thanks for your contribution. The CI failed with some errors: https://travis-ci.org/github/SVF-tools/SVF/jobs/763155150#L1369 Please check.

richardlford:
This was using Value::dump which does not exist in debug builds. A separate pull request has been created that fixes this.

PR #393, opened 2021-03-17
Add --print-symbol-table option

richardlford:
It is helpful to be able to correlate the SymID numbers in the symbol table with the LLVM IR. This pull request adds a

--print-symbol-table

option which will print the symbol table after it has been created.
This version uses outs() instead of dbgs(), and uses Value:print instead
of Value::dump [which only exists in debug builds].

### DIFF
--- a/lib/SVF-FE/SymbolTableInfo.cpp
+++ b/lib/SVF-FE/SymbolTableInfo.cpp
@@ -45,7 +45,6 @@ using namespace std;
 using namespace SVF;
 using namespace SVFUtil;
 
-
 DataLayout* SymbolTableInfo::dl = nullptr;
 SymbolTableInfo* SymbolTableInfo::symInfo = nullptr;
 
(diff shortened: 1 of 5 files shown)


richardlford:
Looks like the checks passed this time.  @yuleisui, would you like to merge this in, or do you have some comments or changes you would like made?

lib/SVF-FE/SymbolTableInfo.cpp
```diff
@@ -45,6 +45,8 @@ using namespace std;
 using namespace SVF;
 using namespace SVFUtil;
 
+static llvm::cl::opt<bool> SymTabPrint("print-symbol-table", llvm::cl::init(false),
```
yuleisui, line 48: Could you move this option to `options.h/cpp`?

richardlford:
@yuleisui, I moved the option to the Options.* files as you suggested. I also decided to use the new value2String function. And I modified .gitignore to ignore *.obj directories like z3.obj.

richardlford:
@yuleisui, I think it is ready again unless you have more comments.

PR #397, opened 2021-03-25

richardlford:
4. It is possible one might want to view the ICFG before the
   Program Assignment Graph (PAG) has been built. However
   the ICFG stores some data related to the PAG and the
   graph writer for the ICFG tries to access the PAG.
   This change modifes PAG::getPAG by adding an additional
   parameter, skipCreate, which defaults to false.
   When the ICFG dot graph is being created, it calls
   PAG::getPAP with skipCreate false. This will return the
   PAG pointer, if it has been created, but return null if not.
   If it is null, then it skips trying to dump PAG-related
   information.

### DIFF
--- a/include/Graphs/PAG.h
+++ b/include/Graphs/PAG.h
@@ -153,6 +153,13 @@ public:
         }
         return pag;
     }
+
+    /// Return pointer to PAG if it has been built, otherwise nullptr.
+    static inline PAG* tryGetPAG()
+    {
+        return pag;
+    }
+
     static void releasePAG()
     {
         if (pag)
--- a/include/Graphs/PAG.h
+++ b/include/Graphs/PAG.h
@@ -144,11 +144,16 @@ public:
     }
 
     /// Singleton design here to make sure we only have one instance during any analysis
+    /// Sometimes we want to check if PAG has been made, but not make it if it has not.
+    /// In that case use skipCreate = true.
     //@{
-    static inline PAG* getPAG(bool buildFromFile = false)
+    static inline PAG* getPAG(bool buildFromFile = false, bool skipCreate = false)
     {
         if (pag == nullptr)
         {
+            if (skipCreate) {
+                return nullptr;
+            }
             pag = new PAG(buildFromFile);
         }
         return pag;
--- a/include/Graphs/PAG.h
+++ b/include/Graphs/PAG.h
@@ -144,16 +144,11 @@ public:
     }
 
     /// Singleton design here to make sure we only have one instance during any analysis
-    /// Sometimes we want to check if PAG has been made, but not make it if it has not.
-    /// In that case use skipCreate = true.
     //@{
-    static inline PAG* getPAG(bool buildFromFile = false, bool skipCreate = false)
+    static inline PAG* getPAG(bool buildFromFile = false)
     {
         if (pag == nullptr)
         {
-            if (skipCreate) {
-                return nullptr;
-            }
             pag = new PAG(buildFromFile);
         }
         return pag;
--- a/lib/Graphs/ICFG.cpp
+++ b/lib/Graphs/ICFG.cpp
@@ -27,6 +27,7 @@
  *      Author: Yulei Sui
  */
 
+#include <Util/Options.h>
 #include "SVF-FE/LLVMUtil.h"
 #include "Util/SVFModule.h"
 #include "Graphs/ICFG.h"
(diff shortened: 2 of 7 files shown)


lib/Graphs/ICFG.cpp
```diff
@@ -460,11 +471,15 @@ struct DOTGraphTraits<ICFG*> : public DOTGraphTraits<PAG*>
...
-                rawstr << edge->toString();
+            rawstr << value2String(bNode->getInst()) << " \t";
+            PAG* pag = PAG::getPAG(false, true /* skipCreate */);
+            if (pag != nullptr) {
+                PAG::PAGEdgeList&  edges = pag->getInstPTAPAGEdgeList(bNode);
+                for (PAG::PAGEdgeList::iterator it = edges.begin(), eit = edges.end(); it != eit; ++it)
+                {
+                    const PAGEdge* edge = *it;
+                    rawstr << edge->toString();
+                }
```
yuleisui, line 482: I would suggest making an option in `Options.cpp` for users to choose whether they would like to dump the PAG information or not.

include/Graphs/PAG.h
```diff
@@ -144,11 +144,16 @@ class PAG : public GenericGraph<PAGNode,PAGEdge>
...
+    /// In that case use skipCreate = true.
     //@{
-    static inline PAG* getPAG(bool buildFromFile = false)
+    static inline PAG* getPAG(bool buildFromFile = false, bool skipCreate = false)
     {
         if (pag == nullptr)
         {
+            if (skipCreate) {
+                return nullptr;
+            }
```
yuleisui, line 156: This is only used when dumping ICFG. I suggest making an option rather than changing this code fragment.

lib/Graphs/ICFG.cpp
richardlford: That seems like a good idea, though I think if the PAG is available one would want that information when dumping the ICFG. But when dumping from within the debugger, you may be at a point where the PAG has not yet been created. In that case, if you had set the option to dump the PAG info, the ICFG dumper would still need to check whether the PAG was present, as you would not want to trigger creation of the PAG from withing the dumper (which is what was happening to me). I suggest not adding an option, but to avoid complicating getPAG, I'll add a different method, say, getPAGIfAvailable, that can be used by the dumper. Would that we acceptable?

include/Graphs/PAG.h
richardlford: As suggested above, can I revert this change, but add a separate method to get the PAG if available, for the sake of the dumper?

lib/Graphs/ICFG.cpp
yuleisui: Hmm, sometimes, when debugging, we would like to correlate an ICFGNode with a PAGEdge. This is particularly useful for connecting control-dependence and data-dependence. Your way can not achieve this. 

Note that ICFG is created inside the PAG's constructor, which means PAG will always available if you try to use ICFG. It looks a bit odd trying to get a null PAG instance just for the purpose of debugging ICFG.

yuleisui: You can make a change to add an option as a default one to dump only ` value2String(bNode->getInst())` but not PAGEdges. It will be a good compromise for someone who would like to see the connections between PAG and ICFG later.

include/Graphs/PAG.h
```diff
@@ -144,15 +144,27 @@ class PAG : public GenericGraph<PAGNode,PAGEdge>
     }
 
     /// Singleton design here to make sure we only have one instance during any analysis
+    /// Sometimes we want to check if PAG has been made, but not make it if it has not.
+    /// In that case use skipCreate = true.
     //@{
-    static inline PAG* getPAG(bool buildFromFile = false)
+    static inline PAG* getPAG(bool buildFromFile = false, bool skipCreate = false)
```
richardlford, line 150: I intended to revert this back to the previous. I'll do another push.

include/Graphs/PAG.h
```diff
@@ -153,6 +153,13 @@ class PAG : public GenericGraph<PAGNode,PAGEdge>
         }
         return pag;
     }
+
+    /// Return pointer to PAG if it has been built, otherwise nullptr.
+    static inline PAG* tryGetPAG()
+    {
+        return pag;
+    }
+
```
yuleisui, line 162: I don't think we do need this method. You can make the option `IncludePAGInICFGDump` as false by default. It will solve your problem.

richardlford:
I'm closing this pull request and opening a new one as I was having trouble squishing it into a clean commit.

PR #400, opened 2021-03-26

richardlford:
4. It is possible one might want to view the ICFG before the
   Program Assignment Graph (PAG) has been built. However
   the ICFG stores some data related to the PAG and the
   graph writer for the ICFG tries to access the PAG.
   This change adds PAG::tryGetPAG to get the PAG if present,
   but skip the PAG processing if it is not present.
5. Add -include-pag-icfg-dump option to control PAG info in ICFG dump
   to preserve the current behavior. When the option is selected,
   the PAG pointer is obtained, if the PAG has been constructed,
   so the PAG information is added. If the PAG has not been constructed,
   the extra PAG information will not be present.

### DIFF
--- a/include/Graphs/PAG.h
+++ b/include/Graphs/PAG.h
@@ -153,6 +153,13 @@ public:
         }
         return pag;
     }
+
+    /// Return pointer to PAG if it has been built, otherwise nullptr.
+    static inline PAG* tryGetPAG()
+    {
+        return pag;
+    }
+
     static void releasePAG()
     {
         if (pag)
--- a/lib/Graphs/ICFG.cpp
+++ b/lib/Graphs/ICFG.cpp
@@ -27,6 +27,7 @@
  *      Author: Yulei Sui
  */
 
+#include <Util/Options.h>
 #include "SVF-FE/LLVMUtil.h"
 #include "Util/SVFModule.h"
 #include "Graphs/ICFG.h"
@@ -64,6 +65,9 @@ const std::string ICFGNode::toString() const {
     return rawstr.str();
 }
 
+void ICFGNode::dump() const {
+    outs() << this->toString() << "\n";
+}
 
 const std::string GlobalBlockNode::toString() const {
     std::string str;
@@ -397,6 +401,14 @@ void ICFG::dump(const std::string& file, bool simple)
 }
 
 /*!
+ * View ICFG
+ */
+void ICFG::view()
+{
+    llvm::ViewGraph(this, "Interprocedural Control-Flow Graph");
+}
+
+/*!
  * Update ICFG for indirect calls
  */
 void ICFG::updateCallGraph(PTACallGraph* callgraph)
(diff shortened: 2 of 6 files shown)


include/Graphs/PAG.h
```diff
@@ -153,6 +153,13 @@ class PAG : public GenericGraph<PAGNode,PAGEdge>
         }
         return pag;
     }
+
+    /// Return pointer to PAG if it has been built, otherwise nullptr.
+    static inline PAG* tryGetPAG()
+    {
+        return pag;
+    }
+
```
yuleisui, line 162: I don't think we do need this method. You can make the option IncludePAGInICFGDump as false by default. It will solve your problem.

lib/Graphs/ICFG.cpp
```diff
@@ -460,11 +472,16 @@ struct DOTGraphTraits<ICFG*> : public DOTGraphTraits<PAG*>
...
         {
             rawstr << "IntraBlockNode ID: " << bNode->getId() << " \t";
-            PAG::PAGEdgeList&  edges = PAG::getPAG()->getInstPTAPAGEdgeList(bNode);
-            for (PAG::PAGEdgeList::iterator it = edges.begin(), eit = edges.end(); it != eit; ++it)
-            {
-                const PAGEdge* edge = *it;
-                rawstr << edge->toString();
+            rawstr << value2String(bNode->getInst()) << " \t";
+            if (Options::IncludePAGInICFGDump) {
+                if (PAG* pag = PAG::tryGetPAG()) {
```
yuleisui, line 477: This branch condition will never be false since `PAG::tryGetPAG()` will always return a PAG.

include/Graphs/PAG.h
richardlford: It is needed because there is a small window (in the current code) between when the ICFG is made and when the PAG is created. In that interval the ICFG would exist, but the pag would be nullptr. If, in the debugger, you wanted to dump the ICFG, say, just after it has been created, to make sure it is ok, or just to see what it looks like, the ICFG dumper would be executing when the PAG is null, but you would not want the dumper to create the PAG (since a dumper should never have side effects like that). Hence the need for this method (and to avoid complicating the usual getPAG).

lib/Graphs/ICFG.cpp
richardlford: But if the PAG has not yet been created, what it will return is null, so it must be checked. That could happen during the window of time (mentioned above), between when the ICFG is created and the PAG is made.

include/Graphs/PAG.h
richardlford: I don't want to have IncludePAGInICFGDump be false because I want to see the PAG information when it becomes present. I just don't want the ICFG to blow up during the interval between when the ICFG is made and when the PAG shows up (I know in the current code that is a small interval, but one could imagine the code being modified so that the ICFG and PAG were not so closely linked).

lib/Graphs/ICFG.cpp
yuleisui: The `tryGetPAG ` is risky once exposed to users and might cause a null pointer. PAG is the most fundamental graph in SVF and it should always be available. 

I do not think we have a window here. The ICFG once built, the PAG should always available.  Do you mean someone may try to create an ICFG instance and then try to dump it inside the PAG constructor?
https://github.com/SVF-tools/SVF/blob/master/lib/Graphs/PAG.cpp#L265

If that is the case, I think we should consider making ICFG's constructor private and make the whole analysis share one ICFG similar to PAG.

richardlford:
I'm removing the tryGetPAG as you suggested. I'm closing the pull request and opening a new so as to get a clean diff.

PR #401, opened 2021-03-26

richardlford:
4. Add -include-pag-icfg-dump option to control PAG info in ICFG dump
   to preserve the current behavior.

### DIFF
--- a/lib/Graphs/ICFG.cpp
+++ b/lib/Graphs/ICFG.cpp
@@ -27,6 +27,7 @@
  *      Author: Yulei Sui
  */
 
+#include <Util/Options.h>
 #include "SVF-FE/LLVMUtil.h"
 #include "Util/SVFModule.h"
 #include "Graphs/ICFG.h"
@@ -64,6 +65,9 @@ const std::string ICFGNode::toString() const {
     return rawstr.str();
 }
 
+void ICFGNode::dump() const {
+    outs() << this->toString() << "\n";
+}
 
 const std::string GlobalBlockNode::toString() const {
     std::string str;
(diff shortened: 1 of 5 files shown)


lib/Graphs/ICFG.cpp
```diff
@@ -460,11 +472,14 @@ struct DOTGraphTraits<ICFG*> : public DOTGraphTraits<PAG*>
...
-            {
-                const PAGEdge* edge = *it;
-                rawstr << edge->toString();
+            rawstr << value2String(bNode->getInst()) << " \t";
+            if (Options::IncludePAGInICFGDump) {
+                PAG::PAGEdgeList&  edges = PAG::getPAG()->getInstPTAPAGEdgeList(bNode);
+                for (PAG::PAGEdgeList::iterator it = edges.begin(), eit = edges.end(); it != eit; ++it)
+                {
+                    const PAGEdge* edge = *it;
+                    rawstr << edge->toString();
```
yuleisui, line 481: One more thing, unfortunately,  `toString()` will also output the LLVM instruction, which makes the node label with duplicate instruction. See here: https://github.com/SVF-tools/SVF/blob/39385bbb18b4b11ebdd4be5b4290064f9200dca0/lib/Graphs/PAG.cpp#L53-L67

I would suggest make a `if` and `else` branch:
```
if (Options::IncludePAGInICFGDump) {
...
rawstr << edge->toString();
}
else{
rawstr << value2String(bNode->getInst()) << " \t";
}
```

Any better idea?

richardlford: What you have was needed, but also if there are no PAG edges the instruction must be added explicitly.

richardlford:
I'm closing this pull request and submitting another with the commits squished.

PR #403, opened 2021-03-28

richardlford:
   IntraBlockNodes if there were no PAG edges, or if the
   user selected not to show PAG information.
   This changed adds the needed code.
4. Add -include-pag-icfg-dump option to control PAG info in ICFG dump
   to preserve the current behavior.

### DIFF
--- a/lib/Graphs/ICFG.cpp
+++ b/lib/Graphs/ICFG.cpp
@@ -27,6 +27,7 @@
  *      Author: Yulei Sui
  */
 
+#include <Util/Options.h>
 #include "SVF-FE/LLVMUtil.h"
 #include "Util/SVFModule.h"
 #include "Graphs/ICFG.h"
@@ -64,6 +65,9 @@ const std::string ICFGNode::toString() const {
     return rawstr.str();
 }
 
+void ICFGNode::dump() const {
+    outs() << this->toString() << "\n";
+}
 
 const std::string GlobalBlockNode::toString() const {
     std::string str;
@@ -397,6 +401,14 @@ void ICFG::dump(const std::string& file, bool simple)
 }
 
 /*!
+ * View ICFG
+ */
+void ICFG::view()
+{
+    llvm::ViewGraph(this, "Interprocedural Control-Flow Graph");
+}
+
+/*!
  * Update ICFG for indirect calls
  */
 void ICFG::updateCallGraph(PTACallGraph* callgraph)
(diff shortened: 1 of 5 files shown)


lib/Graphs/ICFG.cpp
```diff
@@ -486,11 +509,13 @@ struct DOTGraphTraits<ICFG*> : public DOTGraphTraits<PAG*>
...
-            {
-                const PAGEdge* edge = *it;
-                rawstr << edge->toString();
+            if (Options::IncludePAGInICFGDump) {
+                PAG::PAGEdgeList&  edges = PAG::getPAG()->getInstPTAPAGEdgeList(glob);
+                for (PAG::PAGEdgeList::iterator it = edges.begin(), eit = edges.end(); it != eit; ++it)
+                {
+                    const PAGEdge* edge = *it;
+                    rawstr << edge->toString();
+                }
```
yuleisui, line 518: You forgot to handle the `else` branch here if `Options::IncludePAGInICFGDump` does not hold.

lib/Graphs/ICFG.cpp
```diff
@@ -460,11 +472,22 @@ struct DOTGraphTraits<ICFG*> : public DOTGraphTraits<PAG*>
...
+                rawstr << value2String(bNode->getInst()) << " \t";
+                PAG::PAGEdgeList::iterator it = edges.begin(), eit = edges.end();
+                if (it == eit) {
+                    // No edges, so we need to dump the instruction.
+                    rawstr << value2String(bNode->getInst()) << " \t";
+                }  else {
+                    for (; it != eit; ++it)
+                    {
+                        const PAGEdge* edge = *it;
+                        rawstr << edge->toString();
```
yuleisui, line 486: You have two `value2String` which may output duplicate instructions, making it unncessary hard for debugging.

Can you change it to something as below:

```
PAG::PAGEdgeList&  edges = PAG::getPAG()->getInstPTAPAGEdgeList(bNode);
if(edges.empty()){
       rawstr << value2String(bNode->getInst()) << " \t";
}
else {
       for (PAG::PAGEdgeList::iterator it = edges.begin(), eit = edges.end(); it != eit; ++it) {
                const PAGEdge* edge = *it;
                rawstr << edge->toString();
       }
}

```

richardlford:
Since, as you previously suggested, the PAG is central to SVF and the ICFG is only created when the PAG is created, I've decided the option to not dump PAG information when dumping the ICFG was not useful and only complicated things. So I've about to create a new pull request without that option.

PR #405, opened 2021-03-30
Improve Program Assignment Graph (PAG) Debugging experience

richardlford:
1. In the .dot dump of the PAG, and some other .dot dumps, change the
circle shapes to something different as the circles get really high
when the contents is wide. In particular made the following changes

circle         -> box
Mcircle        -> house
doublecircle   -> invhouse

2. Add a PAGNode::dump() method that will write
the value of PAGNode::toString() to outs().

3. Added PAG::view() method callable from within the
debugger that will pop up the formatted .dot dump,
provided the user configures xdg-open to display
.dot files.

### DIFF
--- a/lib/Graphs/ConsG.cpp
+++ b/lib/Graphs/ConsG.cpp
@@ -649,7 +649,7 @@ struct DOTGraphTraits<ConstraintGraph*> : public DOTGraphTraits<PAG*>
     static std::string getNodeAttributes(NodeType *n, ConstraintGraph*)
     {
         PAGNode* node = PAG::getPAG()->getPAGNode(n->getId());
-        return node->getNodeAttributes();
+        return node->getNodeAttrForDotDisplay();
     }
 
     template<class EdgeIter>
--- a/lib/Graphs/ConsG.cpp
+++ b/lib/Graphs/ConsG.cpp
@@ -649,40 +649,7 @@ struct DOTGraphTraits<ConstraintGraph*> : public DOTGraphTraits<PAG*>
     static std::string getNodeAttributes(NodeType *n, ConstraintGraph*)
     {
         PAGNode* node = PAG::getPAG()->getPAGNode(n->getId());
-
-        if (SVFUtil::isa<ValPN>(node))
-        {
-            if(SVFUtil::isa<GepValPN>(node))
-                return "shape=hexagon";
-            else if (SVFUtil::isa<DummyValPN>(node))
-                return "shape=diamond";
-            else
-                return "shape=box";
-        }
-        else if (SVFUtil::isa<ObjPN>(node))
-        {
-            if(SVFUtil::isa<GepObjPN>(node))
-                return "shape=doubleoctagon";
-            else if(SVFUtil::isa<FIObjPN>(node))
-                return "shape=septagon";
-            else if (SVFUtil::isa<DummyObjPN>(node))
-                return "shape=house";
-            else
-                return "shape=invhouse";
-        }
-        else if (SVFUtil::isa<RetPN>(node))
-        {
-            return "shape=Mrecord";
-        }
-        else if (SVFUtil::isa<VarArgPN>(node))
-        {
-            return "shape=octagon";
-        }
-        else
-        {
-            assert(0 && "no such kind node!!");
-        }
-        return "";
+        return node->getNodeAttributes();
     }
 
     template<class EdgeIter>
--- a/lib/Graphs/ConsG.cpp
+++ b/lib/Graphs/ConsG.cpp
@@ -657,7 +657,7 @@ struct DOTGraphTraits<ConstraintGraph*> : public DOTGraphTraits<PAG*>
             else if (SVFUtil::isa<DummyValPN>(node))
                 return "shape=diamond";
             else
-                return "shape=circle";
+                return "shape=box";
         }
         else if (SVFUtil::isa<ObjPN>(node))
         {
@@ -666,9 +666,9 @@ struct DOTGraphTraits<ConstraintGraph*> : public DOTGraphTraits<PAG*>
             else if(SVFUtil::isa<FIObjPN>(node))
                 return "shape=septagon";
             else if (SVFUtil::isa<DummyObjPN>(node))
-                return "shape=Mcircle";
+                return "shape=house";
             else
-                return "shape=doublecircle";
+                return "shape=invhouse";
         }
         else if (SVFUtil::isa<RetPN>(node))
         {
--- a/lib/Graphs/PAG.cpp
+++ b/lib/Graphs/PAG.cpp
@@ -50,6 +50,45 @@ const std::string PAGNode::toString() const {
     return rawstr.str();
 }
 
+/// Get shape and/or color of node for .dot display.
+const std::string PAGNode::getNodeAttributes() const {
+    // TODO: Maybe use over-rides instead of these ifs,
+    // But this puts them conveniently together.
+    if (SVFUtil::isa<ValPN>(this))
+    {
+        if(SVFUtil::isa<GepValPN>(this))
+            return "shape=hexagon";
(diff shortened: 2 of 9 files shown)


lib/Graphs/ConsG.cpp
```diff
@@ -666,9 +666,9 @@ struct DOTGraphTraits<ConstraintGraph*> : public DOTGraphTraits<PAG*>
             else if(SVFUtil::isa<FIObjPN>(node))
                 return "shape=septagon";
             else if (SVFUtil::isa<DummyObjPN>(node))
-                return "shape=Mcircle";
+                return "shape=house";
             else
-                return "shape=doublecircle";
+                return "shape=invhouse";
```
yuleisui, line 671: How about using `pentagon`, `octagon`, and `septagon` for three types of objects?

yuleisui: Maybe `doubleoctagon`, `pentagon` and `octagon`?

yuleisui: Hi Richard,

Thanks for your summary. Yes, it would be good to distinguish them and make them more readable.
Add a new `getNodeAttributes() const` would be better. Maybe the first option from you is good enough.

I am fine with these shapes 
```
circle (other ValPN) -> box
septagon (FIObjPN) -> box3d
Mcircle (DummyObjPN) -> tab
```

For `GetObjPN` (other ObjPN), can we use `component` which is also good for semantic purposes, since GetObjPN is a field `object` of its base `FIObjPN`.

For node colors, it could make different types more distinguishable, but we also have colors on edges. It might be too fancy for the graph unless you can show us a good example.

yuleisui:
Once the shapes are done, could you please also help with updating the new figures here:

https://github.com/svf-tools/SVF/wiki/Analyze-a-Simple-C-Program#5-pag

richardlford:
Yes, I can update the wiki. In my comment above I asked some questions. Should I take from your comment that I can go ahead with my preferred solution, or do you want to answer the questions. Since we are on opposite sides of the world, I'll wait until my morning to complete this by which time you will have had time to respond to the questions.

yuleisui:
> Yes, I can update the wiki. In my comment above I asked some questions. Should I take from your comment that I can go ahead with my preferred solution, or do you want to answer the questions. Since we are on opposite sides of the world, I'll wait until my morning to complete this by which time you will have had time to respond to the questions.

Please go ahead with your preferred solutions.

lib/Graphs/PAG.cpp
```diff
@@ -50,6 +50,45 @@ const std::string PAGNode::toString() const {
     return rawstr.str();
 }
 
+/// Get shape and/or color of node for .dot display.
+const std::string PAGNode::getNodeAttributes() const {
```
yuleisui, line 54: If this is only used in `DOTGraphTraits`. I would suggest putting only in that class.

lib/Graphs/PAG.cpp
```diff
@@ -58,6 +97,7 @@ const std::string ValPN::toString() const {
     std::string str;
     raw_string_ostream rawstr(str);
     rawstr << "ValPN ID: " << getId();
+    rawstr << "\n";
```
yuleisui, line 100: How about making it only one line? In case If someone would like to debug in the terminal

lib/Graphs/PAG.cpp
richardlford: I'm not sure how to do that. Perhaps that can wait for later. There are several specializations of DOTGraphTraits and I'm not sure how to make one instantiation make use of another's instance.

lib/Graphs/PAG.cpp
yuleisui: May I know why you add `\n` here? Keep it one line would be good for debugging in terminal.

richardlford: The new-line is needed to make the .dot output look nicer (lines not too long).

yuleisui: But the first line will be too short. Can we make it as one line for now? You can add brackets `[]` for nodeID or for instruction to make them more distinguishable?

lib/Graphs/PAG.cpp
richardlford: OK. I'm renaming is as you suggested (in my next push).

lib/Graphs/PAG.cpp
richardlford: OK. I'm making it one line (in my next push).

richardlford:
I've pushed the changes you requested. When ValPN::toString() is used by the .dot display routines, it is preceded in the line by the name of the function, so the two-line version is better for the .dot display, but the one-line version will be satisfactory. It just spreads the dump out more horizontally.

richardlford:
If you approve of these changes and want me to squish the changes into a single commit, let me know.

lib/Graphs/PAG.cpp
```diff
@@ -62,6 +105,7 @@ const std::string ObjPN::toString() const {
     std::string str;
     raw_string_ostream rawstr(str);
     rawstr << "ObjPN ID: " << getId();
+    rawstr << "\n";
```
yuleisui, line 108: Please remove this new line.

richardlford: @yuleisui, I have removed the newline in ObjPN::toString and pushed.

lib/Graphs/PAG.cpp
```diff
@@ -70,6 +113,7 @@ const std::string GepValPN::toString() const {
     std::string str;
     raw_string_ostream rawstr(str);
     rawstr << "GepValPN ID: " << getId() << " with offset_" + llvm::utostr(getOffset());
+    rawstr << "\n";
```
yuleisui, line 116: I mean all the newlines should be removed for `toString` methods since we need to debug in the terminal. Two lines make it a bit tricky.

yuleisui: Hi Richard, maybe could you please share the figures before and after removing the `\n`, so that we can know which way would be better? Thanks,

yuleisui:
> If you approve of these changes and want me to squish the changes into a single commit, let me know.

We can see which way is better if you could share both pictures. This change will affect many other SVF users when they look at dot or debugging in terminal.

richardlford:
I'm on vacation today, but when I get back to work I'll post the PAG display with and without the newline.

richardlford:
Here are PAG dumps for the swap program, in shorter and longer format.

- [pag_final_longer.dot.pdf](https://github.com/SVF-tools/SVF/files/6272877/pag_final_longer.dot.pdf)
- [pag_final_shorter.dot.pdf](https://github.com/SVF-tools/SVF/files/6272881/pag_final_shorter.dot.pdf)

I've also pushed a change to make the newlines that make the lines shorter conditional on an option. A possible alternative is to just use the --dump-pag option instead of a new option.

yuleisui:
Thanks Richard, the shorter version looks good to me. Please let me know whether the current pull request is your shorter version?

yuleisui:
Shorter version could be used by default in your patch.

lib/Util/Options.cpp
```diff
@@ -255,6 +255,12 @@ namespace SVF
         llvm::cl::desc("Dump dot graph of PAG")
     );
 
+    const llvm::cl::opt<bool> Options::PAGDotGraphShorter(
+            "dump-pag-shorter",
+            llvm::cl::init(false),
```
yuleisui, line 260: It looks pretty good to me for the shorter version. Please make it as default.

richardlford: I have made the shorter version the default.

yuleisui:
Merged. Please also help to update the wiki page with your new PAG and Constraint Graph.
https://github.com/svf-tools/SVF/wiki/Analyze-a-Simple-C-Program

PR #410, opened 2021-04-08
Update program assignment graph and constraint graph images

richardlford:
This change updates the images in the wiki for the simple C program (swap), for the program assignment graph (PAG) and the constraint graph.


richardlford:
After this is merged I'll add some additional explanatory text in the Wiki.

richardlford:
I have updated the descriptions of the PAG and Constaint Graph on the Wiki
to reflect the new shapes and the larger variety of colors for edges (not
previously documented).

PR #413, opened 2021-04-12
Replace std::hash<std::pair> with SVF::svf_hash

gerion0:
Overloading std::hash for the usage of std::pair (or more generally
another STL class) can conflict when used with other libraries that
do the same.

This commit workarounds this with a special class that fallbacks to
std::hash for all classes except std::pair and uses a special
implementation otherwise.

Since I have used `if constexpr`, this commit raises the C++ standard to C++17.
`if constexpr` is not needed (technically) and can be workarounded with
standard template metaprogramming. However, I find it more readable as is.

In my case the overloading of `std::hash<std::pair>` conflicts with [graph-tool](https://git.skewed.de/count0/graph-tool/-/blob/master/src/graph/graph_util.hh#L451)
but of course, this is only an example.

### DIFF
--- a/include/MemoryModel/LocationSet.h
+++ b/include/MemoryModel/LocationSet.h
@@ -290,7 +290,7 @@ private:
 
 template <> struct std::hash<SVF::LocationSet> {
     size_t operator()(const SVF::LocationSet &ls) const {
-        std::hash<std::pair<SVF::Size_t, SVF::Size_t>> h;
+        SVF::Hash<std::pair<SVF::Size_t, SVF::Size_t>> h;
         return h(std::make_pair(ls.getOffset(), ls.getByteOffset()));
     }
 };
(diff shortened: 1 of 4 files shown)


yuleisui:
Can I assume the performance would be the same as std hash?

gerion0:
Yes, the performance is the same, see this [godbolt](https://godbolt.org/z/znTe3qrrh). Everything except of the szudzik function is optimized away (even with -O1).

gerion0:
I have simplified the code by direct overloading of `svf_hash`:
- no need for C++17 anymore.
- shorter code

Normal usage of std::hash is optimized away [as before](https://godbolt.org/z/vfjGncMvW).

yuleisui:
Thanks for that. I would not mind merging it if @mbarbar is ok since the hash was implemented by @mbarbar.

mbarbar:
Good idea. I would prefer if its name was CamelCase and without the `svf_` prefix as `SVF::svf_` is abit redundant, so just `Hash`. Otherwise, looks good to me.

gerion0:
Done (no strong opinion about the name).

PR #414, opened 2021-04-12
Improve debugging of call graph and value flow graph

richardlford:
1. Added PTACallGraph::view(), callable from debugger,
to display the call graph while debugging.

2. Added VFG::view(), callable from debugger,
to display the [sparse] value flow graph while debugging.

3. Use value2String in a few places to get the special
treatment for functions that it gives (instead of dumping the
IR for the whole function).

4. There were missing cases when dumping a sparse value flow
graph, resulting in assertion violations. I added code to
cover the misssing cases. You may want to check that
the colors and styles I selected are what you want.

5. I modified dumpMSSA() so it unconditionally dumps the
memory SSA without checking for Options::DumpMSSA.
In the sole place where it had been called I put the
Options::DumpMSSA check. This will allow dumpMSSA() to
be called from the debugger if desired without concern
for the setting of Options::DumpMSSA.

### DIFF
--- a/include/Graphs/PTACallGraph.h
+++ b/include/Graphs/PTACallGraph.h
@@ -437,6 +437,9 @@ public:
 
     /// Dump the graph
     void dump(const std::string& filename);
+
+    /// View the graph from the debugger
+    void view();
 };
 
 } // End namespace SVF
--- a/include/Graphs/VFG.h
+++ b/include/Graphs/VFG.h
@@ -160,6 +160,9 @@ public:
     /// Dump graph into dot file
     void dump(const std::string& file, bool simple = false);
 
+    /// Dump graph into dot file
+    void view();
+
     /// Update VFG based on pointer analysis results
     void updateCallGraph(PointerAnalysis* pta);
 
--- a/lib/Graphs/PTACallGraph.cpp
+++ b/lib/Graphs/PTACallGraph.cpp
@@ -316,6 +316,10 @@ void PTACallGraph::dump(const std::string& filename)
       GraphPrinter::WriteGraphToFile(outs(), filename, this);
 }
 
+void PTACallGraph::view()
+{
+    llvm::ViewGraph(this, "Call Graph");
+}
 
 namespace llvm
 {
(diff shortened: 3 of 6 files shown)


yuleisui:
Would you mind posting a pdf of the dot graph? Please also update the SVFG graph on SVF's wiki once after the merge.

richardlford:
> Would you mind posting a pdf of the dot graph? Please also update the SVFG graph on SVF's wiki once after the merge.

@yuleisui, yes, I will, but I decided I would like to study the VFG and SVFG a little more first. 

I have a question. An alloca instruction, or an external call to something like malloc that allocates storage is represented by both a ValSym, for the address returned, and by an ObjSym, to represent the allocated memory. In the memory SSA representation, it seems like the allocation site should be considered a definition, with the value assigned being "undefined", or perhaps some external allocating functions might return zero-initialized memory. I'm still studying this out, but it seemed we might want to add something like an AllocCHI that would derive from MSSACHI, and that would represent the definition point for the object at the point it is allocated. Would this make sense? I'm just thinking about the case where we want to detect uninitialized dynamically-allocated variables. If the definition at the allocation point reached a use, then the use would be of an uninitialied variable. Let me know what you think. Thanks. I'll reply more later after further study.

yuleisui:
> > Would you mind posting a pdf of the dot graph? Please also update the SVFG graph on SVF's wiki once after the merge.
> 
> @yuleisui, yes, I will, but I decided I would like to study the VFG and SVFG a little more first.
> 
> I have a question. An alloca instruction, or an external call to something like malloc that allocates storage is represented by both a ValSym, for the address returned, and by an ObjSym, to represent the allocated memory. In the memory SSA representation, it seems like the allocation site should be considered a definition, with the value assigned being "undefined", or perhaps some external allocating functions might return zero-initialized memory. I'm still studying this out, but it seemed we might want to add something like an AllocCHI that would derive from MSSACHI, and that would represent the definition point for the object at the point it is allocated. Would this make sense? I'm just thinking about the case where we want to detect uninitialized dynamically-allocated variables. If the definition at the allocation point reached a use, then the use would be of an uninitialied variable. Let me know what you think. Thanks. I'll reply more later after further study.

Yes CHI operator means the definition of an object. We have a CHI after each heap allocation (e.g., malloc), but not alloca instruction at this stage.

PR #419, opened 2021-04-19
Fix DOT graphs, debugging aids

richardlford:
1. Some of the DOT graph writers were using "style=double" which the dot program says is invalid. I replaced this with "penwidth=2" which I believe gives the desired effect.
2. I added an "isNodeHidden" dot trait function for the SVFG dot graphs to inhibit outputting nodes which no input or output edges. These are uninteresting and just end up compressing the rest of the graph.
3. While studying the pointer analysis in the debugger, I wanted to get a dump of the pointed-to sets, and also, lists of such. I added SVFUtil::dumpSparseSet and SVFUtil::dumpPointsToList. These needed to take a reference as parameter as gdb was not able to construct a copy. To avoid type-mismatch errors for this I added SVFUtil::equalPointsTo and use in MemRegion.h when declaring sets or maps of Points-to sets.

### DIFF
--- a/include/MSSA/MemRegion.h
+++ b/include/MSSA/MemRegion.h
@@ -141,9 +141,9 @@ public:
     ///Define mem region set
     typedef OrderedSet<const MemRegion*, MemRegion::equalMemRegion> MRSet;
     typedef Map<const PAGEdge*, const SVFFunction*> PAGEdgeToFunMap;
-    typedef OrderedSet<PointsTo, MemRegion::equalPointsTo> PointsToList;
+    typedef OrderedSet<PointsTo, SVFUtil::equalPointsTo> PointsToList;
     typedef Map<const SVFFunction*, PointsToList > FunToPointsToMap;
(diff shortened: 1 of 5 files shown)


richardlford:
@yuleisui, I'm also ready to make Wiki changes to update the graphs. I notice that currently, the WIKI references images that are in the main repository. It seems like it would be better if they referenced images in the Wiki repository. Someone changing them would then clone the wiki repository like this,

`git clone https://github.com/SVF-tools/SVF.wiki.git`

Then add or modify the images, commit and push. If changing the text of the wiki does not require write access to the main repository, then it seems that images referenced from the wiki should also not require it.

So, is it OK if I move the images for the Simple C program, at least, to the Wiki repository, rather than the main repository. Of course, I'll not submit a pull request to delete them from the main repository until they have been moved and updated in the wiki repository.

Rich

yuleisui:
Thanks, Richard. Please move wiki images to `git clone https://github.com/SVF-tools/SVF.wiki.git`.

PR #421, opened 2021-04-21
WPAPass: Used -opt-svfg to control creation of optimized SVFG.

richardlford:
Currently a command like

```
wpa -ander -svfg -dump-svfg swap.ll
```

produces an unoptimized SVFG. With this change, a command like this:

```
wpa -ander -svfg --dump-vfg -opt-svfg=false swap.ll
```

will produce an optimized SVFG. Note that the sense of
-opt-svfg seems backward. Should we change this?

This was needed to produced the optimized dot graph of the SVFG
for the wiki.

### DIFF
--- a/lib/WPA/WPAPass.cpp
+++ b/lib/WPA/WPAPass.cpp
@@ -153,7 +153,15 @@ void WPAPass::runPointerAnalysis(SVFModule* svfModule, u32_t kind)
     {
         SVFGBuilder memSSA(true);
         assert(SVFUtil::isa<AndersenBase>(_pta) && "supports only andersen/steensgaard for pre-computed SVFG");
-        SVFG *svfg = memSSA.buildFullSVFGWithoutOPT((BVDataPTAImpl*)_pta);
+        SVFG *svfg;
+        if (Options::OPTSVFG)
+        {
+            // This seems backward, but the description says if true, then unoptimized.
+            svfg = memSSA.buildFullSVFGWithoutOPT((BVDataPTAImpl*)_pta);
+        } else
+        {
+            svfg = memSSA.buildFullSVFG((BVDataPTAImpl*)_pta);
+        }
         /// support mod-ref queries only for -ander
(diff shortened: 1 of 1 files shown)


mbarbar:
I think the semantics of `opt-svfg` should be cleared up (probably the description). This patch is at odds with the understanding of `opt-svfg` in `FlowSensitive.cpp` (which understands it the non-backwards way you alluded to):

https://github.com/SVF-tools/SVF/blob/84be61108ade5ca541c5fd8f0c821dd42899385d/lib/WPA/FlowSensitive.cpp#L53-L56

yuleisui:
Yes, agree. We will need a separate option for SVFG used in FlowSensitive analysis.

richardlford:
Could you explain the original purpose of the opt-svfg option. In thinking
about it, it occurs to me that some algorithms might like the original full
svfg before the "optimization".

yuleisui:
Hi Rich,

The SVFG generated by the following four methods https://github.com/SVF-tools/SVF/blob/45460c79a7e8843f54bc0432e0f5bd6e5758e00e/lib/MSSA/SVFGBuilder.cpp#L44-L62 
can be used for flow-sensitive pointer analysis or purely used for dumping the graph for bug detection.

The `Options::OPTSVFG` after the merge is now used for both purposes, which I think @mbarbar might want it to be separated?

mbarbar:
Not quite what I meant.

As I understand, this is what's going on:
* `-opt-svfg` means build the SVFG *with* optimisations. `-opt-svfg=false` means build the SVFG *without* optimisations.
* The description for `-opt-svfg` implies otherwise. I think this is a documentation error. I think this is what @richardlford means by "backwards".
* `FlowSensitive` understands things according to my first point, and this patch understands things according to the opposite.

I think one of two choices:
* Make all users of `opt-svfg` understand it the same way: either as understood by `FlowSensitive` (in which case the description needs to change), or as understood in this patch (in which case the option name should change).
* Separate the options for `FlowSensitive` and other SVFG users as suggested by @yuleisui .

In either case, the name or the description should change.

PR #427, opened 2021-04-26
Move tests to cmake

lukekoko:
Moved tests from travis to cmake. When building the project cmake will check if the Test-suite folder exists. if it does it will build the test bc files and add testing to cmake.


yuleisui:
It seems that ctir has a problem on mac os? It should run ctir on mac?

https://travis-ci.org/github/SVF-tools/SVF/jobs/768408109#L575

PR #430, opened 2021-04-27
Enable DOT dumping of connected subgraphs

richardlford:
Sometimes SVF graphs contain a main part that is connected
and other parts that are not. The other parts might be part
of the testing infrastructure that are not used by the
current test. When a DOT dump of the graph is produced, these
other unconnected parts make the overall graph harder to
read as they typically make the dump very wide and the
nodes small. In order to focus on what is important,
this change enables dumping of connected subgraphs.

If --dot-largest-subgraph is specified, then only the nodes
in the largest connected subgraph will be dumped and the
other nodes will be ignored. Often this is what you will want.

If --dot-separate-subgraphs is specified, then each connected
subgraph will be dumped in a different file. The files will
have names like, e.g., pag_final_size48_id4.dot. This means
that this is the dump of the program assignment graph in its
final form, and that subgraph 4 has 48 nodes in it. We include
the size in the name so you can pick out the subgraphs most
likely to be interesting.

Sometimes the dump of a connected subgraph will appear not
to be connected. That is because some of the nodes have been
requested to be hidden. If you want to see them, use the
--dot-show-all option.

In order to implement the above the GenericGraph was
augmented and restructured somewhat. We needed to create
the connected subgraphs and track their sizes (so we
can pick the largest, and also use their size in the
names of the dump files). We also needed to track the
_current subgraph_ of a graph so that when dumping we
can ignore the nodes not in the current subgraph.

The ignoring is implemented using isNodeHidden method in
the DOTGraphTraits. Because this method only has access to
the node and we also needed access to the graph, we
added a pointer to the containing graph in each graph node.
This pointer is filled in when the node is added to the graph
and cleared if the node is removed from the graph.

In particular, to make the subgraph information accessible, the fields
of GenericNodes and GenericGraphs that do not depend on the generic
parameters are factored out into GenericGraphBase and
GenericNodeBase. I also gave a name (equalGEdge_struct) to one
anonymous struct since CLION was complaining that it was not
compliant.

The logic of checking the --dot-* options when deciding whether to
hide a node is centralized in the GenericNodeBase::reallyHideNode
method (defined in SVFUtil.cpp).

The logic to actually produce the selective dot dumps is in the new
GraphPrinter::SelectiveWriteGraphToFile method. This now also includes
the option to immediate view the dot file. This feature is now used by
the view() methods.  One problem I encountered was that I hit
compilation errors when I tried to include Options.h into
GraphPrinter.h because of the fact that Options.h includes
PointerAnalysis.h and WPAPass.h in order to access their enumeration
types and the result is a tangle of includes. To solve this for the
present, I've introduced SimpleOptions.{h,cpp} for options that do not
need to include any other complicating include files. I've placed the
new options in those files. We could consider moving other simple
options there. Alternately, if the enumeration types in
PointerAnalysis and WPAPass were in separate header files we might be
able to avoid these problems.

The GraphPrinter::WriteGraphToFile now returns the name of the file if
it was successful so that it can be immediately displayed if
requested.

GraphPrinter.h is no longer included in BasicTypes.h as that was
leading to some compilation errors. Now it is only included by
clients that need it.

The dump and view methods of the various graphs were modified to use
the above facilities, and graphs missing an isNodeHidden methods were
given one so that the subgraph facility would work for them.

### DIFF
--- a/include/Graphs/GenericGraph.h
+++ b/include/Graphs/GenericGraph.h
@@ -31,9 +31,14 @@
 #define GENERICGRAPH_H_
 
 #include "Util/BasicTypes.h"
+#include "Util/WorkList.h"
+#include <llvm/ADT/GraphTraits.h>
 
 namespace SVF
 {
+
+typedef u32_t SubgraphIdTy;
+class GenericGraphBase;
 
 /*!
  * Generic edge on the graph as base class
@@ -90,12 +95,13 @@ public:
     {
         return dst;
     }
+
     //@}
 
     /// Add the hash function for std::set (we also can overload operator< to implement this)
     //  and duplicated elements in the set are not inserted (binary tree comparison)
     //@{
-    typedef struct
+    typedef struct equalGEdge_struct
     {
         bool operator()(const GenericEdge<NodeType>* lhs, const GenericEdge<NodeType>* rhs) const
         {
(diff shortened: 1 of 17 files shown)


yuleisui:
Hi Richard,

The patch is quite a lot this time. Would you be able to split it into 2 or 3 for us to review and discuss?

Thanks,

richardlford:
> Hi Richard,
> 
> The patch is quite a lot this time. Would you be able to split it into 2 or 3 for us to review and discuss?
> 
> Thanks,
 @yuleisui, yes. And with the recent upgrade to LLVM 12.0 there is a need for some other changes. Some questions:

1. Do you think this general functionality is something useful?
2. I'm thinking there might be a different way to implement things that would not require the GenericGraph refactoring. Would you prefer something that did not complicate GenericGraph?
3. I think there is some need of include file cleanup in SVF. The LLVM coding and programming standards require that each include file be able to cleanly compile in isolation. Some of the problems I had with this change lead me to believe that not all SVF include files follow this rule.
4. If the enumeration types referenced by Options.h were defined in separate header files, it would eliminate the need for extra includes in the Options.h file. Do you view something like this as desirable?
Do you have any other general comments about things you like or dislike about this change?

yuleisui:
> > Hi Richard,
> > The patch is quite a lot this time. Would you be able to split it into 2 or 3 for us to review and discuss?
> > Thanks,
> 
> @yuleisui, yes. And with the recent upgrade to LLVM 12.0 there is a need for some other changes. Some questions:
> 
> 1. Do you think this general functionality is something useful?
I think we can keep the GenericGraph and GenericNode as they are now and not being too many layers.

> 2. I'm thinking there might be a different way to implement things that would not require the GenericGraph refactoring. Would you prefer something that did not complicate GenericGraph?
I prefer not to make it more complicated..

> 3. I think there is some need of include file cleanup in SVF. The LLVM coding and programming standards require that each include file be able to cleanly compile in isolation. Some of the problems I had with this change lead me to believe that not all SVF include files follow this rule.
Yes, that is a good idea.

> 4. If the enumeration types referenced by Options.h were defined in separate header files, it would eliminate the need for extra includes in the Options.h file. Do you view something like this as desirable?
Yes, I agree, please go ahead.

>    Do you have any other general comments about things you like or dislike about this change?
Can we make `creating connected graph` a separated method or class and accepts a graph as its input?

There is a problem when conducting analysis on top of the connected graph since we artificially introducing edges and nodes would that affect the precision or correctness of an analysis?

richardlford:
Will revisit in other PRs.

PR #432, opened 2021-04-28
Fix CI pass error

sherlly:
1. fix OS version conflict
2. fix type casting error in GEPTypeBridgeIterator

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -12,7 +12,7 @@
 SVFHOME=$(pwd)
 sysOS=`uname -s`
 MacLLVM="https://github.com/llvm/llvm-project/releases/download/llvmorg-12.0.0/clang+llvm-12.0.0-x86_64-apple-darwin.tar.xz"
-UbuntuLLVM="https://github.com/llvm/llvm-project/releases/download/llvmorg-12.0.0/clang+llvm-12.0.0-x86_64-linux-gnu-ubuntu-20.04.tar.xz"
+UbuntuLLVM="https://github.com/llvm/llvm-project/releases/download/llvmorg-12.0.0/clang+llvm-12.0.0-x86_64-linux-gnu-ubuntu-16.04.tar.xz"
 MacZ3="https://github.com/Z3Prover/z3/releases/download/z3-4.8.8/z3-4.8.8-x64-osx-10.14.6.zip"
 UbuntuZ3="https://github.com/Z3Prover/z3/releases/download/z3-4.8.8/z3-4.8.8-x64-ubuntu-16.04.zip"
 MacCTIR="https://github.com/mbarbar/ctir/releases/download/ctir-10.c3/ctir-clang-v10.c3-macos10.15.zip"
--- a/include/SVF-FE/GEPTypeBridgeIterator.h
+++ b/include/SVF-FE/GEPTypeBridgeIterator.h
@@ -69,11 +69,19 @@ public:
 
     Type *getIndexedType() const
     {
+        assert(false && "needs to be refactored");
         if ( CurTy.getInt() )
             return CurTy.getPointer();
 #if LLVM_VERSION_MAJOR >= 11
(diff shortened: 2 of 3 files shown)


build.sh
```diff
@@ -12,7 +12,7 @@
 SVFHOME=$(pwd)
 sysOS=`uname -s`
 MacLLVM="https://github.com/llvm/llvm-project/releases/download/llvmorg-12.0.0/clang+llvm-12.0.0-x86_64-apple-darwin.tar.xz"
-UbuntuLLVM="https://github.com/llvm/llvm-project/releases/download/llvmorg-12.0.0/clang+llvm-12.0.0-x86_64-linux-gnu-ubuntu-20.04.tar.xz"
+UbuntuLLVM="https://github.com/llvm/llvm-project/releases/download/llvmorg-12.0.0/clang+llvm-12.0.0-x86_64-linux-gnu-ubuntu-16.04.tar.xz"
```
yuleisui, line 15: Any problem with 20.04?

include/SVF-FE/GEPTypeBridgeIterator.h
```diff
@@ -72,8 +72,16 @@ class generic_bridge_gep_type_iterator : public std::iterator<std::forward_itera
...
+        Type * CT = CurTy.getPointer();
+        auto * ST = dyn_cast<StructType>(CT);
+        auto Array = dyn_cast<ArrayType>(CT);
+        auto Vector = dyn_cast<VectorType>(CT);
+        if (ST)
+            return ST->getTypeAtIndex(getOperand());
+        else if (Array)
+            return Array->getElementType();
+        else if (Vector)
+            return Vector->getElementType();
```
yuleisui, line 84: do we need a `else` branch here?

include/SVF-FE/GEPTypeBridgeIterator.h
```diff
@@ -102,7 +110,17 @@ class generic_bridge_gep_type_iterator : public std::iterator<std::forward_itera
...
+            auto Array = dyn_cast<ArrayType>(CT);
+            auto Vector = dyn_cast<VectorType>(CT);
+            if (ST)
+                CurTy.setPointer(ST->getTypeAtIndex(getOperand()));
+            else if (Array)
+                CurTy.setPointer(Array->getElementType());
+            else if (Vector)
+                CurTy.setPointer(Vector->getElementType());
+            else
+                CurTy.setPointer(nullptr);
```
yuleisui, line 123: will this be set nullptr?

build.sh
sherlly: > Any problem with 20.04?
I think the problem is that the latest CTIR only support Ubuntu 18.04 (which needs libtinfo.so.5 ), while the latest LLVM support Ubuntu 20.04 (which needs libtinfo.so.6). See in https://travis-ci.org/github/SVF-tools/SVF/jobs/768682324#L905.

include/SVF-FE/GEPTypeBridgeIterator.h
sherlly: The old CompositeType support three type (ArrayType, StructType and VectorType), so I think maybe the else branch is unreachable?

sherlly: Anyway I can add a else branch such as `return CT`.

build.sh
yuleisui: I see. it might cause a problem, because, the generated svflib in travis will be published to `npm`. Can some projects inside a ubuntu 20.04 OS still use the svflib built under a 16.04 version?

include/SVF-FE/GEPTypeBridgeIterator.h
yuleisui: then please add an assertion in the else branch.

PR #441, opened 2021-05-23
Improve VSFS performance

mbarbar:
This PR introduces two changes:

**BFS instead of DFS for address-taken propagation**
As discussed, propagation is now breadth-first. Since there is no way to add a _propagate o:v to o:v'_ constraint to the solver, I encode this as a dummy node, which holds this information and is pushed into the worklist as required. Otherwise, `propagateVersion` is no longer recursive and some `stmtReliance` stuff is pushed out to `processStore` (previously implied by `propagateVersion`).

**Ignoring existing constraints for call-graph updates**
The changes in `updateConnectedNode` ensure that we don't propagate `o:v` to `o:v'` if it has been done before. `o:v` is not necessarily new to require propagation if a reliance on `o:v` has previously been established.

`vis` is the smallest bitcode I noticed the regression in. It has (dense):
Old vfspta: 25s, new vfspta: 5s-6s, fspta: 15s
`bash` is the largest BC I have on hand that I could run on my 8 GB machine. It has (dense):
Old vfspta: 130s, new vfspta: 65s, fspta: starts swapping too much
A lot of smaller bitcode which I tested didn't really see any significant difference.

Unfortunately, with the server still down I can't test on larger bitcode.

### DIFF
--- a/include/Graphs/SVFGNode.h
+++ b/include/Graphs/SVFGNode.h
@@ -434,6 +434,7 @@ private:
  */
 class DummyVersionPropSVFGNode : public VFGNode
 {
+public:
     DummyVersionPropSVFGNode(NodeID id, NodeID object, Version version)
         : VFGNode(id, DummyVProp), object(object), version(version)
     { }
--- a/include/Graphs/SVFGNode.h
+++ b/include/Graphs/SVFGNode.h
@@ -429,6 +429,42 @@ private:
     const CallBlockNode* callInst;
 };
 
+/*
+ * Dummy node which encodes propagation of an object/version pair.
+ */
+class DummyVersionPropSVFGNode : public VFGNode
+{
+    DummyVersionPropSVFGNode(NodeID id, NodeID object, Version version)
+        : VFGNode(id, DummyVProp), object(object), version(version)
+    { }
+
+    NodeID getObject(void) const { return object; }
+    Version getVersion(void) const { return version; }
+
+    /// Methods to support type inquiry through isa, cast, and dyn_cast:
+    //@{
+    static inline bool classof(const DummyVersionPropSVFGNode *)
+    {
+        return true;
+    }
+
+    static inline bool classof(const VFGNode *node)
+    {
+        return node->getNodeKind() == DummyVProp;
+    }
+
+    static inline bool classof(const GenericVFGNodeTy *node)
+    {
+        return node->getNodeKind() == DummyVProp;
+    }
+    //@}
+
+private:
+    const NodeID object;
+    const Version version;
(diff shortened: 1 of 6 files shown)


include/Graphs/SVFGNode.h
```diff
@@ -429,6 +429,43 @@ class InterMSSAPHISVFGNode : public MSSAPHISVFGNode
     const CallBlockNode* callInst;
 };
 
+/*
+ * Dummy node which encodes propagation of an object/version pair.
+ */
+class DummyVersionPropSVFGNode : public VFGNode
```
yuleisui, line 435: Can we make sure that the users won't use this type of node and it can only be used in VSFS?  Just wondering any way we can make it as a subtype of `MPhil` or `MIntraPhi`?

include/Graphs/VFGNode.h
```diff
@@ -55,7 +55,7 @@ class VFGNode : public GenericVFGNodeTy
     {
         Addr, Copy, Gep, Store, Load, Cmp, BinaryOp, UnaryOp, TPhi, TIntraPhi, TInterPhi,
         MPhi, MIntraPhi, MInterPhi, FRet, ARet, AParm, FParm,
-        FunRet, APIN, APOUT, FPIN, FPOUT, NPtr
+        FunRet, APIN, APOUT, FPIN, FPOUT, NPtr, DummyVProp
```
yuleisui, line 58: Maybe, we can put this class just under the VSFS class instead here? Will we use it anywhere else?

mbarbar: It's not used elsewhere, but `SVFG` needs to know about it. The FS worklist takes unsigned ints (IDs), and while we can technically just give it an ID that we keep track of without creating an SVFG node, we run the risk of using an integer that exists as an ID in the SVFG, causing bugs that will, probably, be terrible to debug.

include/Graphs/SVFGNode.h
mbarbar: It can be, technically. Though there might be a bit of confusion, for example, what would fields like `cpts`, `res`, and `opVers` be? They'll just sit uninitialised. We'll also have to add our own fields which are irrelevant to `MPhi`/`MIntraPhi` nodes.

include/Graphs/VFGNode.h
yuleisui: I see.

include/Graphs/SVFGNode.h
yuleisui: Correct. It might not be good to subclass them.

PR #443, opened 2021-05-25
Add support for opt-svfg in vfspta

mbarbar:
As title.

The heavy lifting is done in `prelabel` where no longer do we look at the outgoing edges of delta nodes, but the outgoing edges they _could_ have (according to the pre-analysis).

Other additions helpful for dealing with this stuff:
* Added `dumpLocVersionMaps`.
* Delta node testing is the SVFG's problem now.
  * Others may find this useful since it answers "can this node change during on-the-fly callgraph resolution?".
  * The method requires an over-approximated callgraph passed in as an argument.
* Some assertions/conditions changed in `updateConnectedNodes`.

Like before, with the server still down, testing on things like `bash`, `vis`, `ninja`, etc., nothing too large. `bash` had its time halved (25s, Andersen's take 12s), and the others are too small to see other than a reliable 1s or so improvement.

### DIFF
--- a/include/Graphs/SVFG.h
+++ b/include/Graphs/SVFG.h
@@ -250,6 +250,11 @@ public:
         return nodeNum;
     }
 
+    /// Returns true if l is a delta node, i.e., may change due to
+    /// on-the-fly call graph resolution. approxCallGraph is the over-approximate
+    /// call graph built by the pre-analysis.
+    virtual bool isDeltaNode(NodeID l, PTACallGraph *approxCallGraph) const;
+
     inline const DummyVersionPropSVFGNode *addDummyVersionPropSVFGNode(const NodeID object, const NodeID version)
     {
         DummyVersionPropSVFGNode *dvpNode = new DummyVersionPropSVFGNode(totalVFGNode++, object, version);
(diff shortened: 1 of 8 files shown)


include/Graphs/SVFG.h
```diff
@@ -250,6 +250,11 @@ class SVFG : public VFG
         return nodeNum;
     }
 
+    /// Returns true if l is a delta node, i.e., may change due to
+    /// on-the-fly call graph resolution. approxCallGraph is the over-approximate
+    /// call graph built by the pre-analysis.
+    virtual bool isDeltaNode(NodeID l, PTACallGraph *approxCallGraph) const;
```
yuleisui, line 256: What is a DeltaNode? Is that the dummySVFGNode in the last patch?

mbarbar: Any node which may have new incoming edges due to on-the-fly callgraph resolution. So things like indirect call targets.

yuleisui: Can we use a better name? Will this method be called by SFS too?

mbarbar: Yes sure, I need ideas though :) (this is the name used in the paper by the way).

SFS has no need for this. Currently only VSFS uses it (it used to be in `VersionedFlowSensitive`).

yuleisui: I see. Can we put this method in `VersionedFlowSensitive.cpp` then?

mbarbar: Done.

Issue #444, opened 2021-05-25
ContextDDA::isHeapCondMemObj not checking for NULL mem->getRefVal()


pietroborrello:
I have an abort triggered in 
```
/usr/lib/llvm-10/include/llvm/Support/Casting.h:105: 
static bool llvm::isa_impl_cl<llvm::Instruction, const llvm::Value *>::doit(const From *) [To = llvm::Instruction, From = const llvm::Value *]: 
Assertion `Val && "isa<> used on a null pointer"' failed.
```

Due to ContextDDA passing a NULL reference to `SVFUtil::dyn_cast<Instruction>()` in [ContextDDA::isHeapCondMemObj](https://github.com/SVF-tools/SVF/blob/2ecd7e23b19941dcb52929fa29cbe97a6a1a3eed/lib/DDA/ContextDDA.cpp#L318).
This is basically due to a MemObj being flagged as an HeapObj but having a NULL `refVal`.

I'm not sure whether having such an object is a different bug in itself, or if a HeapObj with a NULL `refVal` is valid, however I see   [MemObj::getType()](https://github.com/SVF-tools/SVF/blob/2ecd7e23b19941dcb52929fa29cbe97a6a1a3eed/lib/MemoryModel/MemModel.cpp#L287) checking for a NULL value in `refVal` before doing the same check, so probably it is just a missing check, and such an object is valid.

### details
I found this assert triggering while doing ContextDDA queries on [libxml2](https://gitlab.gnome.org/GNOME/libxml2.git), which is a pretty big project, it triggers after about 5h of analysis.

### version
llvm: 10
svf: 2.0

yuleisui:
It is strange. If a mem object is flagged as a heap. It must be an instruction.

https://github.com/SVF-tools/SVF/blob/45460c79a7e8843f54bc0432e0f5bd6e5758e00e/lib/MemoryModel/MemModel.cpp#L146

Could you help check where the heap object is created other than here in your SVF version?

pietroborrello:
Examining the aborting execution that I have in gdb, it seems that the MemObj is actually a `DummyObjNode` based on its `nodeKind`. Let me know any useful information that I could provide.
It seems also `MemObj::init` sets the HEAP_OBJ flag, both in my version and in master branch.

https://github.com/SVF-tools/SVF/blob/2ecd7e23b19941dcb52929fa29cbe97a6a1a3eed/lib/MemoryModel/MemModel.cpp#L243

[...]

PR #445, opened 2021-05-25
contextDDA: add check on NULL refVal in isHeapCondMemObj

pietroborrello:
issue: #444 

PS: I'm not sure if the check on `DummyValNode` is correct

### DIFF
--- a/lib/DDA/ContextDDA.cpp
+++ b/lib/DDA/ContextDDA.cpp
@@ -315,7 +315,15 @@ bool ContextDDA::isHeapCondMemObj(const CxtVar& var, const StoreSVFGNode*)
     assert(mem && "memory object is null??");
     if(mem->isHeap())
     {
-        if(const Instruction* mallocSite = SVFUtil::dyn_cast<Instruction>(mem->getRefVal()))
+        if (!mem->getRefVal()) {
+            PAGNode *pnode = _pag->getPAGNode(getPtrNodeID(var));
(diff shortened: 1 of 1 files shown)


lib/DDA/ContextDDA.cpp
```diff
@@ -315,7 +315,15 @@ bool ContextDDA::isHeapCondMemObj(const CxtVar& var, const StoreSVFGNode*)
     assert(mem && "memory object is null??");
     if(mem->isHeap())
     {
-        if(const Instruction* mallocSite = SVFUtil::dyn_cast<Instruction>(mem->getRefVal()))
+        if (!mem->getRefVal()) {
+            PAGNode *pnode = _pag->getPAGNode(getPtrNodeID(var));
+            assert((pnode->getNodeKind() == PAGNode::DummyObjNode || 
+                pnode->getNodeKind() == PAGNode::DummyValNode ||
+                pnode->getNodeKind() == PAGNode::CloneDummyObjNode) &&
+                "emtpy refVal in non-dummy object");
```
yuleisui, line 323: Please use "SVFUtil::isa<>()" to simplify the code

lib/DDA/ContextDDA.cpp
```diff
@@ -315,7 +315,15 @@ bool ContextDDA::isHeapCondMemObj(const CxtVar& var, const StoreSVFGNode*)
...
+            return false;
```
yuleisui, line 324: This should return true since we assume dummy each heap object is not a singleton.

PR #446, opened 2021-05-26
Fix some issues on andersen analysis result recovering

sherlly:
1. IO handle problem on reading big file
2. Insufficient information recorded when storing the analysis result
3. Insensitive field node need to remove to keep the analysis result consistent

### DIFF
--- a/lib/WPA/Andersen.cpp
+++ b/lib/WPA/Andersen.cpp
@@ -505,7 +505,7 @@ bool Andersen::collapseField(NodeID nodeId)
         if (fieldId != baseId)
         {
             // use the reverse pts of this field node to find all pointers point to it
-            const NodeBS &revPts = getRevPts(fieldId);
+            const NodeBS revPts = getRevPts(fieldId);
             for (const NodeID o : revPts)
             {
                 // change the points-to target from field to base node
--- a/include/MemoryModel/MutablePointsToDS.h
+++ b/include/MemoryModel/MutablePointsToDS.h
@@ -11,22 +11,6 @@
 namespace SVF
 {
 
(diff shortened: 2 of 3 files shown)


lib/WPA/Andersen.cpp
```diff
@@ -95,8 +95,10 @@ void AndersenBase::analyze()
     initialize();
 
     bool readResultsFromFile = false;
-    if(!Options::ReadAnder.empty())
+    if(!Options::ReadAnder.empty()) {
         readResultsFromFile = this->readFromFile(Options::ReadAnder);
+        PointerAnalysis::finalize();
```
yuleisui, line 100: can we use finalize() not the PointerAnalysis one here?

include/MemoryModel/PointerAnalysisImpl.h
```diff
@@ -198,6 +199,24 @@ class BVDataPTAImpl : public PointerAnalysis
...
+
+            if(memObj && memObj->isFieldInsensitive()){
+                for(NodeID id : t.second) {
+                    if (GepObjPN *gepNode = SVFUtil::dyn_cast<GepObjPN>(pag->getPAGNode(id))) {
+                        const LocationSet ls = gepNode->getLocationSet();
+                        GepObjNodeMap.erase(std::make_pair(base, ls));
+                        pag->removeGNode(gepNode);
+                    } else
+                        assert(id == base && "Not a GepObj Node!");
+                }
```
yuleisui, line 216: why not put these lines into the if branch in line 196?

lib/WPA/Andersen.cpp
yuleisui: OK

PR #447, opened 2021-05-27
Account for reverse points-to sets when removing objects

mbarbar:
As discussed, this updates reverse points-to sets when calling functions like `clearPts`. It appears to me that changes only need to be done to the base mutable `PTData`, because no other `PTData` handles reverse points-to sets on their own.

I also moved the `insertKey` function to `SVFUtil` (makes more sense to belong there).

I've moved the test for `this->rev` to the caller, not the callee. In theory, this should have a performance impact for long running analyses since the arguments don't need to be processed (though inline, the arguments need to be processed before the test because some have side effects, though I haven't checked the assembly).

In Andersen's, I made the caller to `getRevPts` not take a reference because that looped-over reverse points-to set has iterators invalidated due to the calls to `clearPts` in the loop. There's obviously a small performance hit there. I do feel it is a bit unsafe though, if someone was to change it back to a reference without realising. Any suggestions?

### DIFF
--- a/include/MemoryModel/MutablePointsToDS.h
+++ b/include/MemoryModel/MutablePointsToDS.h
@@ -59,19 +59,19 @@ public:
 
     virtual inline bool addPts(const Key &dstKey, const Data& element) override
     {
-        addSingleRevPts(revPtsMap[element], dstKey);
+        if (this->rev) addSingleRevPts(revPtsMap[element], dstKey);
         return addPts(ptsMap[dstKey], element);
     }
 
     virtual inline bool unionPts(const Key& dstKey, const Key& srcKey) override
     {
-        addRevPts(ptsMap[srcKey], dstKey);
+        if (this->rev) addRevPts(ptsMap[srcKey], dstKey);
         return unionPts(ptsMap[dstKey], getPts(srcKey));
     }
 
     virtual inline bool unionPts(const Key& dstKey, const DataSet& srcDataSet) override
     {
-        addRevPts(srcDataSet,dstKey);
+        if (this->rev) addRevPts(srcDataSet,dstKey);
         return unionPts(ptsMap[dstKey], srcDataSet);
     }
 
@@ -82,14 +82,14 @@ public:
 
     virtual void clearPts(const Key& var, const Data& element) override
     {
-        clearSingleRevPts(revPtsMap[element], var);
+        if (this->rev) clearSingleRevPts(revPtsMap[element], var);
         ptsMap[var].reset(element);
     }
 
     virtual void clearFullPts(const Key& var) override
     {
         DataSet &pts = ptsMap[var];
-        clearRevPts(pts, var);
+        if (this->rev) clearRevPts(pts, var);
         pts.clear();
     }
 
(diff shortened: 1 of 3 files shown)


include/MemoryModel/MutablePointsToDS.h
```diff
@@ -75,19 +59,19 @@ class MutablePTData : public PTData<Key, KeySet, Data, DataSet>
 
     virtual inline bool addPts(const Key &dstKey, const Data& element) override
     {
-        addSingleRevPts(revPtsMap[element], dstKey);
+        if (this->rev) addSingleRevPts(revPtsMap[element], dstKey);
```
yuleisui, line 62: can we put this `if` branch inside `addSingleRevPts ` or `addRevPts `

lib/WPA/Andersen.cpp
```diff
@@ -505,7 +505,7 @@ bool Andersen::collapseField(NodeID nodeId)
         if (fieldId != baseId)
         {
             // use the reverse pts of this field node to find all pointers point to it
-            const NodeBS &revPts = getRevPts(fieldId);
+            const NodeBS revPts = getRevPts(fieldId);
```
yuleisui, line 508: Why remove the reference?

yuleisui:
> As discussed, this updates reverse points-to sets when calling functions like `clearPts`. It appears to me that changes only need to be done to the base mutable `PTData`, because no other `PTData` handles reverse points-to sets on their own.
> 
> I also moved the `insertKey` function to `SVFUtil` (makes more sense to belong there).
> 
> I've moved the test for `this->rev` to the caller, not the callee. In theory, this should have a performance impact for long running analyses since the arguments don't need to be processed (though inline, the arguments need to be processed before the test because some have side effects, though I haven't checked the assembly).
> 
Not quite sure whether the performance impact would be large or not. It seems to me that `this->rev` is more readable in callees.

> In Andersen's, I made the caller to `getRevPts` not take a reference because that looped-over reverse points-to set has iterators invalidated due to the calls to `clearPts` in the loop. There's obviously a small performance hit there. I do feel it is a bit unsafe though, if someone was to change it back to a reference without realising. Any suggestions?

I think it is fine and the performance should be ok.

include/MemoryModel/MutablePointsToDS.h
mbarbar: Done.

lib/WPA/Andersen.cpp
mbarbar: The `clearPts` call in the loop (over `revPts`) will invalidate the current iterator into `revPts`, I believe, since `clearPts` will call `clearSingleRevPts`.

yuleisui:
> As discussed, this updates reverse points-to sets when calling functions like `clearPts`. It appears to me that changes only need to be done to the base mutable `PTData`, because no other `PTData` handles reverse points-to sets on their own.
>
Just a note that for any future `PTData` other than mutable `PTData`, we still need to implement `getRevPts` and also clear RevPts once we call `clearPts`. This because Andersen's analysis will use `getRevPts` here: https://github.com/SVF-tools/SVF/blob/master/lib/WPA/Andersen.cpp#L508

> I also moved the `insertKey` function to `SVFUtil` (makes more sense to belong there).
> 
> I've moved the test for `this->rev` to the caller, not the callee. In theory, this should have a performance impact for long running analyses since the arguments don't need to be processed (though inline, the arguments need to be processed before the test because some have side effects, though I haven't checked the assembly).
> 
> In Andersen's, I made the caller to `getRevPts` not take a reference because that looped-over reverse points-to set has iterators invalidated due to the calls to `clearPts` in the loop. There's obviously a small performance hit there. I do feel it is a bit unsafe though, if someone was to change it back to a reference without realising. Any suggestions?

mbarbar:
> Just a note that for any future PTData other than mutable PTData, we still need to implement getRevPts and also clear RevPts once we call clearPts. This because Andersen's analysis will use getRevPts here: https://github.com/SVF-tools/SVF/blob/master/lib/WPA/Andersen.cpp#L508

Noted.

PR #452, opened 2021-05-28
Document debug node allocation as default

mbarbar:
This fell through the cracks when the default returned to debug (thanks @pietroborrello for pointing out).

### DIFF
--- a/lib/Util/Options.cpp
+++ b/lib/Util/Options.cpp
@@ -15,9 +15,9 @@ namespace SVF
         llvm::cl::init(NodeIDAllocator::Strategy::DEBUG),
         llvm::cl::desc("Method of allocating (LLVM) values and memory objects as node IDs"),
         llvm::cl::values(
-            clEnumValN(NodeIDAllocator::Strategy::DENSE, "dense", "allocate objects together and values together, separately (default)"),
+            clEnumValN(NodeIDAllocator::Strategy::DENSE, "dense", "allocate objects together and values together, separately"),
             clEnumValN(NodeIDAllocator::Strategy::SEQ, "seq", "allocate values and objects sequentially, intermixed"),
-            clEnumValN(NodeIDAllocator::Strategy::DEBUG, "debug", "allocate value and objects sequentially, intermixed, except GEP objects as offsets")));
+            clEnumValN(NodeIDAllocator::Strategy::DEBUG, "debug", "allocate value and objects sequentially, intermixed, except GEP objects as offsets (default)")));
 
     const llvm::cl::opt<unsigned> Options::MaxFieldLimit(
         "field-limit",


pietroborrello:
Should DEBUG really be the default? Given the (maybe remote) possibility of dummy objects' ID clashing with GEP nodes?

mbarbar:
The reason why we returned initially was because `dense` (and to a lesser extent, `seq`) have ugly numbers for IDs. In dense for example, all the values are allocated within the range `UINT_MAX - N` for some `N` so that the objects can be `0` to some `M`. I think, moving forward, fixing `debug` to never clash would be necessary whether it's kept default or not.

yuleisui:
After a second thought.

I think the DEBUG option is correct @pietroborrello. @mbarbar, for you following case:
```
Assuming we fix the assertion, the value we test against is a moving target. Consider:

1. Initial PAG construction results in 99 symbols.
2. The GEP multiplier is thus 100.
3. The analysis creates a GEP object at 105 (passing the assertion).
4. Andersen's (or another analysis) adds a dummy object.
5. The GEP multiplier is thus now 1000.
6. Though 105 already exists as a GEP object, and when the analysis tries to add a bunch of dummy objects... 105 will be "overwritten".
```

Step 3 will be captured as invalid by the assertion, and will never pass.

https://github.com/SVF-tools/SVF/blob/67b4bb703bfc31083cad4bd827ea5e151f10b131/lib/Util/NodeIDAllocator.cpp#L96

The assertion is defensive and conservative but correct.

yuleisui:
> > Step 3 will be captured as invalid by the assertion, and will never pass.
> 
> I'm not sure I understood, how would step 3 fail? At the time of step 3, `numSymbols` would be 99, so the assertion will pass, as in the list, right? Probably I did not get the `Assuming we fix the assertion`.
> 
`numSymbols` is 99 and once you want to create a gepobj with id 105, it will trigger an assertion.

> I was thinking about having a 64bit node ID. This would allow to reserve 32bits for the offset, and 32bits for the obj id (or just the nearest power of 10), and allowing just to assert that the object id will never go past the static maximum id. This paying the cost of having slightly uglier debug GEP ids with a bunch of zeros.

mbarbar:
@yuleisui the assertion shouldn't fail because `id` is larger (105).

@pietroborrello I like this idea with 64-bit node IDs as it would save us from overflow, it appears. I initially thought this will almost double memory usage, but since we are using bit-vectors that's not the case.

Unfortunately, LLVM's `SparseBitVector` which is in use uses `unsigned` which are usually 32 bits. In fact they appear to mix unsigned and signed integers in the interface, making some methods signed overflow (undefined in C, probably C++ too?) when large numbers >2-ish billion are in play (see https://llvm.org/doxygen/SparseBitVector_8h_source.html#l00779 for example).

It doesn't appear hard to modify it, but obviously that's undesirable. Maintaining a custom/forked sparse bit-vector would also save (a lot of, I suspect) memory because LLVM's sparse bit-vector uses doubly linked lists but SVF only needs ones with singly linked lists.

PR #453, opened 2021-05-29
Add assertions when creating a clashed node (guard the DEBUG strategy in NodeIDAllocator)

yuleisui:
@mbarbar @pietroborrello 

This is a quick solution to make sure that if we creating clashed nodes in the DEBUG strategy, the program will crash.

Please let me know any comments or suggestions.

### DIFF
--- a/include/Graphs/PAG.h
+++ b/include/Graphs/PAG.h
@@ -771,15 +771,13 @@ public:
     /// Add a value (pointer) node
     inline NodeID addValNode(const Value*, PAGNode *node, NodeID i)
     {
-        assert(hasGNode(i) == false && "This NodeID has been used and clashing here, Please check NodeIDAllocator");
-        assert(i < UINT_MAX && "exceeding the maximum node limits");
+        assert(hasGNode(i) == false && "This NodeID clashes here. Please check NodeIDAllocator. Switch Strategy::DEBUG to SEQ or DENSE");
         return addNode(node,i);
     }
     /// Add a memory obj node
     inline NodeID addObjNode(const Value*, PAGNode *node, NodeID i)
     {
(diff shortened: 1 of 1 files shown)


mbarbar:
The first asserts in the pair look good to me, as long as these are the only places in the PAG where nodes are created.

The `i < UINT_MAX` should always succeed regardless of overflow because `i` is a 32-bit unsigned integer so it cannot be more than `UINT_MAX`.

To check for overflow, maybe the formula in the allocator can give a 64-bit number, asserted against `(uint64_T)UINT_MAX`, and then cast back to `NodeID` to return?

yuleisui:
> The first asserts in the pair look good to me, as long as these are the only places in the PAG where nodes are created.
> 

Yes, they are.

> The `i < UINT_MAX` should always succeed regardless of overflow because `i` is a 32-bit unsigned integer so it cannot be more than `UINT_MAX`.
> 
> To check for overflow, maybe the formula in the allocator can give a 64-bit number, asserted against `(uint64_T)UINT_MAX`, and then cast back to `NodeID` to return?

Do you want me to remove the second assertion or you want to put one in allocator? Maybe you can write down the exact assertion?

mbarbar:
>Do you want me to remove the second assertion

I think so, it should be never fail.

> or you want to put one in allocator? Maybe you can write down the exact assertion?

If in the allocator we did...
```
uint64_t id64 = *FORMULA*;
assert(id64 < UINT_MAX && "... overflow ...");  // UINT_MAX should be promoted to 64-bit by the compiler I think.
id = (NodeID)id64;
```

I don't think you can do this outside. I'm not sure how overflow checking could be done given _just_ the result (which would have already overflowed). If there is an overflow though and the allocator tries to reuse that result, the first GNode asserts should trigger, probably.

yuleisui:
Just removed the second assertion and keep only the first one. #452 #451

PR #468, opened 2021-06-10
Cache consume and yield maps in VFS

mbarbar:
In many instances we were doing...
```
    if hasYield(l, o):
        y = yield[l][o];  // or sometimes y = yieldl[o]
```
resulting in two map lookups (modulo nesting): one for `hasYield` and one to actually get the value. An alternative would be to use `find`, but it's not pleasant having all these extra one-use iterators around. This patch defines `getYield` and `getConsume` to instead do...
```
    y = getYield(l, o);  // Internally, this is a map lookup.
    if y != invalidVersion:
        use y;
```
thus resulting in a map lookup  (wrapper on `find`) and a comparison.

`get{Yield,Consume}` also cache the first layer of lookup, i.e. `yield[l]` or `consume[l]`, allowing the user to not have to worry about this (we often would do something like `&consumel = consume[l]` then perform `consumel[o]` lookups in a loop}.

It seems overall there is a tiny but consistent speedup on my machine.

I've also included in this patch some style changes and extra `const`s, hopefully not too distracting. The meat is in `set{Version,Consume,Yield}`, `get{Version,Consume,Yield}`, and `{consume,yield}Cache`.

### DIFF
--- a/include/WPA/VersionedFlowSensitive.h
+++ b/include/WPA/VersionedFlowSensitive.h
@@ -42,10 +42,16 @@ public:
     /// (o -> (v -> versions with rely on o:v).
     typedef Map<NodeID, Map<Version, Set<Version>>> VersionRelianceMap;
 
-    enum VersionType {
-        CONSUME,
-        YIELD,
-    };
+    /// For caching the first step in LocVersionMaps.
+    typedef struct VersionCache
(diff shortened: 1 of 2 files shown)


yuleisui:
Looks good to me.

PR #478, opened 2021-07-05
Add hash-consed points-to sets

mbarbar:
As we discussed, this adds hash-consed points-to sets.

Options: `-ptd=persistent` for hash-consing and `-ptd=mutable` for what we currently have (default). The bulk of this is in `PersistentPointsToCache` and `PersistentPointsToDS`.

I've renamed `IncMutable` -> `MutableInc` so that the type is the prefix.

I've added a couple of options `-fs-time-limit` and `-ander-time-limit` which abort the analysis if they go over the specified time. `-ander-time-limit` does nothing if `-fs-time-limit` is set since `-fs-time-limit` includes everything even auxiliary analysis (contrary to what I mentioned in the email). We can probably tweak exactly what is and isn't included. I've also hardcoded some exit codes for scripts to detect, but they're magic numbers. Any suggestions?

I've also specialised the hash function for `NodeID` pairs.

I think that's the jist of it.


lib/WPA/Andersen.cpp
```diff
@@ -106,6 +129,15 @@ void AndersenBase::analyze()
...
 
+        bool ourAlarm = !callersAlarm();
+        // If it's not our alarm, it's probably FS's alarm and we don't want to mess with that.
+        // TODO: maybe make AnderTimeLimit and FsTimeLimit work together?
+        if (ourAlarm && Options::AnderTimeLimit != 0)
+        {
+            signal(SIGALRM, &timeLimitReached);
+            alarm(Options::AnderTimeLimit);
+        }
+
```
yuleisui, line 140: The code here is a bit more inside the solver and can we make this as a method and return a bool value or some other ways to shorten and hidden this?  It would also be good to explain `ourAlarm` 's meaning.

mbarbar: Good point, I've abstracted it all into a couple of methods.

mbarbar:
Mm, yes. I've moved them. Eventually we can probably add "time-up" messages.

yuleisui:
Looks good!

PR #480, opened 2021-07-08
Include JSON headers only where required

adrianherrera:
This can cause namespace collisions when trying to include SVF as an external project.


yuleisui:
Thanks!

PR #486, opened 2021-07-15
SABER: add default implementations for some virtual funcs

adrianherrera:
Hey SVF devs!

Some small mods to the `SrcSnkDDA` class to provide default implementations for some of the virtual methods.

I am trying to extend `SrcSnkDDA` to build my own taint analyzer, and it seems that I have to provide implementations for methods that I don't particularly need. For example, I can implement `initSrcs` and `initSnks` without ever needing to call `isSourceLinkFun` (similarly, `isSinkLikeFun`). Likewise, `isSource` and `isSink` will always just check that the given node is contained within the set of source/sink nodes.

What do you think?

### DIFF
--- a/include/SABER/LeakChecker.h
+++ b/include/SABER/LeakChecker.h
@@ -75,34 +75,24 @@ public:
     /// Initialize sources and sinks
     //@{
     /// Initialize sources and sinks
-    virtual void initSrcs();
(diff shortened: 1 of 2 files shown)


yuleisui:
Look ok to me, will merge after the CI.

Issue #485, opened 2021-07-15
Go from SVFG to LLVM values and back


gerion0:
I have searched for a function to get a SVFG node from an arbitrary LLVM value and go this way back.

However, this seems more difficult than expected. There exist a few methods for specific types of nodes but they often result in an assertion and so prevent further processing.

Here is a [minimal example/testcase for demonstration](https://gitlab.com/geri0n/svf-test):
```cpp
#include <llvm/Support/CommandLine.h>
#include <vector>
#include <string>
#include <SVF-FE/PAGBuilder.h>
#include <WPA/Andersen.h>
#include <llvm/IR/ValueSymbolTable.h>

static llvm::cl::opt<std::string> InputFilename(llvm::cl::Positional,
        llvm::cl::desc("<input bitcode>"), llvm::cl::init("-"));

const SVF::VFGNode* get_vfg_node(const SVF::SVFG& vfg, const llvm::Value* value) {
	SVF::PAG* pag = SVF::PAG::getPAG();
	SVF::PAGNode* pNode = pag->getPAGNode(pag->getValueNode(value));
	assert(pNode != nullptr);
	const SVF::VFGNode* vNode = vfg.getDefSVFGNode(pNode);
	assert(vNode != nullptr);
	return vNode;
}

int main(int argc, char ** argv) {
	int arg_num = 0;
	char **arg_value = new char*[argc];
	std::vector<std::string> moduleNameVec;
	SVF::SVFUtil::processArguments(argc, argv, arg_num, arg_value, moduleNameVec);
	llvm::cl::ParseCommandLineOptions(arg_num, arg_value,
                                "Whole Program Points-to Analysis\n");

	SVF::SVFModule* svfModule = SVF::LLVMModuleSet::getLLVMModuleSet()->buildSVFModule(moduleNameVec);
	llvm::Module* llvmModule = SVF::LLVMModuleSet::getLLVMModuleSet()->getMainLLVMModule();

	SVF::PAGBuilder builder;
	SVF::PAG *pag = builder.build(svfModule);

	SVF::Andersen *ander = SVF::AndersenWaveDiff::createAndersenWaveDiff(pag);

	SVF::SVFGBuilder svfBuilder;
	SVF::SVFG& svfg = *svfBuilder.buildFullSVFG(ander);
	svfg.dump("svfg");

	for (const auto& [_, node] : svfg) {
		llvm::errs() << "Node " << *node << "\n";
[...]

PR #492, opened 2021-07-28
Add functions for mapping from SVFG nodes to LLVM values and back

gerion0:
See #485.

I have tested the code with https://gitlab.com/geri0n/svf-test/-/tree/llvmtovfg.

The back map seems not to work with these nodes (since the PAGNodes are not linked to llvm::Values):
- ActualParmSVFGNode
- FormalRetVFGNode
- IntraPHIVFGNode

Moreover an `assert` in the map filling code does not work: `pag->getPAGEdgeNum()` give me way more edges, than I can find in the graph.

### DIFF
--- a/include/Graphs/PAG.h
+++ b/include/Graphs/PAG.h
@@ -70,6 +70,7 @@ public:
     typedef Map<const SVFFunction*,PAGEdgeSet> FunToPAGEdgeSetMap;
     typedef Map<const ICFGNode*,PAGEdgeList> Inst2PAGEdgesMap;
     typedef Map<NodeID, NodeID> NodeToNodeMap;
+    typedef Map<const Value*,PAGEdgeSet> ValueToEdgeMap;
     typedef std::pair<NodeID, Size_t> NodeOffset;
     typedef std::pair<NodeID, LocationSet> NodeLocationSet;
     typedef Map<NodeOffset,NodeID> NodeOffsetMap;
@@ -97,6 +98,7 @@ private:
     CSToArgsListMap callSiteArgsListMap;	///< Map a callsite to a list of all its actual parameters
     CSToRetMap callSiteRetMap;	///< Map a callsite to its callsite returns PAGNodes
     FunToRetMap funRetMap;	///< Map a function to its unique function return PAGNodes
+    ValueToEdgeMap valueToEdgeMap;	///< Map llvm::Values to all corresponding PAGEdges
     static PAG* pag;	///< Singleton pattern here to enable instance of PAG can only be created once.
     CallSiteToFunPtrMap indCallSiteToFunPtrMap; ///< Map an indirect callsite to its function pointer
     FunPtrToCallSitesMap funPtrToCallSitesMap;	///< Map a function pointer to the callsites where it is used
@@ -561,6 +563,25 @@ public:
     {
         return mem->getSymId();
     }
+    /// Get all PAG Edges that corresponds to an LLVM value
+    inline const PAGEdgeSet& getValueEdges(const Value *V)
+    {
+        auto it = valueToEdgeMap.find(V);
+        if (it == valueToEdgeMap.end()) {
+            //special empty set
+            return valueToEdgeMap.at(nullptr);
+        }
+        return it->second;
+    }
+
+    inline void mapValueToEdge(const Value *V, PAGEdge *edge)
+    {
+        auto inserted = valueToEdgeMap.emplace(V, PAGEdgeSet{edge});
+        if (!inserted.second) {
+            inserted.first->second.emplace(edge);
+        }
+    }
+
     /// Get memory object - Return memory object according to pag node id
     /// return whole allocated memory object if this node is a gep obj node
     /// return nullptr is this node is not a ObjPN type
--- a/include/Graphs/SVFG.h
+++ b/include/Graphs/SVFG.h
@@ -173,6 +173,11 @@ public:
         return getSVFGNode(getDef(pagNode));
     }
 
+    /// Return the corresponding SVFGNodes to a given llvm::Value.
+    /// return an empty list, if the no mapping is possible
+    std::set<const SVFGNode*> fromLLVMValue(const llvm::Value* value) const;
+
+
     /// Perform statistics
     void performStat();
 
--- a/include/Graphs/VFGNode.h
+++ b/include/Graphs/VFGNode.h
@@ -182,7 +182,7 @@ public:
     }
     //@}
 
-    virtual const std::string toString() const;
+    const std::string toString() const override;
 };
 
 /*!
@@ -221,7 +221,7 @@ public:
     }
     //@}
 
-    virtual const std::string toString() const;
+    const std::string toString() const override;
 };
 
 /*!
@@ -260,7 +260,7 @@ public:
     }
     //@}
 
-    virtual const std::string toString() const;
+    const std::string toString() const override;
 };
 
 /*!
@@ -299,7 +299,7 @@ public:
     }
     //@}
 
-    virtual const std::string toString() const;
+    const std::string toString() const override;
 };
 
 
@@ -371,7 +371,7 @@ public:
         return opVers.end();
     }
     //@}
-    virtual const std::string toString() const;
+    const std::string toString() const override;
 };
 
 
@@ -443,7 +443,7 @@ public:
     }
     //@}
 
-    virtual const std::string toString() const;
+    const std::string toString() const override;
 };
 
 /*!
@@ -555,7 +555,7 @@ public:
     }
     //@}
 
-    virtual const std::string toString() const;
+    const std::string toString() const override;
 };
 
 /*
@@ -626,7 +626,7 @@ public:
     }
     //@}
 
-    virtual const std::string toString() const;
+    const std::string toString() const override;
 };
 
 
@@ -679,7 +679,7 @@ public:
     }
     //@}
 
-    virtual const std::string toString() const;
+    const std::string toString() const override;
 };
 
 
@@ -716,7 +716,7 @@ public:
     }
     //@}
 
-    virtual const std::string toString() const;
+    const std::string toString() const override;
 };
 
 
@@ -760,7 +760,7 @@ public:
     }
     //@}
 
-    virtual const std::string toString() const;
+    const std::string toString() const override;
 };
 
 /*
@@ -809,7 +809,7 @@ public:
     }
     //@}
 
-    virtual const std::string toString() const;
+    const std::string toString() const override;
 };
 
 
@@ -877,7 +877,7 @@ public:
     }
     //@}
 
-    virtual const std::string toString() const;
+    const std::string toString() const override;
 };
 
 /*!
@@ -933,7 +933,7 @@ public:
     }
     //@}
 
-    virtual const std::string toString() const;
+    const std::string toString() const override;
 };
 
 /*!
@@ -997,7 +997,7 @@ public:
     }
     //@}
 
-    virtual const std::string toString() const;
+    const std::string toString() const override;
 };
 
 /*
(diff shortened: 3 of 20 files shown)


gerion0:
How can I reproduce the tests locally? I have tried:
```
$ cd Test-Suite; ./generate_bc.sh
$ cd build; ctest -VV
```
This results in:
```
100% tests passed, 0 tests failed out of 114
```
However, Github reports 405 tests.

yuleisui:
Unfortunately, the test suite failed.  It seems to me that it was because `wpa` was not generated sucessfully.

Could you break down this patch into two?
(1) minor changes for code cleaning/formating or some helpful data structures will be used later (in order to pass the CI)
(2) code only related to SVFG.h/.cpp for your mapping

gerion0:
I will do a binary search via forced push. Please do not merge.

Howe can I reproduce the tests locally?

yuleisui:
> I will do a binary search via forced push. Please do not merge.
> 
> Howe can I reproduce the tests locally?

You can take a look at this github workflow file. It basically tells the configuration for testing locally.
https://github.com/SVF-tools/SVF/blob/master/.github/workflows/github-action.yml

include/Graphs/PAG.h
```diff
@@ -561,6 +563,25 @@ class PAG : public GenericGraph<PAGNode,PAGEdge>
     {
         return mem->getSymId();
     }
+    /// Get PAG Edge according to LLVM value
+    /// return nullptr if an edge cannot be found
+    inline PAGEdgeSet getValueEdge(const Value *V)
```
yuleisui, line 568: This should return a const reference?

include/Graphs/VFGNode.h
```diff
@@ -88,6 +88,12 @@ class VFGNode : public GenericVFGNodeTy
         return icfgNode->getFun();
     }
 
+    /// Return the corresponding LLVM value, if possible, nullptr otherwise.
+    virtual const llvm::Value* getLLVMValue() const
```
yuleisui, line 92: We can remove the llvm:: name space here. Also, we will rewrite LLVM Value later. Could just change ``getLLVMValue'' to ``getValue``?

include/Graphs/VFGNode.h
```diff
@@ -182,7 +188,8 @@ class StmtVFGNode : public VFGNode
     }
     //@}
 
-    virtual const std::string toString() const;
+    const llvm::Value* getLLVMValue() const override;
```
yuleisui, line 191: remove llvm namespace here.

include/Graphs/VFGNode.h
```diff
@@ -371,7 +378,8 @@ class CmpVFGNode: public VFGNode
         return opVers.end();
     }
     //@}
-    virtual const std::string toString() const;
+    const llvm::Value* getLLVMValue() const override;
```
yuleisui, line 381: remove llvm namespace and please do so for all other cases.

lib/Graphs/SVFG.cpp
```diff
@@ -524,6 +544,25 @@ void SVFG::dump(const std::string& file, bool simple)
...
+        if (it != PAGEdgeToStmtVFGNodeMap.end()) {
+            ret.emplace(it->second);
+        }
+    }
+    // add all PAGNodes
+    PAGNode* pagNode = pag->getPAGNode(pag->getValueNode(value));
+    if(hasDef(pagNode)) {
+        ret.emplace(getDefSVFGNode(pagNode));
+    }
+    return ret;
```
yuleisui, line 563: We do not return the local set. Pass a reference set as a parameter?

lib/SVF-FE/PAGBuilder.cpp
```diff
@@ -116,6 +116,22 @@ PAG* PAGBuilder::build(SVFModule* svfModule)
...
+    unsigned node_count = 0;
+    for (auto& node : *pag) {
+        node_count++;
+        for (auto edge_it = node.second->OutEdgeBegin(); edge_it != node.second->OutEdgeEnd(); ++edge_it) {
+            auto edge = *edge_it;
+            pag->mapValueToEdge(edge->getValue(), edge);
+            edge_count++;
+        }
+    }
+    errs() << "NodeCount" << node_count << " NodeNum: " << pag->getTotalNodeNum() << " " << pag->getPAGNodeNum() << "\n";
```
yuleisui, line 130: delete these output messages.

lib/SVF-FE/PAGBuilder.cpp
```diff
@@ -116,6 +116,22 @@ PAG* PAGBuilder::build(SVFModule* svfModule)
...
+    unsigned edge_count = 0;
+    unsigned node_count = 0;
+    for (auto& node : *pag) {
+        node_count++;
+        for (auto edge_it = node.second->OutEdgeBegin(); edge_it != node.second->OutEdgeEnd(); ++edge_it) {
+            auto edge = *edge_it;
+            pag->mapValueToEdge(edge->getValue(), edge);
+            edge_count++;
+        }
+    }
```
yuleisui, line 129: You can init your map here. It will become simpler.
https://github.com/SVF-tools/SVF/blob/master/lib/SVF-FE/PAGBuilder.cpp#L1437

include/Graphs/PAG.h
gerion0: Given that `PAGEdgeSet` is `Set<const PAGEdge*>` and the Set is moved anyway, I see no advantage of a const reference. However, the comment is wrong.

include/Graphs/VFGNode.h
gerion0: Will do. What do mean with "we will rewrite LLVM Value later"?
Namespace dropping is possible because `llvm::Value` is typedefed to `SVF::Value`, isn't it (generally, dropping namespaces in headers is not desirable)?

lib/Graphs/SVFG.cpp
gerion0: The set is moved in these cases, so there shouldn't be a problem.

lib/SVF-FE/PAGBuilder.cpp
gerion0: I have inserted these messages only because the assertion a few lines later does not work. Can you elaborate on that?
Why is `getTotalEdgeNum()` different to `edge_count`?

lib/SVF-FE/PAGBuilder.cpp
gerion0: Thanks, will take a look into it.

include/Graphs/PAG.h
yuleisui: `Reference` can avoid C++ copy construction.
`const` will make the code more readable that the set will not be modified anymore.

include/Graphs/VFGNode.h
yuleisui: Later, we may use `Value` as an interface, it could be an LLVM Value wrapper and could be something else for other compilers.

lib/Graphs/SVFG.cpp
```diff
@@ -156,6 +156,26 @@ const std::string ThreadMHPIndSVFGEdge::toString() const {
     return rawstr.str();
 }
 
+const llvm::Value* StmtSVFGNode::getLLVMValue() const {
```
yuleisui, line 159: please remove the namespace.

include/Graphs/PAG.h
gerion0: I don't get it. See this minimal example:
```
#include <set>

std::set<int> my_function() {
	std::set<int> my_set;
	for (int i = 0; i < 5; ++i) {
		my_set.emplace(i);
	}
	return my_set;
}

int main() {
	std::set<int> main_set = my_function();
	const std::set<int> onother_set = my_function();
}
```
my_function seems to copy the set. However, the compiler optimizes it into a move. So there is only one move.
This is the resulting IR of my_function (where my_function takes a set pointer):
```
define dso_local void @_Z11my_functionv(%"class.std::set"* noalias sret(%"class.std::set") align 8 %agg.result) #0 personality i8* bitcast (i32 (...)* @__gxx_personality_v0 to i8*) {
```

Returning a reference alone would be wrong since it is a reference to a local variable, which does not exist anymore after the function ends.
Returning a const reference alows the compiler to move instead of returning a real reference, too. However, the caller cannot modify the set anymore. Why should the function force this?

Return just the set allows the caller to make it const or to modify it.

yuleisui: I usually do not trust compiler optimizations:). The optimization usually depends on different compilers and options used. In your case, it is not a local variable but `valueToEdgeMap ` is a class member. I did not see anything wrong that you return a reference.

For `const`, the caller (invoking `getValueEdge `) should only read but not modify the set as your name suggests.

gerion0: Sorry, have confused this function with `fromLLVMValue`. Of course a const reference is meaningful here.
A problem arises, if an empty set should be returned. Therefore, I have put that dummy value at `valueToEdgeMap[nullptr]`.

include/Graphs/SVFG.h
```diff
@@ -173,6 +173,11 @@ class SVFG : public VFG
         return getSVFGNode(getDef(pagNode));
     }
 
+    /// Return the corresponding SVFGNodes to a given llvm::Value.
+    /// return an empty list, if the no mapping is possible
+    std::set<const SVFGNode*> fromLLVMValue(const llvm::Value* value) const;
```
yuleisui, line 178: please also remove the namespace here.

lib/Graphs/SVFG.cpp
yuleisui: If you make this function public, it may be accidentally used by other users. I would suggest making the function like:
 `void SVFG::fromLLVMValue(const llvm::Value* value, std::set<const SVFGNode*>& nodeset) const;'  where `nodeset` is a reference referring to set in the caller, so that we will not return a local set.

include/Graphs/SVFG.h
gerion0: Should I also change the function name to `fromValue` then?

lib/Graphs/SVFG.cpp
gerion0: Here applies what I said in the other comment. Returning a set or specifying a reference parameter leads to the same code.

Theoretical, the local set is returned by a copy. However, in practice it is moved, see https://isocpp.org/blog/2013/02/no-really-moving-a-return-value-is-easy-stackoverflow

include/Graphs/SVFG.h
yuleisui: Yes, please.

lib/SVF-FE/PAGBuilder.cpp
yuleisui: It is strange. Could you have a look at where these two counters `getTotalEdgeNum()` and `edge_count` are increamented?

include/Graphs/SVFG.h
yuleisui: Could you please change this name so that I could merge?

gerion0: Was on vacation. See #497.

gerion0:
Appearantly, the newly added function `fromValue` was removed again in cfb66571e. Can I readd them? Why was it removed?

yuleisui:
Yes, please add it back. The reason was that we may have SVFValue to replace llvm::Value later.

gerion0:
I have edited my post previously. Is the proposed solution more less equivalent?

Especially, the old solution relies on the `PAGEdgeToStmtVFGNodeMap` which seems to be removed in the meantime.

yuleisui:
> I have edited my post previously. Is the proposed solution more less equivalent?
> 

Yes.

> Especially, the old solution relies on the `PAGEdgeToStmtVFGNodeMap` which seems to be removed in the meantime.

PR #494, opened 2021-08-03
Fix GraphTraits for GenericGraph

pzieris:
Hi @yuleisui,

I noticed the inverse GraphTraits for `GenericNode` does not dereference edges to their source nodes, but to their destination nodes.

This patch fixes the bug. It also inlines `edge_unary_function()` into the two `GraphTraits` and renames it to `edge_dest()`, as this name is suggested by the `GraphTraits` template.

For completeness, in a separate commit, I also inlined `pair_unary_function()`. This also cleans up the `llvm` namespace a bit.

### DIFF
--- a/include/Graphs/GenericGraph.h
+++ b/include/Graphs/GenericGraph.h
@@ -449,15 +449,6 @@ public:
 namespace llvm
 {
 
-template<class PairTy, class NodeTy>
-struct pair_unary_function
-{
-    NodeTy operator()(PairTy pair) const {
-      return pair.second;
-    }
-};
-
-
 /*!
  * GraphTraits for nodes
  */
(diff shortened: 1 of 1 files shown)


yuleisui:
Thanks for your fix. I guess the previous one was naming problems but not the correctness?

pzieris:
Commit `13343f9` fixes the edge dereference bug and commit `d11a639` is only cosmetic.

include/Graphs/GenericGraph.h
```diff
@@ -512,10 +506,13 @@ struct GraphTraits<Inverse<SVF::GenericNode<NodeTy,EdgeTy>* > >
     typedef NodeTy NodeType;
     typedef EdgeTy EdgeType;
 
-    typedef edge_unary_function<EdgeType*, NodeType*> DerefEdge;
+    static inline NodeType* edge_dest(const EdgeType* E)
```
yuleisui, line 500: I see. Could we use the name edge_src here?

pzieris: Yes, but I would advise against changing it. I chose `edge_dest()` because it's the suggested name in the [GraphTraits template](https://github.com/llvm/llvm-project/blob/8e29b4b323b87f3855dc71abf1e3f3d48952a4e4/llvm/include/llvm/ADT/GraphTraits.h#L66). Hence, graph algorithms might use `edge_dest()` to dereference an edge returned by `child_edge_begin()` (which is not yet implemented for `GenericNode`, but still).

I think in the context of an inverse traversal of `GenericNode`, the *destination* of an edge is the edge's source node, hence `edge_dest()` still fits.

PR #495, opened 2021-08-07
Speed up VSFS

mbarbar:
When updating the call graph, if we connect a yielded version `y` to a consumed version `c`, we were propagating `y` to **every** version which relies on `y`, rather than just to `c`. This fixes that. For some benchmarks, this makes a big difference. `lynx` for example, on the first call to update the call graph, adds over 200k edges (!), thus performing redundant propagation.

The `Set` -> `vector` change may also make a difference in extremely large benchmarks.


yuleisui:
Looks good to me. Merged.

PR #498, opened 2021-08-16
RFC: CMakeLists: +Wall, +Wextra, +Wpedantic

gerion0:
Obviously, this does not pass the tests because of `-Werror`. However, adding the additional warnings usually leads to cleaner code. Do you think, adding them is a good idea?

I don't have the time and knowledge to fix all the warnings. A workaround of course is to disable `-Werror`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,7 +17,7 @@ if(NOT COMMAND add_llvm_library)
         # add -std=gnu++14
         set(CMAKE_CXX_EXTENSIONS ON)
 
-        set(COMMON_FLAGS "-fPIC -Werror")
+        set(COMMON_FLAGS "-fPIC -Werror -Wall -Wextra -Wpedantic")
 
         if(CMAKE_BUILD_TYPE MATCHES "Debug")
             set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${COMMON_FLAGS} -O0 -fno-rtti")


yuleisui:
I am unsure why do you want to disable -Werror? This option give us better code quality. Any reason behind?

yuleisui:
I see. Might be good to just enable your options for SVF's CPP code base excluding cudd, which might save your time to look into warnings.

gerion0:
Compiler warnings are triggered in the whole codebase not only cudd. They also become more: New pull requests or contributions add new code that contains additional warnings. The above flags would prevent that but to the price of additional work. This PR is meant as a question, whether you are interested in those flags (again to the price of additional work to fix the current warnings).

Why do you want to exclude cudd?

yuleisui:
> Compiler warnings are triggered in the whole codebase not only cudd. They also become more: New pull requests or contributions add new code that contains additional warnings. The above flags would prevent that but to the price of additional work. This PR is meant as a question, whether you are interested in those flags (again to the price of additional work to fix the current warnings).
> 
> Why do you want to exclude cudd?

We will not use cudd later and will replace it with z3. Excluding cudd can also trigger fewer warnings which might make it easier to fix the codebase. It would be great if you could help us with your option for only other folders.

yuleisui:
Just an update on this request. `-Wall` has been added, but the other two are not at the moment since we have  `-Werror` and there might be trivial warnings under different compilers (or different versions) that stop the compilation. 

https://github.com/SVF-tools/SVF/commit/51bc1a0b2193864ae52fbe50da369b6adc7aa246

gerion0:
I would still vote for adding also Wextra (e.g. warn about invalid comparisons) and Wpedandic (warn about strict compliance to ISO C++), since this reduces possible error cases. Actually, testing different compilers with these flags can improve the code even more than just one compiler (since it finds even more problems).
In my opinion a warning free code behaves pretty stable across different compilers and compiler versions with respect to additional warnings.

Anyway, I understand that you close this PR, especially since it doesn't fix the problems as such but only triggers CI errors.

gerion0:
Thank you for adding `-Wall`.

PR #508, opened 2021-09-03
Refactor2

Wren71:
-Remove func parameter in add node methods and add FunEntry and FunExit nodes
-Remove pointsTo parameter in add node methods and get via ver->getMr()->getPointsTo()
-Test suite bcs pass
-Vfs bcs test get stuck on the first bc even with the original SVF project

### DIFF
--- a/include/Graphs/SVFG.h
+++ b/include/Graphs/SVFG.h
@@ -394,43 +394,47 @@ protected:
     }
 
     /// Add memory Function entry chi SVFG node
-    inline void addFormalINSVFGNode(const MemSSA::ENTRYCHI* chi)
+    inline void addFormalINSVFGNode(const SVFFunction* func,  const MRVer* resVer, const PointsTo pointsTo)
     {
-        FormalINSVFGNode* sNode = new FormalINSVFGNode(totalVFGNode++,chi);
-        addSVFGNode(sNode, pag->getICFG()->getFunEntryBlockNode(chi->getFunction()));
-        setDef(chi->getResVer(),sNode);
-        funToFormalINMap[chi->getFunction()].set(sNode->getId());
+        FormalINSVFGNode* sNode = new FormalINSVFGNode(totalVFGNode++, pointsTo, func);
+        addSVFGNode(sNode, pag->getICFG()->getFunEntryBlockNode(func));
+        setDef(resVer,sNode);
+        funToFormalINMap[func].set(sNode->getId());
     }
+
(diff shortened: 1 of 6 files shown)


include/Graphs/SVFG.h
```diff
@@ -394,43 +395,47 @@ class SVFG : public VFG
...
 
     /// Add memory Function entry chi SVFG node
-    inline void addFormalINSVFGNode(const MemSSA::ENTRYCHI* chi)
+    inline void addFormalINSVFGNode(const FunEntryBlockNode* funEntry,  const MRVer* resVer)
     {
-        FormalINSVFGNode* sNode = new FormalINSVFGNode(totalVFGNode++,chi);
-        addSVFGNode(sNode, pag->getICFG()->getFunEntryBlockNode(chi->getFunction()));
-        setDef(chi->getResVer(),sNode);
-        funToFormalINMap[chi->getFunction()].set(sNode->getId());
+        FormalINSVFGNode* sNode = new FormalINSVFGNode(totalVFGNode++, resVer->getMR()->getPointsTo(), funEntry->getFun());
```
yuleisui, line 400: We will need to be consistent for the arguments when creating `FormalOUTSVFGNode` and `FormalInSVFGNode`.  If the argument is a `const MRVer* `, we do not need to passing pointsTo since an MRVer carries the pointsTo as a field. Please check

testSuite-test.sh
```diff
@@ -0,0 +1,21 @@
+declare -a arr=("AvgTopLvlPtsSize" "TotalNode" "TotalEdge")
```
yuleisui, line 1: please remove these two .sh files.

include/Graphs/SVFGNode.h
```diff
@@ -91,19 +91,32 @@ class FormalINSVFGNode : public MRSVFGNode
...
         return fun;
     }
+
+    /// Return points-to of the MR
+    inline const PointsTo& getPointsTo() const
+    {
+        return cpts;
+    }
+
+    inline const MRVer* getVer() const
```
yuleisui, line 116: can you rename it to be getMRVer() and the same for all other places?

lib/Graphs/SVFG.cpp
```diff
@@ -36,6 +36,8 @@
 using namespace SVF;
 using namespace SVFUtil;
 
+NodeID totalSVFGNode;
```
yuleisui, line 39: Remove this `totalSVFGNode `

PR #514, opened 2021-09-10
Fix SVFG build error on clang 12

adriaanjacobs:
Clang 12 is able to statically determine that `phi` will always be NULL here. This caused build problems on my end, and seems like just a typo.


yuleisui:
thanks

PR #544, opened 2021-11-03
read and write for SVFG

Wren71:
Co-authored-by: lukekoko <github@lukekoko.net>


yuleisui:
Would you be able to add this write/read into your differential analysis of SVF-Test-Suite using crux-bc https://github.com/SVF-tools/Test-Suite/tree/master/test_cases_bc/crux-bc?

PR #550, opened 2021-11-18
Bound FS call graph

mbarbar:
As discussed, this bounds the FS call graph by that coming from Andersen's.

### DIFF
--- a/lib/WPA/FlowSensitive.cpp
+++ b/lib/WPA/FlowSensitive.cpp
@@ -612,29 +612,39 @@ bool FlowSensitive::isStrongUpdate(const SVFGNode* node, NodeID& singleton)
 bool FlowSensitive::updateCallGraph(const CallSiteToFunPtrMap& callsites)
 {
     double start = stat->getClk();
-    CallEdgeMap potentialNewEdges;
-    onTheFlyCallGraphSolve(callsites, potentialNewEdges);
+    CallEdgeMap newEdges;
+    onTheFlyCallGraphSolve(callsites, newEdges);
 
     // Bound the new edges by the Andersen's call graph.
-    // TODO: we want this to be an assertion.
+    // TODO: we want this to be an assertion eventually.
     const CallEdgeMap &andersCallEdgeMap = ander->getIndCallMap();
-    CallEdgeMap newEdges;
-    for (typename CallEdgeMap::value_type &csfs : potentialNewEdges)
+    for (typename CallEdgeMap::value_type &csfs : newEdges)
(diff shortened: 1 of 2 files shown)


lib/WPA/FlowSensitive.cpp
```diff
@@ -612,8 +612,32 @@ bool FlowSensitive::isStrongUpdate(const SVFGNode* node, NodeID& singleton)
...
 {
     double start = stat->getClk();
+    CallEdgeMap potentialNewEdges;
+    onTheFlyCallGraphSolve(callsites, potentialNewEdges);
+
+    // Bound the new edges by the Andersen's call graph.
+    // TODO: we want this to be an assertion.
+    const CallEdgeMap &andersCallEdgeMap = ander->getIndCallMap();
     CallEdgeMap newEdges;
-    onTheFlyCallGraphSolve(callsites, newEdges);
```
yuleisui, line 616: Can we still invoke this function and then filter out the resulting new edges? This will be consistent with other pointer analyses implementations.

yuleisui: The filtering can be a separate function in order to be more flexible to change to an assertion for soundness later.

mbarbar: Hmm I don't fully understand. Currently it passes in potentialNewEdges into this function, then builds newEdges from that. Do you mean to pass in newEdges then remove elements from it?

mbarbar: Yes, I can make it its own function, what kind of API do you envision?

yuleisui: > Hmm I don't fully understand. Currently it passes in potentialNewEdges into this function, then builds newEdges from that. Do you mean to pass in newEdges then remove elements from it?

Yes

yuleisui: > Yes, I can make it its own function, what kind of API do you envision?

If the implementation is short, we can just put it in this function.

mbarbar: Hmm, can't be in `onTheFlyCallGraphSolve` because it has no notion of auxiliary analysis.

yuleisui: No, I mean inside `FlowSensitive::updateCallGraph`

mbarbar: Oh okay, it is 30 lines now, I'm just testing now before pushing.

PR #560, opened 2021-11-29
Implement PointsTo

mbarbar:
Probably the largest PR, mostly the `PointsTo` and CBV classes.

Also some shuffling had to be done with regards to #includes and forward declarations. I was thinking for the future we can use a tool like [iwyu](https://include-what-you-use.org/), move implementation out of header files (to reduce includes in headers), and maybe define an include hierararchy for the `include` directory (e.g., `Util` should never include `WPA`, etc.).


yuleisui:
> Probably the largest PR, mostly the `PointsTo` and CBV classes.
> 
> Also some shuffling had to be done with regards to #includes and forward declarations. I was thinking for the future we can use a tool like [iwyu](https://include-what-you-use.org/), move implementation out of header files (to reduce includes in headers), and maybe define an include hierararchy for the `include` directory (e.g., `Util` should never include `WPA`, etc.).

I can see that quite a few `Util` files include `WPA` headers.  In this regard, I believe it is better to move the `PointsTo` class under the `MemoryModel` folder?

mbarbar:
Is this in `lib` or `include`? For `include`, I only see `Options.h` (we should probably someday move all these enums into `Options.h` so it includes nothing). For `lib`, it's not as big a deal but not ideal in my opinion. But I can still move `PointsTo` to `MemoryModel` if preferred (it will add an extra dependency on `MemoryModel` everywhere though, including in `Util`).

yuleisui:
> iwyu

Any idea for using iwyu?

> Is this in `lib` or `include`? For `include`, I only see `Options.h` (we should probably someday move all these enums into `Options.h` so it includes nothing). For `lib`, it's not as big a deal but not ideal in my opinion. But I can still move `PointsTo` to `MemoryModel` if preferred (it will add an extra dependency on `MemoryModel` everywhere though, including in `Util`).

It would be good to move PointsTo under `MemoryModel` since we have the points-to data structures and points-to analysis there. It will be easier for users to read.
I saw `PointerAnalysis` includes the header file of `PointsTo.h`, we can then avoid including  `PointsTo.h` for many other places where `PointerAnalysis.h` and its subclasses are included.

mbarbar:
>Any idea for using iwyu?

To be honest I'm not sure how haha. Just I ran into some hairy circular includes, and in some cases, some includes were unused (probably previously used, then the use was removed without removing the include).

>It would be good to move PointsTo under MemoryModel

Yes sure, one second.

>I saw PointerAnalysis includes the header file of PointsTo.h, we can then avoid including PointsTo.h for many other places where PointerAnalysis.h and its subclasses are included.

My personal view is that it's better to be explicit, and it avoids the scrambling that happens when PointerAnalysis may be no longer used and the include removed for example. I'm okay either way though.

yuleisui:
> > Any idea for using iwyu?
> 
> To be honest I'm not sure how haha. Just I ran into some hairy circular includes, and in some cases, some includes were unused (probably previously used, then the use was removed without removing the include).
> 
> > It would be good to move PointsTo under MemoryModel
> 
> Yes sure, one second.
> 
> > I saw PointerAnalysis includes the header file of PointsTo.h, we can then avoid including PointsTo.h for many other places where PointerAnalysis.h and its subclasses are included.
> 
> My personal view is that it's better to be explicit, and it avoids the scrambling that happens when PointerAnalysis may be no longer used and the include removed for example. I'm okay either way though.

Yes, would be good to include only `PointsTo.h` if `PointerAnalysis` is unused.

mbarbar:
Done. I've removed the include for every file that includes PointerAnalysis or PointerAnalysisImpl.

yuleisui:
The total time of Andersen is faster if this patch is applied, however, the Average pts sizes are still inconsistent between these two runs, which hopefully will be investigated later.

---before---
AvgTopLvlPtsSize    510.535
TotalTime           96.564

---after----
AvgTopLvlPtsSize    510.666
TotalTime           85.363


I will merge this soon.

mbarbar:
If this is the bitcode we discussed earlier, the inconsistency is present before any of these patches are applied (I tested on what was upstream at the time). I haven't been able to reproduce on smaller bitcode unfortunately.

yuleisui:
> If this is the bitcode we discussed earlier, the inconsistency is present before any of these patches are applied (I tested on what was upstream at the time). I haven't been able to reproduce on smaller bitcode unfortunately.

Yes, it is the same bc, will take a look at the inconsistency later. I was wondering why this `PointsTo` patch has a better performance. Is that because the core bitvector is used?

mbarbar:
Hmm, CBV is not default (needs `-pt-type=cbv`). It is possible the different results affected the performance... did you see this speed up with other bitcode?

yuleisui:
When I enabled CBV `-pt-type=cbv`, Andersen's analysis becomes slower and the average pts look much larger. It also triggers an assertion error if `wpa -ander -svfg spdk_tgt.bc`

```
AvgTopLvlPtsSize    764.236
TotalTime           184.612
```
```
Assertion failed: (it != IDToNodeMap.end() && "Node not found!"), function getGNode, file /Users/yulei/Documents/workspace/SVF/include/Graphs/GenericGraph.h, line 394.
zsh: abort      wpa -ander -svfg ~/Downloads/spdk_tgt.bc -pt-type=cbv
```

yuleisui:
Also an assertion for `-pt-type=bv`.
```
Assertion failed: (it != IDToNodeMap.end() && "Node not found!"), function getGNode, file /Users/yulei/Documents/workspace/SVF/include/Graphs/GenericGraph.h, line 394.
zsh: abort      wpa -ander -svfg ~/Downloads/spdk_tgt.bc -pt-type=bv
```

mbarbar:
I've seen this assertion very rarely on a small bc, not sure what causes it, and I think I can hit it with SBV.

For the extended time... it is possible that the unclusted CBV for this program is just really bad. Also possible that the variance in points-to set size is a result of the bug and so the time cannot be trusted. Have you seen this problem on non-spdk?

yuleisui:
Just ran ` wpa -ander -svfg -pt-type=bv` on a few crux-bcs (https://github.com/SVF-tools/Test-Suite/tree/master/test_cases_bc/crux-bc). 

`curl.bc` also triggered the same assertion.

mbarbar:
Hmm, is this triggering easily? I have it running on loop (60 or 70 so far) and no crash so far.

yuleisui:
Easily trigger using around 3secs

```
Assertion failed: (it != IDToNodeMap.end() && "Node not found!"), function getGNode, file /Users/yulei/Documents/workspace/SVF/include/Graphs/GenericGraph.h, line 394.
zsh: abort      wpa -ander -svfg curl.bc -pt-type=cbv
wpa -ander -svfg curl.bc -pt-type=cbv  2.57s user 0.13s system 93% cpu 2.880 total
```

mbarbar:
Really strange, running on master, I cannot trigger for curl.bc on loop! It's also taking less time, 1.2s total, on the server.

yuleisui:
I am running on mac. It triggers the assertion for every single run. I just had a test on CI here. Unfortunately, it triggers the same assertion on mac os but not on ubuntu...

https://github.com/SVF-tools/SVF-example/runs/4352288641?check_suite_focus=true#step:5:437

mbarbar:
Oh no, nasty! I'll have to try on mac then.

Let's discuss more tomorrow morning.

yuleisui:
I suspect that it might be a subtle bug that ubuntu's GCC optimization didn't capture.

PR #562, opened 2021-11-29
Implement clustering

mbarbar:
This should be the final PR. The `::cluster` method is large, eventually I can shrink it after experiments get going, and I will remove the dependency on `BVDataPTAImpl`.


lib/WPA/Andersen.cpp
```diff
@@ -187,6 +190,18 @@ void Andersen::initialize()
  */
 void Andersen::finalize()
 {
+    // TODO: check -stat too.
```
yuleisui, line 193: What's these few lines for in the finalize method of Andersen?

lib/WPA/Andersen.cpp
```diff
@@ -811,6 +826,25 @@ void Andersen::updateNodeRepAndSubs(NodeID nodeId, NodeID newRepId)
     consCG->resetSubs(nodeId);
 }
 
+void Andersen::cluster(void) const
+{
+    assert(Options::MaxFieldLimit == 0 && "Andersen::cluster: clustering for Andersen's is currently only supported in field-insesnsitive analysis");
+    Steensgaard *steens = Steensgaard::createSteensgaard(pag);
```
yuleisui, line 832: Why we need a steensgaard here?

mbarbar: We need data to perform clustering.

lib/WPA/Andersen.cpp
mbarbar: This is to print the statistics of how effective the clustering actually was (how many words needed).

lib/WPA/Andersen.cpp
yuleisui: I see. This is to use steensgaard to boost andersen?

mbarbar: Exactly

PR #563, opened 2021-11-30
Fix CBV

mbarbar:
Was not comparing with RHS words size.

This actually thus only seemed to affect persistent which uses operator==. This fixes the inconsistency we saw.

I couldn't reproduce the crashing though, could you test if this fixes that?

### DIFF
--- a/lib/Util/CoreBitVector.cpp
+++ b/lib/Util/CoreBitVector.cpp
@@ -161,7 +161,7 @@ bool CoreBitVector::operator==(const CoreBitVector &rhs) const
     }
 
     // Make sure both got to the end at the same time.
-    return lhsSetIndex >= words.size() && rhsSetIndex >= words.size();
+    return lhsSetIndex >= words.size() && rhsSetIndex >= rhs.words.size();
 }
 
 bool CoreBitVector::operator!=(const CoreBitVector &rhs) const


yuleisui:
> 

It works and no crash now!

yuleisui:
Placeholder notes here: both CBV and BV is slower than SBV for some larger bcs like `bash` in crux-bc.

--------SBV-----
AvgTopLvlPtsSize    254.251
TotalTime           15.409

--------CBV-----
AvgTopLvlPtsSize    254.251
TotalTime           20.134

--------BV-----
AvgTopLvlPtsSize    254.251
TotalTime           22.208

mbarbar:
Good to hear :)

Hmm I see. 

There are 2 issues:
1. It is possibly bad chance that a very bad mapping is produced. My experience though is that CBV is always better than SBV.
2. If you are using persistent... like I mentioned, CBV suffers a little with persistent due to operator==. I want to improve this in the future. 

If you try this with FSPTA, I wonder if you'll see the same, or that the unions start causing enough of a problem that CBV is better.

Another thing I want to do is to eventually add an option to compile with `-march=native` and `-fopenmp` to maybe get some nicer SIMD? Will have to look more closely at the assembly.

PR #566, opened 2021-12-02
Suppress warnings if -Werror=return-type is turned on

yuleisui:
@mbarbar unfortunately the compiler can stripoff assertions and assume that there is no return for a non-void function. 

Any more elegant way to solve this?


mbarbar:
Yeah this is annoying... I can't think of a better way except to define our own assert which prints a message and aborts, but this comes with its own problems.

PR #520, opened 2021-09-29
Add support for LLVM 13

charlesxsh:

PR #567, opened 2021-12-02
Add support for LLVM 13 release

xushengj:
The upstream LLVM git repository has changed major version to 14. There are some breaking changes in LLVM API, and this PR fixes compilation issues against the new API.

Most of this patch is adapted from PR #520

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -10,13 +10,16 @@
 ########
 SVFHOME=$(pwd)
 sysOS=$(uname -s)
-MacLLVM="https://github.com/llvm/llvm-project/releases/download/llvmorg-12.0.0/clang+llvm-12.0.0-x86_64-apple-darwin.tar.xz"
-UbuntuLLVM="https://github.com/llvm/llvm-project/releases/download/llvmorg-12.0.0/clang+llvm-12.0.0-x86_64-linux-gnu-ubuntu-16.04.tar.xz"
+MacLLVM="https://github.com/llvm/llvm-project/releases/download/llvmorg-13.0.0/clang+llvm-13.0.0-x86_64-apple-darwin.tar.xz"
+UbuntuLLVM="https://github.com/llvm/llvm-project/releases/download/llvmorg-13.0.0/clang+llvm-13.0.0-x86_64-linux-gnu-ubuntu-16.04.tar.xz"
 MacZ3="https://github.com/Z3Prover/z3/releases/download/z3-4.8.8/z3-4.8.8-x64-osx-10.14.6.zip"
 UbuntuZ3="https://github.com/Z3Prover/z3/releases/download/z3-4.8.8/z3-4.8.8-x64-ubuntu-16.04.zip"
 MacCTIR="https://github.com/mbarbar/ctir/releases/download/ctir-10.c3/ctir-clang-v10.c3-macos10.15.zip"
 UbuntuCTIR="https://github.com/mbarbar/ctir/releases/download/ctir-10.c3/ctir-clang-v10.c3-ubuntu18.04.zip"
-LLVMHome="llvm-12.0.0.obj"
+
+# Keep LLVM version suffix for version checking and better debugging
+# keep the version consistent with LLVM_DIR in setup.sh and llvm_version in Dockerfile
+LLVMHome="llvm-13.0.0.obj"
 Z3Home="z3.obj"
 CTIRHome="ctir.obj"
 
--- a/.github/workflows/github-action.yml
+++ b/.github/workflows/github-action.yml
@@ -13,19 +13,29 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-18.04, macos-latest]
+        os: [ubuntu-18.04, macos-11]
     steps:
       # checkout the repo
       - uses: actions/checkout@v2
-      # setup the environment 
+      # setup the environment
+      - name: mac-setup
+        if: runner.os == 'macOS'
+        uses: maxim-lobanov/setup-xcode@v1
+        with:
+          xcode-version: 12.5.1
+      - name: mac-setup-workaround
+        if: runner.os == 'macOS'
+        run: ln -sfn /Applications/Xcode_12.5.1.app /Applications/Xcode.app
+      - name: ubuntu-setup
+        if: runner.os == 'Linux'
+        run: |
+          sudo apt-get update
+          sudo add-apt-repository -y ppa:ubuntu-toolchain-r/test
+          sudo apt-get update
+          sudo apt-get install cmake gcc g++ nodejs doxygen graphviz
       - name: env-setup
         run: |
           echo $RUNNER_OS
-          if [ "$RUNNER_OS" = "Linux" ]; then sudo apt-get update       ; fi
-          if [ "$RUNNER_OS" = "Linux" ]; then sudo add-apt-repository -y ppa:ubuntu-toolchain-r/test       ; fi
-          if [ "$RUNNER_OS" = "Linux" ]; then sudo apt-get update       ; fi
-          if [ "$RUNNER_OS" = "Linux" ]; then sudo apt-get install cmake gcc g++ nodejs doxygen graphviz; fi
-          if [ "$RUNNER_OS" = "macOS" ]; then export SDKROOT="/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX11.1.sdk" ; fi
           if [ "${{github.event_name}}" = "push" ] && [ "${{github.ref == 'refs/heads/master' }}" ]; then 
             git config --global user.email ${{secrets.npm_email}}
             git config --global user.name "GitHub Actions Build"
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.4.3)
+cmake_minimum_required(VERSION 3.13.4)
 
 project("SVF")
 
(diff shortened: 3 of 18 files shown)


yuleisui:
It seems to me you didn't change build.sh to download llvm-13 or 14?

xushengj:
No, I did not change that. We always compile LLVM and Z3 ourselves and then invoke CMake manually to compile SVF, so it is not necessary to modify build.sh.

I can manually bump the LLVM version there, but I don't think this gives much value. One reason against using build.sh is that it downloads prebuilt binaries with OS suffix. (Ubuntu 16.04 for example). However, none of our environments is using Ubuntu 16.04. I think a better alternative is to modify build.sh to let it select the binaries with the correct OS suffix. We can create a separate PR for this if needed

yuleisui:
> No, I did not change that. We always compile LLVM and Z3 ourselves and then invoke CMake manually to compile SVF, so it is not necessary to modify build.sh.
> 
> I can manually bump the LLVM version there, but I don't think this gives much value. One reason against using build.sh is that it downloads prebuilt binaries with OS suffix. (Ubuntu 16.04 for example). However, none of our environments is using Ubuntu 16.04. I think a better alternative is to modify build.sh to let it select the binaries with the correct OS suffix. We can create a separate PR for this if needed



> No, I did not change that. We always compile LLVM and Z3 ourselves and then invoke CMake manually to compile SVF, so it is not necessary to modify build.sh.

Thanks, I see. I assume that if we build on top of llvm-13 or llvm-14, this patch will also work?

> 
> I can manually bump the LLVM version there, but I don't think this gives much value. One reason against using build.sh is that it downloads prebuilt binaries with OS suffix. (Ubuntu 16.04 for example). However, none of our environments is using Ubuntu 16.04. I think a better alternative is to modify build.sh to let it select the binaries with the correct OS suffix. We can create a separate PR for this if needed

The LLVM's Ubuntu 16.04 binary also works for general ubuntu's as in our CI (GitHub workflow). The build.sh is simply a one-button build for both ubuntu and mac os machines.

yuleisui:
> 

Would be perfect if you can make a pull request for the `build.sh` to build on top of llvm-13 or llvm-14.

yuleisui:
A compilation error.
https://github.com/SVF-tools/SVF/runs/4403443166?check_suite_focus=true#step:4:189

This github workflow was not written by me, but it seems to be a macos SDK problem (needs to be version 11.3?). 
https://github.com/SVF-tools/SVF/blob/27519cdf2a9c4b9feb073cd3f679fe3bbdbb8265/.github/workflows/github-action.yml#L28

xushengj:
I changed the Ubuntu build to use LLVM 13.0. I think the prebuilt binary for mac is built for a more recent SDK version than the one from the current macos-latest, so I left the MAC build to still use LLVM 12.0.

yuleisui:
> I changed the Ubuntu build to use LLVM 13.0. I think the prebuilt binary for mac is built for a more recent SDK version than the one from the current macos-latest, so I left the MAC build to still use LLVM 12.0.

Thanks. It would be good to make it consistent for the migration. I suspect the reason is that we will need to export the correct SDK path. The below line might need to change? Any idea to fix this?

https://github.com/SVF-tools/SVF/blob/27519cdf2a9c4b9feb073cd3f679fe3bbdbb8265/.github/workflows/github-action.yml#L28

build.sh
```diff
@@ -10,13 +10,13 @@
...
-MacLLVM="https://github.com/llvm/llvm-project/releases/download/llvmorg-12.0.0/clang+llvm-12.0.0-x86_64-apple-darwin.tar.xz"
-UbuntuLLVM="https://github.com/llvm/llvm-project/releases/download/llvmorg-12.0.0/clang+llvm-12.0.0-x86_64-linux-gnu-ubuntu-16.04.tar.xz"
+MacLLVM="https://github.com/llvm/llvm-project/releases/download/llvmorg-13.0.0/clang+llvm-13.0.0-x86_64-apple-darwin.tar.xz"
+UbuntuLLVM="https://github.com/llvm/llvm-project/releases/download/llvmorg-13.0.0/clang+llvm-13.0.0-x86_64-linux-gnu-ubuntu-16.04.tar.xz"
 MacZ3="https://github.com/Z3Prover/z3/releases/download/z3-4.8.8/z3-4.8.8-x64-osx-10.14.6.zip"
 UbuntuZ3="https://github.com/Z3Prover/z3/releases/download/z3-4.8.8/z3-4.8.8-x64-ubuntu-16.04.zip"
 MacCTIR="https://github.com/mbarbar/ctir/releases/download/ctir-10.c3/ctir-clang-v10.c3-macos10.15.zip"
 UbuntuCTIR="https://github.com/mbarbar/ctir/releases/download/ctir-10.c3/ctir-clang-v10.c3-ubuntu18.04.zip"
-LLVMHome="llvm-12.0.0.obj"
+LLVMHome="llvm.obj"
```
yuleisui, line 19: Would be good to make this explicit for version checking and better debugging purposes

xushengj:
All requested changes are done.

yuleisui:
> All requested changes are done.

Thank you and hopefully, one more request and I will merge:

Can you separate the commit of `BreakConstantExpr` from this pr and make it a new one? So that we can take a look at the impact of `BreakConstantExpr` later. It also makes sense of this pr only for the purpose of upgrading to llvm-13.

xushengj:
The change to BreakConstantExpr is required to pass the auto-test when using LLVM 13.0 prebuilt binary. I will create a PR for that change only, then rebase this PR after that one is merged

yuleisui:
> The change to BreakConstantExpr is required to pass the auto-test when using LLVM 13.0 prebuilt binary. I will create a PR for that change only, then rebase this PR after that one is merged

The BreakConstantExpr one has been merged.

xushengj:
Ok. Maybe I should change the title to note that this PR just moves to LLVM 13.0....

yuleisui:
This line also needs to be changed since `setup.sh` will be invoked by `build.sh`

https://github.com/xushengj/SVF/blob/dfb2a3c7336fec3aecfb1891d17cff89d0d59b5b/setup.sh#L21

yuleisui:
> Ok. Maybe I should change the title to note that this PR just moves to LLVM 13.0....

Yes

yuleisui:
When building on my local mac machine using this patch, I got the following error (also bookkeeping purposes for others who work on mac). 

I saw your GitHub action changes, do you have a quick fix or any suggestions? 

```
LLVM_DIR=/Users/yulei/Documents/workspace/SVF-xusheng/llvm-13.0.0.obj
Z3_DIR=/Users/yulei/Documents/workspace/SVF-xusheng/z3.obj
-- The C compiler identification is AppleClang 12.0.5.12050022
-- The CXX compiler identification is AppleClang 12.0.5.12050022
-- Check for working C compiler: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/cc
-- Check for working C compiler: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/c++
-- Check for working CXX compiler: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Found ZLIB: /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX11.3.sdk/usr/lib/libz.tbd (found version "1.2.11") 
-- Found LibXml2: /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX11.3.sdk/usr/lib/libxml2.tbd (found version "2.9.4") 
CMake Error at llvm-13.0.0.obj/lib/cmake/llvm/LLVMDistributionSupport.cmake:18 (include_guard):
  Unknown CMake command "include_guard".
Call Stack (most recent call first):
  llvm-13.0.0.obj/lib/cmake/llvm/AddLLVM.cmake:1 (include)
  CMakeLists.txt:33 (include)
```

xushengj:
> 

I have limited experience with Mac, but from the error message, you probably need to use a more recent version of CMake.

yuleisui:
> > 
> 
> I have limited experience with Mac, but from the error message, you probably need to use a more recent version of CMake.

Yes, you are right. After updating `cmake` to the latest version, the problem was solved.

I found the instruction here for the minimum requirement of cmake for LLVM 
https://llvm.org/docs/CMake.html#quick-start

Can you update this line to Version 3.13.4 to reflect it? (Unfortunately, a higher version number does not mean a more recent release of cmake here, a bit odd)
https://github.com/SVF-tools/SVF/blob/master/CMakeLists.txt#L1

xushengj:
I do not recommend this change. Only LLVM 13.0 release requires CMake 3.13.4. We can still compile SVF against LLVM 12 or maybe older versions that do not require a new CMake version. Changing the CMake minimum version for SVF can break that.

yuleisui:
> 

I suggest doing so since updating `cmake` is easier for end-users and we will move forward anyway, just in case users under all llvm versions (13 or later) will compile.

BTW: this line will also need to change to be consistent with 13.0.0:
https://github.com/SVF-tools/SVF/blob/master/Dockerfile#L10

PR #576, opened 2021-12-17
Make versioning multithreaded

mbarbar:
This makes versioning multithreaded with the `-versioning-threads=N` option. Performance should roughly improve by N-ish times. Single-threaded should also be faster than before. Importantly, memory cost is low. In small bitcode, VFS sometimes will use slightly more memory than FS (overhead) but as things grow, especially in really large bitcode, VFS wins (e.g., 51.5 GB vs 30.5 GB for git). I haven't seen a regression yet for large BC.

To measure times properly, use `-clock-type=wall`. It only works with `opt-svfg=false`.


yuleisui:
When using more than one thread, I assume that opt-svfg will set to false by default?

mbarbar:
Ah no. I'll add an assert.

yuleisui:
Maybe, we can set `opt-svfg=false` when N>1.

mbarbar:
It's VFS that is the problem, not the MT. Specifically the deltaSource function does not handle Opt.

As aasiden, deltaSource is overapproximate. Some more precision may save memory.

17 Dec. 2021 16:38:03 Yulei Sui ***@***.***>:

PR #578, opened 2021-12-17
Fix build errors (-Werror, wpa: missing pthreads)

251:
I wasn't able to build the latest commit with gcc 11.1.0 and had to fix some minor issues.

* wpa: pthreads is not linked
* missing return value for non-void functions (fails due to `-Werror`) in three cases


yuleisui:
Thanks!

PR #582, opened 2021-12-22
Fix memory leaks

### LINKED ISSUE #580: Memory leaks
Hi,

When I'm using SVF on a small program:
```CPP
  auto * svf_module = SVF::LLVMModuleSet::getLLVMModuleSet()->buildSVFModule(*llvm_module);
  svf_module->buildSymbolTableInfo();
  SVF::PAGBuilder builder;
  auto * pag = builder.build(svf_module);
```
my application leaks hundreds of thousands of objects and roughly 14MB of memory. I had to use a rather large list of functions to tear down SVF:

```CPP
  // clean up
  auto * icfg = pag->getICFG();
  SVF::PAG::releasePAG();
  delete icfg;
  SVF::LLVMModuleSet::releaseLLVMModuleSet();
  delete svf_module;
  auto * extAPI = SVF::ExtAPI::getExtAPI();
  delete extAPI;
  SVF::NodeIDAllocator::unset();
  auto * threadAPI = SVF::ThreadAPI::getThreadAPI();
  delete threadAPI;
```

Is this the way it is supposed to be used or did I miss something in the documentation/examples? Furthermore, I had to change SVF to leak less memory (see: https://github.com/251/SVF/tree/fix_leaks):

* `GenericGraph` does not delete edges (https://github.com/251/SVF/commit/a4d6a2b70b3c93a5d90be72317445e39de1ecf18)
* `SVFModule` does not free pointers in `FunctionSet` (https://github.com/251/SVF/commit/dbd4b726909eb73357da32462d8261a4dc2abab6)
* `SymbolTableInfo` does not free `DataLayout` (https://github.com/251/SVF/commit/bfbf2c41aca74a2c088352abfc91cbf11f5a7dd5)

I'd provide a PR but I have no idea what the ownership semantics is in most cases (e.g. Does PAG own ICFG or should it be a shared_ptr for external usage? Who owns the edges in GenericGraph - the graph or a node?).

251:
This fixes #580 and works at least for my small example. There were a few more issues and I hope I fixed them without breaking any invariants.

### DIFF
--- a/include/MemoryModel/PersistentPointsToCache.h
+++ b/include/MemoryModel/PersistentPointsToCache.h
@@ -45,20 +45,26 @@ public:
         initStats();
     }
 
+    /// Clear the cache.
+    void clear()
+    {
+      for (const Data *d : idToPts) delete d;
+      idToPts.clear();
+      ptsToId.clear();
+
+      unionCache.clear();
+      complementCache.clear();
+      intersectionCache.clear();
+    }
+
     /// Resets the cache removing everything except the emptyData it was initialised with.
     void reset(void)
     {
-        for (const Data *d : idToPts) delete d;
-        idToPts.clear();
-        ptsToId.clear();
+        clear();
 
         // Put the empty data back in.
         ptsToId[Data()] = emptyPointsToId();
         idToPts.push_back(new Data());
-
-        unionCache.clear();
-        complementCache.clear();
-        intersectionCache.clear();
 
         idCounter = 1;
         // Cache is empty...
--- a/include/Util/SVFModule.h
+++ b/include/Util/SVFModule.h
@@ -68,6 +68,11 @@ public:
     {
     }
 
+    ~SVFModule()
+    {
+      for (auto * f : FunctionSet)
+        delete f;
+    }
 
     static inline void setPagFromTXT(std::string txt)
     {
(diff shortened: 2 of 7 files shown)


yuleisui:
Shall we set all pointers to `nullptr` after the deletion, so that we can easily locate any exception?

include/Util/SVFModule.h
```diff
@@ -68,6 +71,17 @@ class SVFModule
     {
     }
 
+    ~SVFModule()
+    {
+      for (auto * f : FunctionSet)
+        delete f;
+
+      auto * extAPI = ExtAPI::getExtAPI();
+      delete extAPI;
```
yuleisui, line 80: Shall we set every pointer to `nullptr` after each deletion to avoid misusage?

include/MemoryModel/PersistentPointsToCache.h
```diff
@@ -45,21 +45,27 @@ class PersistentPointsToCache
...
+        clear();
 
         // Put the empty data back in.
         ptsToId[Data()] = emptyPointsToId();
         idToPts.push_back(new Data());
 
-        unionCache.clear();
-        complementCache.clear();
-        intersectionCache.clear();
-
```
yuleisui, line 62: @mbarbar could you take a look at this part to make sure the refactoring is consistent as before?

mbarbar: Looks good to me.

PR #586, opened 2021-12-23
Fix memory leaks

251:
Another proposal to fix some leaks. I'm not that confident about the CHG fix and have to see if the tests come back successfully.

### DIFF
--- a/include/WPA/Andersen.h
+++ b/include/WPA/Andersen.h
@@ -31,6 +31,7 @@
 #define ANDERSENPASS_H_
 
 #include "MemoryModel/PointerAnalysisImpl.h"
+#include "SVF-FE/CommonCHG.h"
 #include "WPA/WPAStat.h"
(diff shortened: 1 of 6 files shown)


include/WPA/Andersen.h
```diff
@@ -65,6 +66,10 @@ class AndersenBase:  public WPAConstraintSolver, public BVDataPTAImpl
         if (consCG != nullptr)
             delete consCG;
         consCG = nullptr;
+
+        auto * chg = getCHGraph();
+        delete chg;
+        chg = nullptr;
```
yuleisui, line 72: Would be good to move this destructor to cpp file. `#include "SVF-FE/CommonCHG.h"` can also be removed.

251: Done.

251:
Thanks for reviewing an merging so quickly!

PR #601, opened 2022-01-09
set allocator to null after delete

chenyang2022morecode:


yuleisui:
This is a good fix. Thanks!

PR #609, opened 2022-01-24
Add reverse-dense strategy

mbarbar:


lib/Util/Options.cpp
```diff
@@ -26,6 +26,7 @@ namespace SVF
         llvm::cl::desc("Method of allocating (LLVM) values and memory objects as node IDs"),
         llvm::cl::values(
             clEnumValN(NodeIDAllocator::Strategy::DENSE, "dense", "allocate objects together and values together, separately"),
```
yuleisui, line 28: Would be good to explain/append a bit more about dense and reverse-dense ID arrangement.

mbarbar: Done

yuleisui:
Thanks, will merge once CI is done.

PR #610, opened 2022-01-25
compute accumulateConstantOffset of a GepStmt

yuleisui:
@mbarbar Can you help test whether we can obtain the correct offset if using GepStmt->accumulateConstantOffset().

Please also feel free to suggest better naming conventions based on our discussion.

### DIFF
--- a/lib/MemoryModel/LocationSet.cpp
+++ b/lib/MemoryModel/LocationSet.cpp
@@ -45,7 +45,7 @@ bool LocationSet::addOffsetValue(const Value* offsetVal, const Type* type)
     return true;
 }
 
-/// Return TRUE if all offset values are constants
+/// Return true if all offset values are constants
 bool LocationSet::isConstantOffset() const
 {
     for(auto it : offsetValues){
@@ -55,7 +55,7 @@ bool LocationSet::isConstantOffset() const
     return true;
 }
 
-/// Return element number given a type
+/// Return element number of a type
 /// (1) StructType, return flatterned number fields
 /// (2) ArrayType, return number of elements
 /// (3) PointerType, return the element number of the pointee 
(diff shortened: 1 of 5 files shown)


lib/MemoryModel/LocationSet.cpp
```diff
@@ -55,12 +55,74 @@ bool LocationSet::isConstantOffset() const
...
+        const Type* type = offsetValues[i].second;
+        const ConstantInt *op = SVFUtil::dyn_cast<ConstantInt>(value);
+        assert(op && "not a constant offset?");
+        /// if this gep only has one operand, the gepIterType must be the pointer type, and we will need to retrieve size of its elementType.
+        if(offsetValues.size()==1){
+            assert(SVFUtil::isa<PointerType>(type) && "If gep has only one operand, its gepIterType must be PointerType!");
+            sz = getElementNum(type);
+        }
+        totalConstOffset += op->getSExtValue() * sz; 
+        sz *= getElementNum(type);
```
mbarbar, line 123: Is this accumulation necessary, since getElementNum handles flattening?

yuleisui: The flattening mechanism builds a map via `const vector<u32_t> &flattenVec = SymbolTableInfo::SymbolInfo()->getFlattenedOffsetVec(structTy)` to map a field index (from a gep instruction) to a flattened index of a struct type `structTy`. 

`j = flattenVec[I];` // `i` is the field index/offset in the gep instruction  and `j` is the index after flattening `structTy`.

For example
````
struct inner_struct{
int x;
int y;
}
struct structTy{
int a[10];
inner_struct b;
float c;
};

%q = getelementptr %structTy, %structTy* %p, i32 0, i32 2
````

The field index is 2 for accessing `c` in `structTy` at this gep instruction. `flattenVec[2]` returns `12=10+2`.

However, `getFlattenedOffsetVec` is only used to flatten structs for now. We will need `getElementNum` to calculate the sizes for all other types, and also recursively calculate the size of pointee if the type is a PointerType.

mbarbar: Oh sorry, I meant the accumulation with respect to `sz` (`*=`). For example, in an array of array of array, we have the size of the first (encompassing itself, the second, and the third), times the second (encompassing itself and the third), time the third (encompassing itself). (May be off by one in the example.)

yuleisui: I didn't quite get that. `sz *= getElementNum(type)` is to calculate the size for the next loop iteration, based on the equation at line 103.

mbarbar: `getElementNum(t)` accounts for flattening, so I may be misunderstanding why `sz` is multiplied rather than just assigned.

PR #634, opened 2022-02-16
Remove some LLVM dependencies

mbarbar:
This looks big but it's mostly removing namespaces, changing types to our types, etc.

The only problem is `TODO-os`. These are instances where we cannot print to `std::cout` because we are trying to print things like LLVM instructions and callsites. We need a solution like a to string function or preferably to just not print instructions (print SVF instructions/edges instead).

### DIFF
--- a/include/Util/CxtStmt.h
+++ b/include/Util/CxtStmt.h
@@ -108,7 +108,7 @@ public:
     /// Dump CxtStmt
     inline void dump() const
     {
-        SVFUtil::outs() << "[ Current Stmt: " << SVFUtil::getSourceLoc(inst) << " " << *inst << "\t Contexts: " << cxtToStr() << "  ]\n";
+        // TODO-os SVFUtil::outs() << "[ Current Stmt: " << SVFUtil::getSourceLoc(inst) << " " << *inst << "\t Contexts: " << cxtToStr() << "  ]\n";
     }
 
 protected:
@@ -174,7 +174,7 @@ public:
     /// Dump CxtThreadStmt
     inline void dump() const
     {
-        SVFUtil::outs() << "[ Current Thread id: " << tid << "  Stmt: " << SVFUtil::getSourceLoc(inst) << " " << *inst << "\t Contexts: " << cxtToStr() << "  ]\n";
+        // TODO-os SVFUtil::outs() << "[ Current Thread id: " << tid << "  Stmt: " << SVFUtil::getSourceLoc(inst) << " " << *inst << "\t Contexts: " << cxtToStr() << "  ]\n";
     }
 
 private:
@@ -281,11 +281,15 @@ public:
         std::string cycle = incycle?", incycle":"";
 
         if(forksite)
-            SVFUtil::outs() << "[ Thread: $" << SVFUtil::getSourceLoc(forksite) << "$ " << *forksite  << "\t Contexts: " << cxtToStr()
-                            << loop << cycle <<"  ]\n";
+        {
+            //TODO-os SVFUtil::outs() << "[ Thread: $" << SVFUtil::getSourceLoc(forksite) << "$ " << *forksite  << "\t Contexts: " << cxtToStr()
+            //TODO-os                << loop << cycle <<"  ]\n";
+        }
         else
+        {
             SVFUtil::outs() << "[ Thread: " << "main   "  << "\t Contexts: " << cxtToStr()
                             << loop << cycle <<"  ]\n";
+        }
     }
 protected:
     CallStrCxt cxt;
--- a/include/Util/CxtStmt.h
+++ b/include/Util/CxtStmt.h
@@ -108,7 +108,7 @@ public:
     /// Dump CxtStmt
     inline void dump() const
     {
-        // TODO-os SVFUtil::outs() << "[ Current Stmt: " << SVFUtil::getSourceLoc(inst) << " " << *inst << "\t Contexts: " << cxtToStr() << "  ]\n";
+        SVFUtil::outs() << "[ Current Stmt: " << SVFUtil::getSourceLoc(inst) << " " << SVFUtil::value2String(inst) << "\t Contexts: " << cxtToStr() << "  ]\n";
     }
 
 protected:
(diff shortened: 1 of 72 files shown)


yuleisui:
It looks one file have conflicts. Can you resolve them first?

mbarbar:
Oh whoops I was working off a stale master. Fixed. But please check that last commit in SymbolTableInfo.cpp... I thought you wanted to get rid of the byte stuff and replace it with field numbers but I guess not?

yuleisui:
> Oh whoops I was working off a stale master. Fixed. But please check that last commit in SymbolTableInfo.cpp... I thought you wanted to get rid of the byte stuff and replace it with field numbers but I guess not?

That fix is OK. However, the CI was not triggered. I don't know why? Was the CI triggered in your repo?

mbarbar:
The checks tab states "Workflow runs completed with no jobs".

lib/MemoryModel/SymbolTableInfo.cpp
```diff
@@ -594,31 +591,6 @@ void SymbolTableInfo::dump()
     outs() << "}\n";
 }
 
-/*
- * Get the type size given a target data layout
- */
-u32_t SymbolTableInfo::getTypeSizeInBytes(const Type* type)
```
yuleisui, line 600: Can we move these two functions to LLVMUtil?

yuleisui (commented):
I can see that most of the  TODO-os, we can use value2String to implement for now. Please fix.

lib/MemoryModel/SymbolTableInfo.cpp
yuleisui: > 

You can try to fix this part and the CI might trigger once you have a new commit.

mbarbar: Returned

mbarbar:
Ah yep my bad, not sure why I thought instructions weren't values...

Done now and CI kicked in.

include/Util/CxtStmt.h
```diff
@@ -108,7 +108,7 @@ class CxtStmt
     /// Dump CxtStmt
     inline void dump() const
     {
-        SVFUtil::outs() << "[ Current Stmt: " << SVFUtil::getSourceLoc(inst) << " " << *inst << "\t Contexts: " << cxtToStr() << "  ]\n";
+        // TODO-os SVFUtil::outs() << "[ Current Stmt: " << SVFUtil::getSourceLoc(inst) << " " << *inst << "\t Contexts: " << cxtToStr() << "  ]\n";
```
yuleisui, line 111: I think we can use value2String(inst) here for printing. All other TODO-os are the same?

mbarbar: Fixed.

lib/SABER/LeakChecker.cpp
```diff
@@ -223,17 +223,19 @@ void LeakChecker::validateSuccessTests(const SVFGNode* source, const SVFFunction
...
 
-    std::string funName = source->getFun()->getName().str();
+    std::string funName = source->getFun()->getName();
 
     if (success)
-        outs() << sucMsg("\t SUCCESS :") << funName << " check <src id:" << source->getId()
-               << ", cs id:" << *getSrcCSID(source)->getCallSite() << "> at ("
-               << getSourceLoc(cs->getCallSite()) << ")\n";
+    {
+        // TODO-os outs() << sucMsg("\t SUCCESS :") << funName << " check <src id:" << source->getId()
```
yuleisui, line 230: These can also be fixed.

lib/SVF-FE/SVFIRBuilder.cpp
```diff
@@ -445,10 +443,7 @@ NodeID SVFIRBuilder::getGlobalVarField(const GlobalVariable *gvar, u32_t offset,
 void SVFIRBuilder::InitialGlobal(const GlobalVariable *gvar, Constant *C,
                                u32_t offset)
 {
-    DBOUT(DPAGBuild,
-          outs() << "global " << *gvar << " constant initializer: " << *C
-          << "\n");
-
+    // TODO-os DBOUT(DPAGBuild, outs() << "global " << gvar << " constant initializer: " << *C << "\n");
```
yuleisui, line 446: Also this one.

lib/DDA/DDAClient.cpp
```diff
@@ -124,8 +124,8 @@ void FunptrDDAClient::performStat(PointerAnalysis* pta)
 
         ++morePreciseCallsites;
         outs() << "============more precise callsite =================\n";
-        outs() << *(nIter->second)->getCallSite() << "\n";
-        outs() << getSourceLoc((nIter->second)->getCallSite()) << "\n";
+        // TODO-os outs() << *(nIter->second)->getCallSite() << "\n";
+        // TODO-os outs() << getSourceLoc((nIter->second)->getCallSite()) << "\n";
```
yuleisui, line 128: These two

lib/MemoryModel/PointerAnalysis.cpp
```diff
@@ -343,7 +343,7 @@ void PointerAnalysis::printIndCSTargets(const CallICFGNode* cs, const FunctionSe
 {
     outs() << "\nNodeID: " << getFunPtr(cs);
     outs() << "\nCallSite: ";
-    cs->getCallSite()->print(outs());
+    // TODO-os cs->getCallSite()->print(outs());
```
yuleisui, line 346: For all print(outs()) can be written as ``outs() << value2String(cs->getCallSite())``

mbarbar:
Fixed.

All that is left are types, something to do with errors, and some file stuff. I'll work on the latter two next.

yuleisui:
> Fixed.
> 
> All that is left are types, something to do with errors, and some file stuff. I'll work on the latter two next.

We can make a method called type2String to print all types.

mbarbar:
The latter two are now handled, though I'm not sure how to test that `dumpPts` in the last commit. What option exercises that?

yuleisui:
> The latter two are now handled, though I'm not sure how to test that `dumpPts` in the last commit. What option exercises that?

Try `-print-pts` or `-print-all-pts`?

mbarbar:
Hmm does not seem to. No file produced on master with fspta.

yuleisui:
> Hmm does not seem to. No file produced on master with fspta.

Can you try Andersen or maybe might need a big dig up to see where it is used?

mbarbar:
Same thing unfortunately... I feel like I remember it was fspta but maybe something got disabled or changed? Might be wrong.

yuleisui:
> Same thing unfortunately... I feel like I remember it was fspta but maybe something got disabled or changed? Might be wrong.

Strange.. If `-print-pts` enabled, we can still print the points-to results, right?

mbarbar:
Yes points-to result are printing to stdout. But there isn't a per-program-point thing for address-taken objects anymore in either option (dumps to file).

yuleisui:
> Yes points-to result are printing to stdout. But there isn't a per-program-point thing for address-taken objects anymore in either option (dumps to file).

I see. I can see that we only have types that need to be printed. Could you add a type2String function and we can finalize all the TODO-os?

mbarbar:
Yep, I'll make a commit ready for review in the morning.

PR #635, opened 2022-02-21
Be more liberal with Z3 location

mbarbar:
We were struggling to use a user-built Z3, this fixes it. .so is not needed as far as I can tell, correct? And when you `make install` by hand the library is in `lib` not `bin`.


yuleisui:
@guanqin-123 could you take a look at this patch?

guanqin-123:
Yes, I think this change makes sense in solving the problem. I have tried, and z3 can be used smoothly.

PR #654, opened 2022-03-18
Aarch64

ddiepo-pjr:
Add aarch64 support to build.sh for Ubuntu Linux.

Fetch the arm version of LLVM.
Build z3 from source, as no pre-built version exists for aarch64.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -148,7 +148,7 @@ then
     then
         if [[ `uname -m` == "aarch64" ]]
         then
-            build_z3_from_source "$Z3_GIT" "$SVFHOME/$Z3Home"
+            build_z3_from_source "$Z3Git" "$SVFHOME/$Z3Home"
         else
             echo "Downloading Z3 binary for $OSDisplayName"
             generic_download_file "$urlZ3" z3.zip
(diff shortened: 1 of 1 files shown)


yuleisui:
Thanks for the contribution! I think a better solution is that we can also build LLVM on any arm machine. LLVM also does not have a pre-built version for the mac M1 chip for now. Hopefully, it will have one later.

Would you be able to help improve the pr to add a building for LLVM from the source if a pre-built binary is not available?
https://github.com/SVF-tools/SVF/wiki/Build-LLVM-and--Clang-from-source-code

yuleisui:
BTW: make -j `nproc` will make the processors compete for resources and the building process in fact will become slower on many machines.

ddiepo-pjr:
I like your idea of building LLVM from source when needed, however I don't have a mac m1 to test with.  My specific problem scenario was with aarch64 on Linux, so I tried to keep this change contained to that.

Regarding the "-j " option, perhaps best is to let the user provide this value, and if not provided default to `nproc`.  The general advice is often "number of threads + 1" (https://stackoverflow.com/questions/3025587/whats-best-value-for-make-j), but as with any general optimization the real answer is always "it depends."

ddiepo-pjr:
> 

I updated the branch to allow the job count to be passed in, and I think fixed it so it will work on a mac if nothing is supplied.  It sounds like 'nproc' is not installed by default on macs.

PR #663, opened 2022-03-28
fix compilation warnings

251:
Removes 3 warnings during compilation: `type qualifiers ignored on function return type`.


yuleisui:
Thanks

PR #673, opened 2022-04-04
Svfg rw test

Jeffrey-J-Ma:


Jeffrey-J-Ma:
Test pull request to see if ctest has changed since Luke initially ran it

PR #692, opened 2022-05-03
update workflow

guanqin-123:


guanqin-123:
+ only publish on master
+ only run test for clients' action
+ auto-format

PR #703, opened 2022-05-11
Remove some LLVM deps

mbarbar:
Most of what we discussed, more to come.


yuleisui:
I think you can precomputing a map by iterating the SVFG to get the values of each node. Later you can use this map for your queries

PR #709, opened 2022-05-25
Add support for LLVM-14 and LLVM-15

kagayaku-hou:


include/SVF-FE/LLVMUtil.h
```diff
@@ -104,6 +104,15 @@ inline bool isHeapAllocExtFunViaArg(const SVFFunction* fun)
     return fun && ExtAPI::getExtAPI()->is_arg_alloc(fun);
 }
 
+static inline Type *getPtrElementType(const PointerType* pty) {
+#if (LLVM_VERSION_MAJOR < 14)
+    return getPtrElementType(pty);
```
yuleisui, line 109: This is a recursive call. Do you want `pty>getElementType()'

include/SVF-FE/LLVMUtil.h
```diff
@@ -104,6 +104,15 @@ inline bool isHeapAllocExtFunViaArg(const SVFFunction* fun)
...
+#else
+    assert(!pty->isOpaque() && "Opaque Pointer is used, please recompile the source adding '-Xclang -no-opaque-pointer'");
+    return pty->getNonOpaquePointerElementType();
```
yuleisui, line 112: Do we have any other way to return a correct type rather than an assertion here?

yuleisui:
Thanks a lot for the contribution. Please find my comments.

include/SVF-FE/LLVMUtil.h
kagayaku-hou: With opaque pointers introduced in llvm-14, pointer type does not contain  element type any more. If you need the element type info, you need to compile the source and disable opaque pointer feature. Please read the following link:

https://llvm.org/docs/OpaquePointers.html

yuleisui: I see. Thanks for the explanation.

kagayaku-hou:
What is the status for this commit?

yuleisui:
> What is the status for this commit?

This will be a big change because SVF-example, SVF-Teaching, and Test-suites are using LLVM-13.0. We will need to refresh all these to make sure the patch works for all.

PR #726, opened 2022-06-11
Move some LLVM to SVF-FE

mbarbar:
Largely TBHC things.

### DIFF
--- a/include/SVF-FE/BasicTypes.h
+++ b/include/SVF-FE/BasicTypes.h
@@ -25,6 +25,7 @@
 #include <llvm/Support/SourceMgr.h>
 
 typedef llvm::LLVMContext LLVMContext;
+typedef llvm::GlobalObject GlobalObject;
 typedef llvm::Use Use;
 typedef llvm::ModulePass ModulePass;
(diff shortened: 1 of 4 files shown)


mbarbar:
Hmm then there will be a WPA dependency in SVF-FE. Is that acceptable?

yuleisui:
> 

We aim to remove the dependence of all folders (except SVF-FE and MTA) from LLVM. Will too many LLVM data structures or types used in TBHC?

mbarbar:
Yeah a lot :/ let me see if I can convert the FlowSensitiveTBHC deps to PAG stuff.

TypeBasedHeapCloning.cpp however is probably impossible.

yuleisui:
> Hmm then there will be a WPA dependency in SVF-FE. Is that acceptable?

I think this is fine.

yuleisui:
> TypeBasedHeapCloning.cpp

Why impossible? We allow the WPA dependency in SVF-FE, but not the other way around.

mbarbar:
Ah then that should be possible I think, I'll PR in the coming days.

PR #730, opened 2022-06-15
Attribute and Remove Map in CFLGraph

TalbenXu:


include/CFL/CFLGraphBuilder.h
```diff
@@ -61,11 +79,11 @@ class CFLGraphBuilder
 
     /// Build Bidirectional graph by copying nodes and edges from any graph inherited from GenericGraph
     template<class N, class E>
-    CFLGraph* buildBigraph(GenericGraph<N,E>* graph)
+    CFLGraph* buildBigraph(GenericGraph<N,E>* graph, Kind startKind)
     {
-        CFLGraph *cflGraph = new CFLGraph();
+        CFLGraph *cflGraph = new CFLGraph(startKind);
```
yuleisui, line 84: Make sure every time you create new memory via new, the memory will be released at some point like in descructors.

TalbenXu: Yes, the cflGraph memory object being pass to cflalias, and will be release in cflalias destructor.

lib/Graphs/CFLGraph.cpp
```diff
@@ -158,19 +114,23 @@ struct DOTGraphTraits<CFLGraph*> : public DefaultDOTGraphTraits
...
-            if (i.second == edge->getEdgeKind())
-            {
-                key = i.first;
-            }
-        }
-        rawstr << "label=" << '"' <<key << '"';
-        if( graph->label2SymMap[key] == graph->startSymbol)
-        {
-            rawstr << ',' << "color=red";
-        }
```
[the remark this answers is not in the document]
TalbenXu: Yes, will remove that, and enhance the dot dump layout on future commit.

include/CFL/CFLGraphBuilder.h
yuleisui: The normalised grammar and cflgraph should be released in the destructor of cflsolver.

include/CFL/CFLGraphBuilder.h
```diff
@@ -83,22 +101,22 @@ class CFLGraphBuilder
...
-                    cflGraph->addAttribute(edgeLabel, offset);
+                    addAttribute(edgeLabel, offset);
                     edgeLabel = CFLGrammar::getAttributedKind(offset, edgeLabel);
                     cflGraph->addCFLEdge(cflGraph->getGNode(edge->getSrcID()), cflGraph->getGNode(edge->getDstID()), edgeLabel);
-                    std::string key = cflGraph->sym2LabelMap[edge->getEdgeKind()];
+                    std::string key = kind2LabelMap[edge->getEdgeKind()];
                     key.pop_back();
                     key.pop_back();    // _i standsfor attribute variable should place at last
                     key.append("bar_i");   // for example Gep_i should be Gepbar_i, not Gep_ibar
-                    cflGraph->addCFLEdge(cflGraph->getGNode(edge->getDstID()), cflGraph->getGNode(edge->getSrcID()), CFLGrammar::getAttributedKind(offset, cflGraph->label2SymMap[key]));
```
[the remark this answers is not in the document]
TalbenXu: Yes, "bar_i" is for every Kind with attribute, so currently if that come with two different attribute let's say "i" and "j", then it will just treat as "i" as attribute, but "j" will be just a new part of Nonterminal.
I think I will also need to support multiple attribute, in the case of that kind of grammar input.

PR #736, opened 2022-06-19
provide API for user of class ExtAPI to register coustomized functions

LeoneChen:
How about providing API of class `ExtAPI` to register coustomized functions (including functions that only have declaration and functions that have definition) to `ExtAPI::info`?

After having this, I can pass my customized heap allocator (not listed in `static const ei_pair ei_pairs[]={...}`) to SVF's ExtAPI before SVF analysis

### DIFF
--- a/include/Util/ExtAPI.h
+++ b/include/Util/ExtAPI.h
@@ -107,6 +107,7 @@ private:
 
     // Singleton pattern here to enable instance of PAG can only be created once.
     static ExtAPI* extAPI;
+    std::unordered_set<std::string> definedFuncNames;
 
 public:
 
@@ -130,7 +131,7 @@ public:
             funName = "llvm." + F->getName().split('.').second.split('.').first.str();
         }
         llvm::StringMap<extf_t>::const_iterator it= info.find(funName);
-        if(it == info.end() || !F->isDeclaration())
+        if(it == info.end() || !(F->isDeclaration() or definedFuncNames.count(funName)))
             return EFT_OTHER;
         else
             return it->second;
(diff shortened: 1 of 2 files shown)


yuleisui:
Thanks for pointing this out. We have recently made efforts in refactoring and revising our external call handling. You can have a look at our latest solution to external call summarization. https://github.com/SVF-tools/SVF/pull/738

LeoneChen:
I'm very glad to see this modification. I think it will help me a lot, thanks for your reply! : )

adrianherrera:
Does #738 provide a mechanism for programatically registering a customised function? I cannot see it, although I could have missed it. From my look it seems to just provide mechanisms for parsing the existing JSON.

It would be nice to dynamically (i.e., at runtime) define additional external functions.

yuleisui:
It supports customised external functions in the form of SVFStmt specification (PAG edges), as per include/Util/ExtAPI.json. The specification is then used to build SVFIR for external calls via `SVFIRBuilder::handleExtCall`

adrianherrera:
Thanks Yulei. But unless I am misunderstanding something (which I very well could be!), this API is "read-only"? I.e., there is no way to "add" to the `ExtAPI` programatically? E.g., something like `ExtAPI::getExtAPI()->addFunc`? Once the JSON is loaded (which happens very early on when things are getting loaded and constructed/analyzed), it is mutable (in memory)? What I am asking (and I think this PR was originally suggesting) was an API to add external APIs programatically.

Again, sorry if I am misunderstanding something!

yuleisui:
I see. `ExtAPI::getExtAPI()->addFunc` looks like a good idea to add external APIs programmatically.  Unfortunately, that PR does not have those, and specifying the side-effects is also a bit tricky since that function might need many arguments to specify what PAGEdges are needed. 

I think you can try dynamically updating `ExtAPI.json`, and whenever it is being re-read in memory via `ExtAPI::get_FunJson` it should get the updated side-effects.

adrianherrera:
Thanks Yulei. I have made a small PR to at least allow the user to dynamically specify their own ExtAPI.json files.

PR #737, opened 2022-06-19
Fix CMakeLists.txt

LeoneChen:
[Compile Error] undefined reference to symbol 'pthread_create@@GLIBC_2.2.5'

```shell
/usr/local/bin/ld: ../../../z3.obj/bin/libz3.a(scoped_timer.o): undefined reference to symbol 'pthread_create@@GLIBC_2.2.5'
/usr/local/bin/ld: /lib/x86_64-linux-gnu/libpthread.so.0: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
make[2]: *** [tools/SABER/CMakeFiles/saber.dir/build.make:91: bin/saber] Error 1
```
```shell
/usr/local/bin/ld: ../../../z3.obj/bin/libz3.a(scoped_timer.o): undefined reference to symbol 'pthread_create@@GLIBC_2.2.5'
/usr/local/bin/ld: /lib/x86_64-linux-gnu/libpthread.so.0: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
make[2]: *** [tools/DDA/CMakeFiles/dvf.dir/build.make:91: bin/dvf] Error 1
```

Thus, I add `pthread` to `target_link_libraries` in `tools/DDA/CMakeLists.txt` and `tools/SABER/CMakeLists.txt`, then it works fine. But I don't understand why SVF2.3 with LLVM-12 doesn't have such problem

### DIFF
--- a/tools/DDA/CMakeLists.txt
+++ b/tools/DDA/CMakeLists.txt
@@ -5,7 +5,7 @@ if(DEFINED IN_SOURCE_BUILD)
 else()
     add_executable( dvf dda.cpp )
 
-    target_link_libraries( dvf Svf Cudd ${llvm_libs} )
+    target_link_libraries( dvf Svf Cudd ${llvm_libs} pthread)
 
     set_target_properties( dvf PROPERTIES
                            RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin )
--- a/tools/SABER/CMakeLists.txt
+++ b/tools/SABER/CMakeLists.txt
@@ -5,7 +5,7 @@ if(DEFINED IN_SOURCE_BUILD)
 else()
     add_executable( saber saber.cpp )
 
-    target_link_libraries( saber Svf Cudd ${llvm_libs})
+    target_link_libraries( saber Svf Cudd ${llvm_libs} pthread)
 
     set_target_properties( saber PROPERTIES
                            RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin )


tools/DDA/CMakeLists.txt
```diff
@@ -5,7 +5,7 @@ if(DEFINED IN_SOURCE_BUILD)
 else()
     add_executable( dvf dda.cpp )
 
-    target_link_libraries( dvf Svf Cudd ${llvm_libs} )
+    target_link_libraries( dvf Svf Cudd ${llvm_libs} pthread)
```
yuleisui, line 8: Not sure why include pthread? DDA or Saber never used threads..

LeoneChen: It's not used by DDA and Saber, but by the libz3.

After testing the current master tree, this problem doesn't exist again.

LeoneChen: I currently think the reason is some components of SVF-2.4 use some definitions in `libz3.a(scoped_timer.o)`, which cause linker to link this `scoped_timer.o` object (as linker link project in object-granuality), here is part defined symbol in `scoped_timer.o`, 

image

and then I search them in `saber` binary, it's included

image

However, the current master tree doesn't link `scoped_timer.o`.

image

LeoneChen: Since there is no problem in newest master tree, I will try to use the newest one

yuleisui:
I see, thanks for reporting this.

LeoneChen:
Thanks for reply, I'm going to close this PR

PR #738, opened 2022-06-20
Extern function summaries

shuangxiangkan:
cJSON is an ultra-lightweight, portable, single-file, simple-as-can-be ANSI-C compliant JSON parser, under MIT license, so cJSON.h and cJSON.c are introduced to parse the json file.

In addition, SVF path should be available through $SVF_DIR, otherwise, ExtAPI.json will not be found in function "get_FunJson" of "ExtAPI.cpp".


shuangxiangkan:
modify cJSON.h and cJSON.c，and get_type() in ExtAPI.cpp.

PR #744, opened 2022-06-26
Add priority to user-defined functions and external functions

shuangxiangkan:
priority = 0: Execute user-defined functions
priority = 1:  Execute function specification in ExtAPI.json


yuleisui:
It has conflicts. You will need to pull before make a new request

PR #747, opened 2022-07-01
Add FSTBHC build option

mbarbar:
Sorry for the delay. This adds the ability to do `WITH_FSTBHC=1 ./build.sh`, otherwise no FSTBHC is built in.

Once merged, I'll add this to the wiki.


build.sh
```diff
@@ -237,6 +238,14 @@ export PATH=$LLVM_DIR/bin:$PATH
...
 echo "Z3_DIR=$Z3_DIR"
 
+# Build with or without FSTBHC? Enable when SVF_CTIR is set too.
+if [ -n "$WITH_FSTBHC" ] || [ -n "$SVF_CTIR" ]
+then
+    FSTBHC_OPTION="-DWITH_FSTBHC=ON"
+else
+    FSTBHC_OPTION="-DWITH_FSTBHC=OFF"
+fi
+
```
yuleisui, line 248: Can we merge `SVF_CTIR` and `WITH_FSTBHC`?

yuleisui: SVF_CTIR can be removed? If `WITH_FSTBHC` is on, we will use the ctir?

mbarbar:
Yep fixed

yuleisui:
> Yep fixed

I am thinking that can we remove ctir variables completely and point the ctir clang downloaded to the normal clang location?

yuleisui:
> > Yep fixed
> 
> I am thinking that can we remove ctir variables completely and point the ctir clang downloaded to the normal clang location?

So that we can  simplify the build.sh a bit more

mbarbar:
I suggest not, it's on Clang 9 and I (currently) have no plans to bring it up to date. This is partly why I suggest bumping SVF to 2.5 and dropping TBHC completely.

yuleisui:
> I suggest not, it's on Clang 9 and I (currently) have no plans to bring it up to date. This is partly why I suggest bumping SVF to 2.5 and dropping TBHC completely.

Yes, agree that it is a better idea to drop TBHC in SVF-2.5. If that is the case, could we maybe do it now in this pull request. Our V2.5 will be released soon after some llvm refactoring in the near future.

mbarbar:
All right, let me remove all references.

mbarbar:
Do we want to keep the DCHG?

yuleisui:
> Do we want to keep the DCHG?

It is fine to keep it.

mbarbar:
I think it will become inaccurate or assert things if there is no ctir metadata.

yuleisui:
> I think it will become inaccurate or assert things if there is no ctir metadata.
> 
> Edit: maybe imprecise is a better word than inaccurate.

I see. Would DCHG be used for cpp callgraph resolutions if DIType can be obtained from clang? I think it could be used for better callgraph construction if front end pass us the metainfo?

mbarbar:
Maybe... though it is currently relying on CTIR. If ctir is available I guess it can use it. From memory, DIType is not enough and might (?) be unsound because the frontend will merge types. This is from memory so I could be wrong.

yuleisui:
> Maybe... though it is currently relying on CTIR. If ctir is available I guess it can use it. From memory, DIType is not enough and might (?) be unsound because the frontend will merge types. This is from memory so I could be wrong.

I think we can leave it in 2.5. It should be easy to remove these two files DCHG.h/cpp later since they are not used anywhere else?

mbarbar:
All right yes sure

yuleisui:
Also remember to remove all the macros like ctir in build.sh and other h/cpp

mbarbar:
Yep that's all removed. The only references to ctir are now: DCHG.h, DCHG.cpp, and in CPPUtil (some constants).

```
mbarbar@alder ~/clone/svf-public % ag ctir
include/SVF-FE/CPPUtil.h
106:/// Constants pertaining to CTir, for C and C++.
108:namespace ctir
113:const std::string derefMDName  = "ctir";
116:const std::string vtMDName     = "ctir.vt";
119:const std::string vtInitMDName = "ctir.vt.init";
121:/// Value we expect a ctir-annotated module to have.

include/SVF-FE/DCHG.h
409:        MDNode *md = cs.getInstruction()->getMetadata(cppUtil::ctir::derefMDName);

lib/SVF-FE/DCHG.cpp
173:        if (gv->hasMetadata(cppUtil::ctir::vtMDName) && gv->getNumOperands() > 0)
175:            DIType *type = SVFUtil::dyn_cast<DIType>(gv->getMetadata(cppUtil::ctir::vtMDName));
176:            assert(type && "DCHG::buildVTables: bad metadata for ctir.vt");
```

CMakeLists.txt
```diff
@@ -80,6 +79,11 @@ if(EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/Test-Suite")
     include(CTest)
 endif()
 
+option(WITH_FSTBHC "Build SVF with FSTBHC enabled" OFF)
```
yuleisui, line 82: This can also be removed

build.sh
```diff
@@ -217,20 +218,25 @@ fi
 ########
```
yuleisui, line 218: This line onwards to line 240 can be removed

CMakeLists.txt
mbarbar: Yes this is removed
```
mbarbar@alder ~/clone/svf-public % ag fstbhc
mbarbar@alder ~/clone/svf-public % ag tbhc
include/SVF-FE/DCHG.h
300:        // For TBHC this is conservative because the union type is lower in the DCHG
mbarbar@alder ~/clone/svf-public %
```

build.sh
mbarbar: Oh sorry! Didn't not push properly...

yuleisui:
Well done and will merge now

yuleisui:
@mbarbar the test validations are gone. No success or failure for each test case (e.g., andersen's analysis) is printed as before https://github.com/SVF-tools/SVF/runs/7126678076?check_suite_focus=true

yuleisui:
Just fixed https://github.com/SVF-tools/SVF/commit/925fb44a55eacd38f2abff5298c2d412985e4606

Issue #760, opened 2022-07-12
Analyzing global variable


tandf:
Hi,

I'm analyzing this program using VFG
```c++
#include <string>

void sinkfunc(std::string sink_str)
{
    std::string sink = sink_str;
}

const std::string str = "foobar";

int main()
{
    sinkfunc(str);
    return 0;
}
```
The `str` global variable part in the VFG graph is quite different from its local variable counterpart below. 

```c++
#include <string>

void sinkfunc(std::string sink_str)
{
    std::string sink = sink_str;
}

int main()
{
    const std::string str = "foobar";
    sinkfunc(str);
    return 0;
}

```

Here are the graphs for the two programs: [svfg.zip](https://github.com/SVF-tools/SVF/files/9088343/svfg.zip)
I think in the global variable version, there is a missing link from *StoreVFGNode ID: 74* to *LoadVFGNode ID: 57*, which corresponds to the edge from *StoreVFGNode ID: 59* to *LoadVFGNode ID: 47* in the local variable version.

Here is my guess of the reason: In the global variable version, the store node ID 74 has in node *FormalINSVFGNode ID 112* with CHI(MR_2V_1), and the load node ID 57 has in node *FormalINSVFGNode ID 109* with CHI(MR_2V_1). The two *FormalINSVFGNode* should be used to connect the load and the store node, but somehow they are not connected.

Could you tell me if VFG is behaving correctly, and if not, how can I fix it? Thanks a lot.

yuleisui:
Can you send us your bcs and indicate you commend liine options?

tandf:
Here are the bcs: [bc.zip](https://github.com/SVF-tools/SVF/files/9088432/bc.zip)
The command line options are `-stat=false -model-consts=true -model-arrays=true --dwarn=false --dump-vfg`

yuleisui:
From my understanding, the global value-flows should be connected. It is a bit messy to examine for this one. Would you be able to try a C code in the same pattern first?

yuleisui:
[svfg_final.dot.zip](https://github.com/SVF-tools/SVF/files/9088729/svfg_final.dot.zip)
In case, I have uploaded the svfg dumpped on my local machine for your comparison.

yuleisui:
It works fine for C which has the same pattern.  LoadVFGNode ID 26 is connected from FormalINSVFGNode from the StoreVFGNode ID 35. Will have to take a look at whether it is because of some string APIs in C++
[...]

PR #774, opened 2022-07-15
llvm.global_ctors/dtors

tandf:
Sort llvm_global_ctors functions according to their priority. Also support destroy functions. See [LLVM Language Reference Manual](https://llvm.org/docs/LangRef.html#the-llvm-global-ctors-global-variable)

### DIFF
--- a/lib/SVF-FE/LLVMModule.cpp
+++ b/lib/SVF-FE/LLVMModule.cpp
@@ -309,8 +309,8 @@ void LLVMModuleSet::addSVFMain()
             }
         }
     }
-    if(orgMain && getModuleNum() > 0 && init_funcs.size() > 0)
-    {
+    if (orgMain && getModuleNum() > 0 && init_funcs.size() > 0 ||
+        destroy_funcs.size() > 0) {
         assert(mainMod && "Module with main function not found.");
         Module & M = *mainMod;
         // char **
--- a/lib/SVF-FE/LLVMModule.cpp
+++ b/lib/SVF-FE/LLVMModule.cpp
@@ -309,8 +309,8 @@ void LLVMModuleSet::addSVFMain()
             }
         }
     }
-    if (orgMain && getModuleNum() > 0 && init_funcs.size() > 0 ||
-        destroy_funcs.size() > 0) {
+    if (orgMain && getModuleNum() > 0 &&
+        (init_funcs.size() > 0 || destroy_funcs.size() > 0)) {
         assert(mainMod && "Module with main function not found.");
         Module & M = *mainMod;
         // char **
--- a/lib/SVF-FE/LLVMModule.cpp
+++ b/lib/SVF-FE/LLVMModule.cpp
@@ -39,19 +39,31 @@ using namespace std;
 using namespace SVF;
 
 /*
-  svf.main() is used to model the real entry point of a C++ program,
-  which initializes all global C++ objects and then call main().
+  svf.main() is used to model the real entry point of a C++ program, which
+  initializes all global C++ objects and then call main().
+
+  LLVM may generate two global arrays @llvm.global_ctors and @llvm.global_dtors
+  that contain constructor and destructor functions for global variables. They
+  are not called explicitly, so we have to add them in the svf.main function.
+  The order to call these constructor and desctructor functions are also
+  specified in the global arrays.
+  
+  Related part in LLVM language reference:
+  https://llvm.org/docs/LangRef.html#the-llvm-global-ctors-global-variable
+
   For example, given a "int main(int argc, char * argv[])", the corresponding
   svf.main will be generated as follows:
     define void @svf.main(i32, i8**, i8**) {
       entry:
-        call void @_GLOBAL__sub_I_cast.cpp()
-        call void @_GLOBAL__sub_I_1.cpp()
-        call void @_GLOBAL__sub_I_2.cpp()
+        call void @ctor1()
+        call void @ctor2()
         %3 = call i32 @main(i32 %0, i8** %1)
+        call void @dtor1()
+        call void @dtor2()
         ret void
     }
- */
+*/
+
 #define SVF_MAIN_FUNC_NAME           "svf.main"
 #define SVF_GLOBAL_CTORS             "llvm.global_ctors"
 #define SVF_GLOBAL_DTORS             "llvm.global_dtors"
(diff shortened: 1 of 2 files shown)


yuleisui:
Can this patch pass your tests in https://github.com/SVF-tools/SVF/issues/760 now?

tandf:
Yes, it can pass my test now.

yuleisui:
Can you indicate which option you have used to pass your test, I assume that `-svf-main` is also turned on?

Do you think this option should always be turned on if analyzing cpp? or any code could be changed to call `addSVFMain ` if we have cpp constructors?

tandf:
I use `-model-array -model-consts -svf-main` to run the test. I also merge this pull request to my own analysis, and ran on multiple program that I collected from the wild, and I didn't find anything problem that I'm sure is caused by this change. Nevertheless, it'll be great if you can help me review and test it.

As for your question that whether this option should always be turned on, I think it should. My understanding of `-svf-main` is that, it creates a new program entry that contains init functions (and destroy functions) created by llvm but are not explicitly called. If my understanding is correct, I think it's necessary to enable this option by default, since it reflects the semantics of the ir. Also, as I mentioned, I tested on multiple cpp IR files collected from the wild with `-svf-main`, and the results seem to be correct so far.

yuleisui:
Can you try to turn off `-model-array -model-consts` and only enable `-svf-main`? It should also pass your test.

tandf:
Yes it works with only `-svf-main`.

For the test, I run with svf-ex. I run on out.ll as attached in #775. I manually added the following content to test destroy functions. After disassembling the out.svf.bc file I can see the correct @svf.main() function injected
```
@llvm.global_dtors = appending global [2 x { i32, void ()*, i8* }] [{ i32, void ()*, i8* } { i32 6234, void ()* @test1, i8* null }, { i32, void ()*, i8* } { i32 3213, void ()* @test2, i8* null }]
define internal void @test1() #0 section ".text.startup" {
  ret void
}
define internal void @test2() #0 section ".text.startup" {
  ret void
}
```

yuleisui:
This pr is important. To make it easier to maintain and extensible later, would you be able to add some comments for the code? Maybe, you can also put a small example in your code comments.

lib/SVF-FE/LLVMModule.cpp
```diff
@@ -227,18 +228,79 @@ void LLVMModuleSet::initialize()
...
+            queue;
+    std::vector<llvm::Function *> result;
+
+    llvm::ConstantArray *globalFuncArray =
+            llvm::dyn_cast<llvm::ConstantArray>(global->getInitializer());
+    if (globalFuncArray) {
+        for (unsigned int i = 0; i < globalFuncArray->getNumOperands(); ++i) {
+            llvm::ConstantStruct *globalFuncItem =
+                    llvm::dyn_cast<llvm::ConstantStruct>(
+                            globalFuncArray->getOperand(i));
```
yuleisui, line 259: could you use SVF::dyn_cast and put this dyn_cast into the `if` branch?

if(ConstantStruct *globalFuncItem = SVF::dyn_cast<llvm::ConstantStruct>(globalFuncArray->getOperand(I))){
...
}

lib/SVF-FE/LLVMModule.cpp
```diff
@@ -227,18 +228,79 @@ void LLVMModuleSet::initialize()
...
+    };
+
+    std::priority_queue<LLVMGlobalFunction, std::vector<LLVMGlobalFunction>,
+                        greater<LLVMGlobalFunction>>
+            queue;
+    std::vector<llvm::Function *> result;
+
+    llvm::ConstantArray *globalFuncArray =
+            llvm::dyn_cast<llvm::ConstantArray>(global->getInitializer());
+    if (globalFuncArray) {
```
yuleisui, line 255: put the dyn_cast into if branch and use SVF::dyn_cast.

if(ConstantArray *globalFuncArray = SVF::dyn_cast<llvm::ConstantArray>(global->getInitializer()){
}

lib/SVF-FE/LLVMModule.cpp
```diff
@@ -227,18 +228,79 @@ void LLVMModuleSet::initialize()
...
+std::vector<llvm::Function *> LLVMModuleSet::getLLVMGlobalFunctions(
+        llvm::GlobalVariable *global)
+{
+    class LLVMGlobalFunction {
+    // llvm.global_ctors and llvm.global_dtors
+    public:
+        unsigned priority;
+        llvm::Function *func;
+        LLVMGlobalFunction(){};
+        LLVMGlobalFunction(unsigned _priority, llvm::Function *_func)
```
yuleisui, line 240: unsigned=>u32_t
llvm::Function => Function

lib/SVF-FE/LLVMModule.cpp
```diff
@@ -227,18 +228,79 @@ void LLVMModuleSet::initialize()
     }
 }
 
+std::vector<llvm::Function *> LLVMModuleSet::getLLVMGlobalFunctions(
+        llvm::GlobalVariable *global)
+{
+    class LLVMGlobalFunction {
+    // llvm.global_ctors and llvm.global_dtors
+    public:
+        unsigned priority;
```
yuleisui, line 237: unsigned => u32_t

lib/SVF-FE/LLVMModule.cpp
```diff
@@ -227,18 +228,79 @@ void LLVMModuleSet::initialize()
...
+                    llvm::dyn_cast<llvm::ConstantStruct>(
+                            globalFuncArray->getOperand(i));
+            if (globalFuncItem) {
+                ConstantInt *priority = llvm::dyn_cast<llvm::ConstantInt>(
+                        globalFuncItem->getOperand(0));
+                Function *func = llvm::dyn_cast<llvm::Function>(
+                        globalFuncItem->getOperand(1));
+
+                if (priority && func) {
+                    queue.push(LLVMGlobalFunction((unsigned)priority
```
yuleisui, line 267: (unsigned)priority => priority.getZExtValue()

lib/SVF-FE/LLVMModule.cpp
```diff
@@ -227,18 +228,79 @@ void LLVMModuleSet::initialize()
...
+
+    llvm::ConstantArray *globalFuncArray =
+            llvm::dyn_cast<llvm::ConstantArray>(global->getInitializer());
+    if (globalFuncArray) {
+        for (unsigned int i = 0; i < globalFuncArray->getNumOperands(); ++i) {
+            llvm::ConstantStruct *globalFuncItem =
+                    llvm::dyn_cast<llvm::ConstantStruct>(
+                            globalFuncArray->getOperand(i));
+            if (globalFuncItem) {
+                ConstantInt *priority = llvm::dyn_cast<llvm::ConstantInt>(
```
yuleisui, line 261: ConstantInt *priority => const ConstantInt *priority

lib/SVF-FE/LLVMModule.cpp
```diff
@@ -227,18 +228,79 @@ void LLVMModuleSet::initialize()
...
 }
 
+std::vector<llvm::Function *> LLVMModuleSet::getLLVMGlobalFunctions(
+        llvm::GlobalVariable *global)
+{
+    class LLVMGlobalFunction {
+    // llvm.global_ctors and llvm.global_dtors
+    public:
+        unsigned priority;
+        llvm::Function *func;
```
yuleisui, line 238: llvm::Function *func; => const Function* func;

lib/SVF-FE/LLVMModule.cpp
```diff
@@ -227,18 +228,79 @@ void LLVMModuleSet::initialize()
...
+        LLVMGlobalFunction(){};
+        LLVMGlobalFunction(unsigned _priority, llvm::Function *_func)
+              : priority(_priority), func(_func){};
+        bool operator>(const LLVMGlobalFunction &other) const
+        {
+            return priority > other.priority;
```
yuleisui, line 244: Do we change to priority > other.priority && _func > other._func if two globalfunctions' priority is the same?

lib/SVF-FE/LLVMModule.cpp
```diff
@@ -227,18 +228,79 @@ void LLVMModuleSet::initialize()
...
-        {
-            if(func.getName().startswith(SVF_GLOBAL_SUB_I_XXX))
-                init_funcs.push_back(&func);
-            if(func.getName().equals(SVF_MAIN_FUNC_NAME))
+    for (Module &mod : modules) {
+        // Collect ctor and dtor functions
+        for (Module::global_iterator it = mod.global_begin(),
+                                     eit = mod.global_end();
+             it != eit; ++it) {
+            GlobalVariable *global = &*it;
```
yuleisui, line 293: can we add a const qualifier for GlobalVariable and the below lines if we don't change their attributes?

lib/SVF-FE/LLVMModule.cpp
tandf: done

tandf:
Hi @yuleisui. In the new commits, I update the pr based on your comments, and also add some comments to explain the logic. Please feel free to tell me if there is anything else I can do.

yuleisui:
Thanks and merged

PR #787, opened 2022-07-22
replace cudd with z3

jiawei-95:


jumormt:
(Do not merge) still WIP

PR #793, opened 2022-07-25
get correct types for values used in memcpy

Flokkka:
In the following two cases, SVF fails to determine the correct type of a value used in a memcpy:

1. 
```llvm
%call   = malloc 80
%0      = bitcast i8* %call to %struct.x264_t*
%1      = bitcast %struct.x264_param_t* %param to i8*
call void memcpy(%call, %1, 80)
```

2.
```llvm
%0 = bitcast %struct.x264_param_t* %param to i8*, !dbg !3505
%2 = bitcast %struct.x264_t** %arrayidx to i8**
%3 = load i8*, i8** %2
call void @memcpy(%0, %3, 704)
```

This leads to underapproximations in the pointer analysis.

### DIFF
--- a/lib/SVF-FE/SVFIRBuilder.cpp
+++ b/lib/SVF-FE/SVFIRBuilder.cpp
@@ -1008,6 +1008,21 @@ const Value* SVFIRBuilder::getBaseValueForExtArg(const Value* V)
         if(totalidx == 0 && !SVFUtil::isa<StructType>(value->getType()))
             value = gep->getPointerOperand();
     }
+
+    LLVMContext &cxt = LLVMModuleSet::getLLVMModuleSet()->getContext();
+    if (value->getType() == PointerType::getInt8PtrTy(cxt)) {
+        if (const CallBase* cb = SVFUtil::dyn_cast<CallBase>(value)) {
+            for (const User* user: cb->users()) {
+                if (const BitCastInst* bitCast = SVFUtil::dyn_cast<BitCastInst>(user))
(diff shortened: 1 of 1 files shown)


yuleisui:
Thanks for bringing this up. Could you also make code comments by explaining your code and including these two tiny examples.

lib/SVF-FE/SVFIRBuilder.cpp
```diff
@@ -1008,6 +1008,37 @@ const Value* SVFIRBuilder::getBaseValueForExtArg(const Value* V)
...
+    // %3 = load i8*, i8** %2
+    // call void @memcpy(%0, %3, 80)
+    LLVMContext &cxt = LLVMModuleSet::getLLVMModuleSet()->getContext();
+    if (value->getType() == PointerType::getInt8PtrTy(cxt)) {
+        // (1)
+        if (const CallBase* cb = SVFUtil::dyn_cast<CallBase>(value)) {
+            for (const User* user: cb->users()) {
+                if (const BitCastInst* bitCast = SVFUtil::dyn_cast<BitCastInst>(user))
+                    return bitCast;
+            }
```
yuleisui, line 1033: We should make sure the bitcast is the unique cast, otherwise, we may get the type from another cast not flowing into this memcpy?

Is the following what you want?
```````
if(SVFUtil::isHeapAllocExtCallViaRet(cb)){
  if(const Value* bitCast = getUniqueUseViaCastInst(cb))
     return bitCast;
}
````

PR #799, opened 2022-07-29
Remove redundant collapsePWCNode in WPASolver

kisslune:


yuleisui:
Also need to remove  `static SCCEdgeFlag sccEdgeFlag;` and `mergePWC()`

PR #811, opened 2022-08-05
cpp_test

TalbenXu:


lib/SVF-FE/SVFIRBuilder.cpp
```diff
@@ -1404,6 +1404,20 @@ void SVFIRBuilder::handleExtCall(CallSite cs, const SVFFunction *callee)
                         }
                         break;
                     }
+                    case ExtAPI::EXT_GEP:
+                    {
+                        NodeID vnS = parseNode(args[0], cs, inst);
+                        LocationSet l1;
+                        NodeID vnV = parseNode(args[1], cs, inst);
+                        LocationSet l2;
+                        NodeID vnD = parseNode(args[2], cs, inst);
```
yuleisui, line 1413: We should add a line to parse the location set:
```
u32_t offset = parseNode(args[3], cs, inst);
if (vnD && vnV && vnS && offset)
{
   LocationSet l(offset);
   addNormalGepEdge(vnS, vnV, l);
   addNormalGepEdge(vnV, vnD, l);
}
```

include/Util/ExtAPI.json
```diff
@@ -3270,5 +3270,10 @@
         "type": "CPP_EFT_DYNAMIC_CAST",
         "overwrite_app_function": 0,
         "copy": ["A0", "L"]
+    },
+    "_ZNSt5arrayIPK1ALm2EE4backEv":{
+        "type": "CPP_EFT_DYNAMIC_CAST",
+        "overwrite_app_function": 1,
+        "gep": ["A0","V","L"]
```
yuleisui, line 3277: Should have an additional argument for offset:

"gep": ["A0","V","L", "0"]

PR #821, opened 2022-08-08
Make ExtAPI extensible at run time, after parsing

adriaanjacobs:
In our own analyses, we discover that some user-defined functions implement memory allocation functionality. To make the points-to results more precise in this case, we inform SVF about this.

We used to have a much more invasive patch to SVF for this, but since the introduction of the JSON ExtAPI format, this small patch now suffices.

There's many other ways to go about this change. In fact, SVF already supports this in a way by just parsing & generating a new ExtAPI up front. However, that requires duplicating quite a bit of parsing/JSON format code that SVF already contains anyway. This patch adds a small `add_entry` function to ExtAPI that modifies the in-memory cJSON object directly. 

Let me know if this is not a direction you would like to go with ExtAPI, or if you have any comments. Happy to do this in a different way if need be.

PS: there's some stylistic/compiler warning changes I made too in various other parts. Let me know if you'd prefer this in another patch, or not at all.

### DIFF
--- a/include/Util/ExtAPI.h
+++ b/include/Util/ExtAPI.h
@@ -231,8 +231,12 @@ public:
 
     static void destory();
 
+    void add_entry(const char* funName, extType type, bool overwrite_app_function);
+
     // Get the corresponding name in ext_t, e.g. "EXT_ADDR" in {"addr", EXT_ADDR},
-    extf_t get_opName(std::string s);
+    extf_t get_opName(const std::string& s);
+    // opposite for extType
+    const std::string& extType_toString(extType type);
 
     // Return the extf_t of (F).
     // Get external function name, e.g "memcpy"
@@ -246,6 +250,7 @@ public:
 
     // Get property of the operation, e.g. "EFT_A1R_A0R"
     extType get_type(const SVF::SVFFunction *callee);
+    extType get_type(const std::string& funName);
 
     // Get priority of he function, return value
     // 0: Apply user-defined functions
--- a/include/Util/ExtAPI.h
+++ b/include/Util/ExtAPI.h
@@ -231,6 +231,7 @@ public:
 
     static void destory();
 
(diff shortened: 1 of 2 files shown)


yuleisui:
Thanks for adding this. The ExtAPI part looks good to me. Would be better if you could add some comments for the `add_entry` method. 
For the warnings of `maybe unused` variables, they are indeed used later. I am not sure why there are warnings and would be good to separate them into a different patch.

lib/WPA/VersionedFlowSensitiveStat.cpp
```diff
@@ -46,8 +46,8 @@ void VersionedFlowSensitiveStat::performStat()
     versionStat();
     ptsSizeStat();
 
-    u32_t fiObjNumber = 0;
-    u32_t fsObjNumber = 0;
+    [[maybe_unused]] u32_t fiObjNumber = 0;
```
yuleisui, line 49: Do not understand why this variable maybe unused. It is used at line 62.

adriaanjacobs: It's incremented there, but the actual value is never stored or used anywhere else. I'm compiling with clang trunk, which probably recently got a bit smarter about this kind of construct. Perhaps this used to be a statistic somewhere but is now never saved?

adriaanjacobs:
I'm separating the [[maybe_unused]] stuff into a different PR right now, but AFAIK github doesn't do dependent PRs, so we'll see how that goes.

lib/WPA/VersionedFlowSensitiveStat.cpp
yuleisui: I see. It would be good to store them in the stat map for later printing

adriaanjacobs: I created #822 to separate this out, if that's alright. I think that then has to be merged first (not sure how dependent PRs work in GitHub)

PR #822, opened 2022-08-08
Add some [[maybe_unused]] annotations to help suppress warnings

adriaanjacobs:
As per the discussion in #821, some of the variables here are supposedly not used, causing `-Werror` compilation failures. 

I'm compiling with a recent version of clang, which can probably detect that this variables are initialized, then incremented a few times, but their eventual value is never used. This is basically a more complicated case of a write-only variable, which is also commonly warned on.

We could also remove these counters as a whole, did they used to be logged somewhere perhaps?

### DIFF
--- a/lib/WPA/FlowSensitiveStat.cpp
+++ b/lib/WPA/FlowSensitiveStat.cpp
@@ -127,12 +127,10 @@ void FlowSensitiveStat::performStat()
 
     unsigned numOfCopy = 0;
     unsigned numOfStore = 0;
-    [[maybe_unused]] unsigned numOfNode = 0;
     SVFG::iterator svfgNodeIt = fspta->svfg->begin();
     SVFG::iterator svfgNodeEit = fspta->svfg->end();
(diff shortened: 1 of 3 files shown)


yuleisui:
Rather than removing I think it would be good to store them in PTNumStatMap so that we can print them out later.

adriaanjacobs:
I'm not sure about the remaining `numOfNode` stats. The total number of SVFG nodes is already kept track of by SVFGStat, so I imagine we can just delete those here? Should we also delete the counts for `copy` and `store`? Those also seem to be SVFG related, instead of FlowSensitive/VersionedFlowSensitive related.

yuleisui:
> I'm not sure about the remaining `numOfNode` stats. The total number of SVFG nodes is already kept track of by SVFGStat, so I imagine we can just delete those here? Should we also delete the counts for `copy` and `store`? Those also seem to be SVFG related, instead of FlowSensitive/VersionedFlowSensitive related.


Yes, please delete them and make sure all the attributes [[maybe_unused]] be removed and settled.

PR #829, opened 2022-08-17
Code clean

kisslune:
1. remove duplicate def of mergeNodeToRep in AndersenWaveDiff
2. remove redundant def of collapsePWCNode in WPASolver


yuleisui:
Notes: Also need to remove `static SCCEdgeFlag sccEdgeFlag`; and `mergePWC()`

PR #835, opened 2022-08-20
Fix ff-eq-base diff-test

Jeffrey-J-Ma:


yuleisui:
Please also remember to update SVF-Tools/Test-Suite

yuleisui:
Still failed when running on new test suite
https://github.com/SVF-tools/SVF/runs/7932534650?check_suite_focus=true

PR #838, opened 2022-08-22
WAVE will always collapse PWCs

kisslune:


lib/WPA/Andersen.cpp
```diff
@@ -187,7 +187,6 @@ void Andersen::initialize()
 {
     resetData();
     setDiffOpt(Options::PtsDiff);
```
yuleisui, line 189: Should remove this `setDiffOpt` too.

kisslune: Diff pts is now always triggered

PR #854, opened 2022-08-27
Improve meld labeling performance

mbarbar:
This improves the performance of meld labeling, namely for the printf BC, but in general too.

Problems with printf:
* There are a lot of objects compared to what the BC is actually doing (see low avg top-level points-to set size vs. number of objects). There is a loop performing `getPts` per Store node per object.
* The SVFG grew and there is a loop iterating number-of-nodes-in-svfg times _per_ object.
* After SVF-2.4, we introduced an overhaul of meld labeling that is generally better but for the reasons mentioned above, it is not better for printf. With printf's small points-to sets (but large SVFG, many objects), the extremely naive labeling of 2.4 worked better.

Stats:
SVF-2.4:
```
OLD 1 THREAD
=== printf ===
AvgTopLvlPtsSize    11.4371
meldLabelingTime    1.421
TotalTime           10.624
=== dhcpcd ===
meldLabelingTime    0.619
TotalTime           4.161
AvgTopLvlPtsSize    114.406
=== gawk ===
AvgTopLvlPtsSize    259.449
TotalTime           237.97
meldLabelingTime    150.421
=== bash ===
AvgTopLvlPtsSize    168.033
meldLabelingTime    4.941
TotalTime           13.799
=== lynx ===
meldLabelingTime    230.548
AvgTopLvlPtsSize    168.85
TotalTime           423.679
```

Master before this PR:
```
ORIGIN 1 THREAD
=== printf ===
AvgTopLvlPtsSize    11.6907
meldLabelingTime    515.098
TotalTime           526.648
=== dhcpcd ===
AvgTopLvlPtsSize    110.884
meldLabelingTime    0.882
TotalTime           4.135
=== gawk ===
AvgTopLvlPtsSize    250.697
meldLabelingTime    16.687
TotalTime           74.879
=== bash ===
meldLabelingTime    12.575
TotalTime           18.741
AvgTopLvlPtsSize    183.723
=== lynx ===
meldLabelingTime    37.887
AvgTopLvlPtsSize    166.187
TotalTime           192.218
ORIGIN 4 THREADS (with -clock-type=wall)
=== printf ===
meldLabelingTime    152.631
AvgTopLvlPtsSize    11.6907
TotalTime           163.179
=== dhcpcd ===
AvgTopLvlPtsSize    110.884
meldLabelingTime    0.299385
TotalTime           3.92757
=== gawk ===
AvgTopLvlPtsSize    250.697
meldLabelingTime    6.02532
TotalTime           69.2984
=== bash ===
TotalTime           10.289
AvgTopLvlPtsSize    183.723
meldLabelingTime    4.44136
=== lynx ===
AvgTopLvlPtsSize    166.187
meldLabelingTime    14.5726
TotalTime           168.198
```

This PR:
```
NEW 1 THREAD
=== printf ===
AvgTopLvlPtsSize    11.6907
meldLabelingTime    122.538
TotalTime           129.038
=== dhcpcd ===
AvgTopLvlPtsSize    110.884
meldLabelingTime    0.909
TotalTime           4.366
=== gawk ===
AvgTopLvlPtsSize    250.697
meldLabelingTime    17.009
TotalTime           72.897
=== bash ===
AvgTopLvlPtsSize    183.723
meldLabelingTime    12.854
TotalTime           18.508
=== lynx ===
meldLabelingTime    35.552
AvgTopLvlPtsSize    166.187
TotalTime           159.868

NEW 4 THREADS (with -clock-type=wall)
=== printf ===
AvgTopLvlPtsSize    11.6907
meldLabelingTime    46.6277
TotalTime           53.2753
=== dhcpcd ===
meldLabelingTime    0.292291
AvgTopLvlPtsSize    110.884
TotalTime           3.72266
=== gawk ===
AvgTopLvlPtsSize    250.697
meldLabelingTime    6.08737
TotalTime           61.8428
=== bash ===
meldLabelingTime    4.54459
TotalTime           10.26
AvgTopLvlPtsSize    183.723
=== lynx ===
AvgTopLvlPtsSize    166.187
meldLabelingTime    13.743
TotalTime           141.794
```

Sorry I should have run them all with `wall` clock.

I don't think we will get 2.4 numbers as easily for printf, but for now I think it should be good enough. 10s vs 50s looks bad but I think it's okay. If we keep getting programs with the same object/SVFG/PTS patterns, we might want to revisit.

Remaining bottlenecks:
* There is a hot loop calling `test` a lot. In my head, this might be a bit tricky to improve without causing memory problems.
* `VFS::detectSCCs` is performed per object, and improving it will improve things overall. I spent a bit of time on it when I wrote it, but there should be a few remaining micro things.

Note: this PR vs master gives the same AvgTop, but vs. 2.4, it doesn't. Some changes must have happened along the way (as far as I recall, VFS was always in line with FS stats, and FS and VFS weren't really changed so the change must have been in something like PAG, SVFG, etc.). Also I think it's hard to compare numbers when these AvgTop differ a lot like in bash (over 10).

Finally, the command that should be used is:
```
./bin/wpa -vfspta -versioning-threads=4 -clock-type=wall -pt-type=cbv -node-alloc-strat=dense bc
```
`versioning-threads` should be set to how many threads for versioning. `cbv` is important because of the `test` calls. It should also improve Andersen's/FS solving. `dense` similarly. From the profile, it seems `sbv` was used, and I'm not sure how many threads were used.

### DIFF
--- a/lib/WPA/VersionedFlowSensitive.cpp
+++ b/lib/WPA/VersionedFlowSensitive.cpp
@@ -123,15 +123,29 @@ void VersionedFlowSensitive::meldLabel(void)
 
     assert(Options::VersioningThreads > 0 && "VFS::meldLabel: number of versioning threads must be > 0!");
 
-    // Nodes which have at least one object on them given a prelabel.
-    std::vector<const SVFGNode *> prelabeledNodes;
+    // Nodes which have at least one object on them given a prelabel + the Andersen's points-to
+    // set of interest so we don't keep calling getPts. For Store nodes, we'll fill that in, for
+    // MR nodes, we won't as its getPointsTo is cheap.
+    // TODO: preferably we cache both for ease and to avoid the dyn_cast/isa, but Andersen's points-to
+    // sets are PointsTo and MR's sets are NodeBS, which are incompatible types. Maybe when we can
+    // use std::option.
+    std::vector<std::pair<const SVFGNode *, const PointsTo *>> prelabeledNodes;
     // Fast query for the above.
     std::vector<bool> isPrelabeled(svfg->getTotalNodeNum(), false);
     while (!vWorklist.empty())
     {
         const NodeID n = vWorklist.pop();
-        prelabeledNodes.push_back(svfg->getSVFGNode(n));
         isPrelabeled[n] = true;
+
+        const SVFGNode *sn = svfg->getSVFGNode(n);
+        const PointsTo *nPts = nullptr;
+        if (const StoreSVFGNode *store = SVFUtil::dyn_cast<StoreSVFGNode>(sn))
+        {
+            const NodeID p = store->getPAGDstNodeID();
+            nPts = &(this->ander->getPts(p));
+        }
+
+        prelabeledNodes.push_back(std::make_pair(sn, nPts));
     }
 
     // Delta, delta source, store, and load nodes, which require versions during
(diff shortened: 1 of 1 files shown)


yuleisui:
Thanks and merged.

PR #876, opened 2022-10-03
refactor loop, loopinfo, llvm::user

JasonZhongZexin:


lib/SVF-FE/LLVMModule.cpp
```diff
@@ -200,6 +201,14 @@ void LLVMModuleSet::build()
                         svffun->getPostDomTreeMap()[bb] = Set<const BasicBlock *>();
                     }
                 }
+                const Loop *loop = loopInfo.getLoopFor(bb);
```
yuleisui, line 204: move this line to the following if branch condition
if(const Loop *loop = loopInfo.getLoopFor(bb)){

}

include/Util/BasicTypes.h
```diff
@@ -251,6 +266,38 @@ class SVFFunction : public SVFValue
         return this->exitBB;
     }
 
+    void getExitBlocksOfLoop(const BasicBlock* bb, Set<const BasicBlock*>& exitbbs) const
+    {
+        if (hasLoopInfo(bb))
+        {
+            const std::vector<const BasicBlock*> blocks = getLoopInfo(bb);
```
yuleisui, line 273: A copy construction should be avoided.

const std::vector<const BasicBlock*>& blocks = getLoopInfo(bb);

include/Util/BasicTypes.h
```diff
@@ -251,6 +266,38 @@ class SVFFunction : public SVFValue
...
+            if (!blocks.empty())
```
yuleisui, line 274: blocks should never be empty? Because you always push something into the map `addToBB2LoopMap `?

assert(!blocks.empty() && "no available loop info?")

include/Util/BasicTypes.h
```diff
@@ -251,6 +266,38 @@ class SVFFunction : public SVFValue
...
+                    }
+                }
+            }
+        }
+    }
+
+    bool isLoopHeader(const BasicBlock* bb) const
+    {
+        if (hasLoopInfo(bb)){
+            const std::vector<const BasicBlock*> blocks = getLoopInfo(bb);
```
yuleisui, line 292: avoid copy construction

include/Util/BasicTypes.h
```diff
@@ -251,6 +266,38 @@ class SVFFunction : public SVFValue
...
+            if (!blocks.empty())
```
yuleisui, line 293: assert

PR #879, opened 2022-10-08
fix the command of popen()

2truth:
error message - sh: 1: /path_to_SVF_DIR: Permission denied
cause - popen("$SVF_DIR", "r")
path - jsonFilePath = getJsonFile("$SVF_DIR"); -> GetStdoutFromCommand() -> popen()

### DIFF
--- a/lib/Util/ExtAPI.cpp
+++ b/lib/Util/ExtAPI.cpp
@@ -149,7 +149,7 @@ ExtAPI *ExtAPI::getExtAPI(const std::string &path)
             return extOp;
         }
 
-        jsonFilePath = getJsonFile("$SVF_DIR");
+        jsonFilePath = getJsonFile("echo $SVF_DIR");
         if (!stat(jsonFilePath.c_str(), &statbuf))
         {
             root = parseJson(jsonFilePath, statbuf.st_size);


yuleisui:
Why is the line you added not covered by our tests?

https://github.com/SVF-tools/SVF/pull/879/files#annotation_4858369545

2truth:
If SVF is built from source code, branch01 is always taken and ends the call. 
When SVF is installed by npm, branch01 is skipped cause the PROJECT_PATH is set by your building envrionment  "/home/runner/work/SVF/SVF" and users don't have the path. Then assign02, the fix, will be covered.

* assign01 - https://github.com/SVF-tools/SVF/blob/6e0d7f31b230cb90e16ec45277abca9449609f35/lib/Util/ExtAPI.cpp#L145
* branch01 - https://github.com/SVF-tools/SVF/blob/6e0d7f31b230cb90e16ec45277abca9449609f35/lib/Util/ExtAPI.cpp#L146
* assign02 - https://github.com/SVF-tools/SVF/blob/6e0d7f31b230cb90e16ec45277abca9449609f35/lib/Util/ExtAPI.cpp#L152

PR #895, opened 2022-11-03
BasicTypes: fix missing include

gerion0:
Without this line, SVF does not compile for me currently.


yuleisui:
which llvm version you are using?

gerion0:
I have tested this with 14 and 15 (just the compilation part).

yuleisui:
> I have tested this with 14 and 15 (just the compilation part).

I see.

gerion0:
The critical line is: `typedef llvm::GCStatepointInst GCStatepointInst;`. According to https://llvm.org/doxygen/classllvm_1_1GCStatepointInst.html this is in an extra file.

PR #907, opened 2022-11-15
LLVM 15 fixes

gerion0:
This fixes one typo and replace one deprecated function according to [the LLVM documentation](https://llvm.org/docs/OpaquePointers.html).

BTW, is there any plan how to continue with LLVM 16? As far as I can see, SVF relies heavily on pointer types which will be removed entirely with LLVM 16. We use SVF for [ARA](https://github.com/luhsra/ara) (currently an old version), which in turn also relies on pointer types for other analyses. Migrating to LLVM 16 would make the analyses much more inaccurate. Do you know of any discussion about that topic?

### DIFF
--- a/include/SVF-FE/LLVMUtil.h
+++ b/include/SVF-FE/LLVMUtil.h
@@ -68,7 +68,7 @@ static inline Type *getPtrElementType(const PointerType* pty)
 #if (LLVM_VERSION_MAJOR < 14)
     return pty->getElementType();
 #else
-    assert(!pty->isOpaque() && "Opaque Pointer is used, please recompile the source adding '-Xclang -no-opaque-pointer'");
+    assert(!pty->isOpaque() && "Opaque Pointer is used, please recompile the source adding '-Xclang -no-opaque-pointers'");
     return pty->getNonOpaquePointerElementType();
 #endif
 }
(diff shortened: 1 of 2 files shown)


yuleisui:
Thanks for the change. 

In fact, the pointee type is still (implicitly) embedded in each of their memory access instructions (llvm-16 document). We will incorporate the change into the newer version of SVF. It is not hard after my quick read.

SVF will soon have its own SVFValue and SVFType (as an extra buffer layer for different languages not only LLVM), which will provide a more accessible refactoring base for later LLVMs.

gerion0:
> In fact, the pointee type is still (implicitly) embedded in each of their memory access instructions (llvm-16 document). We will incorporate the change into the newer version of SVF. It is not hard after my quick read.

It's the question if you are able to always get the memory access location. We use it mainly to improve/change the CallGraph of the SVF [by adding all possible targets for function pointers](https://github.com/luhsra/ara/blob/main/ara/steps/native/resolve_function_pointer.cpp).

In older LLVM versions the call of an indirect function looks like (simplified): `call some_function_pointer(Foo*, int)` which allows to restrict the pointer to all functions with exactly that signature. With LLVM 15+, this results in: `call some_function_pointer(ptr, int)` which makes the whole analyse a lot more inaccurate (especially for C++, the this pointer is not resolved anymore so the mapping between object and function is not possible anymore). I'm not sure how to get back the memory access instruction of ptr (for example in C++ the this pointer is an offsetted acces to a vtable which is not simple to follow).
AFAIK, you need the callgraph _before_ being able to build the value flow graph which allows to resolve these pointers again.

gerion0:
Ok, thanks for the explanation.

PR #909, opened 2022-11-15
python3 perf_compare.py diff_tests/perf_history/

colindelotavo:
Jobs have been updated to compare the latest performance logs before pushing to SVF-tools/Test-Suite.


.github/workflows/github-action.yml
```diff
@@ -88,8 +88,15 @@ jobs:
       - name: ctest performance tests on cruxbc
         working-directory: ${{github.workspace}}/Release-build
         if: runner.os == 'Linux'
-        run: 
+        run: |
           ctest -R diff-perf-cruxbc -VV
+          # git pull
```
yuleisui, line 93: We need git pull first

colindelotavo: > We need git pull first

Uncommented in recent commit.

.github/workflows/github-action.yml
```diff
@@ -88,8 +88,15 @@ jobs:
...
         if: runner.os == 'Linux'
-        run: 
+        run: |
           ctest -R diff-perf-cruxbc -VV
+          git pull
+          cp -r ${{github.workspace}}/Release-build/Testing/Temporary/LastTest.log $GITHUB_WORKSPACE/Test-Suite/diff_tests/perf_history/perf-$(date +'%Y-%m-%dT%H:%M:%S').txt
+          cp -r ${{github.workspace}}/Release-build/Testing/Temporary/LastTest.log $GITHUB_WORKSPACE/Test-Suite/diff_tests/perf-latest.txt
+          cd $GITHUB_WORKSPACE/Test-Suite/diff_tests
+          python3 -m pip install --upgrade pip setuptools wheel
+          pip3 install -r ./requirements.txt
```
yuleisui, line 98: can you wrap these in your python script?

yuleisui: Will these installations cause the CI to slow much

colindelotavo: If we really wanted to, we could wrap it in the python script, but this is not recommended, as described [here](https://pip.pypa.io/en/latest/user_guide/#using-pip-from-your-program). After adding `time` commands, I've found the installation time to only be a few seconds in total, so the runtime is negligible and shouldn't affect the CI runtime much.

.github/workflows/svf-lib_publish.yml
```diff
@@ -94,8 +94,11 @@ jobs:
             git pull
             cp -r ${{github.workspace}}/Release-build/Testing/Temporary/LastTest.log $GITHUB_WORKSPACE/Test-Suite/diff_tests/perf_history/perf-$(date +'%Y-%m-%dT%H:%M:%S').txt
             cp -r ${{github.workspace}}/Release-build/Testing/Temporary/LastTest.log $GITHUB_WORKSPACE/Test-Suite/diff_tests/perf-latest.txt
-            cd $GITHUB_WORKSPACE/Test-Suite/diff_tests  
-            git add .     
+            cd $GITHUB_WORKSPACE/Test-Suite/diff_tests
+            python3 -m pip install --upgrade pip setuptools wheel
+            pip3 install -r ./requirements.txt
```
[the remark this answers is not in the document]
colindelotavo: I've opened PR #913.

PR #910, opened 2022-11-16
Build system enhancements

MartinNowack:
Those are a couple of smaller fixes that simplify the build system:
* It allows to automatically detect and use systems' LLVM/Z3 libraries if available.
* Support MacOS Z3 dylib
* be more verbose of detected and used libraries

This also allows to use `cmake` directly avoiding the `build.sh` script if not needed, i.e. for alternative configurations.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,6 @@
 Release*/
 Debug*/
 build/
-include/Util/config.h
 html/
 Test-Suite/
 Release+Asserts/
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,18 +17,20 @@ if(NOT COMMAND add_llvm_library)
 
         add_compile_options("-Werror" "-Wall")
 
-        if(CMAKE_BUILD_TYPE MATCHES "Debug")
-            set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${COMMON_FLAGS} -O0 -fno-rtti")
-            set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${COMMON_FLAGS} -O0")
-        else()
-            set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${COMMON_FLAGS} -O3 -fno-rtti")
-            set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${COMMON_FLAGS} -O3")
         find_package(LLVM REQUIRED CONFIG
           HINTS "${LLVM_DIR}")
 
         message(STATUS "Found LLVM ${LLVM_PACKAGE_VERSION}")
         message(STATUS "Using LLVMConfig.cmake in: ${LLVM_DIR}")
 
+        if (NOT LLVM_ENABLE_RTTI)
+          add_compile_options("-fno-rtti")
+          message(STATUS "Disable RTTI")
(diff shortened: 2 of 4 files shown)


yuleisui:
It looks to me that CI was not triggered. Can you resubmit a pull request with a clean commit? Thanks

.gitignore
```diff
@@ -1,7 +1,6 @@
 Release*/
 Debug*/
 build/
-include/Util/config.h
```
yuleisui, line 4: This is an auto generated file for SVF home path and should be ignored.

MartinNowack: It will be generated into the build directory and is used from there.

https://github.com/SVF-tools/SVF/pull/910/commits/d2ec56b8082aeda9d3851128c3bd740d58d74e3a

yuleisui:
Thanks and merged.

CMakeLists.txt
```diff
@@ -2,36 +2,43 @@ cmake_minimum_required(VERSION 3.13.4)
...
         set(CMAKE_CXX_STANDARD 14)
         set(CMAKE_CXX_STANDARD_REQUIRED ON)
         # add -std=gnu++14
         set(CMAKE_CXX_EXTENSIONS ON)
 
-        set(COMMON_FLAGS "-fPIC -Werror -Wall")
+        add_compile_options("-Werror" "-Wall")
 
-        if(CMAKE_BUILD_TYPE MATCHES "Debug")
-            set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${COMMON_FLAGS} -O0 -fno-rtti")
```
yuleisui, line 26: Have you considered debugging mode? I didn't find -O0 in your change?

yuleisui:
I just found under the release mode the compiler generates a much larger libSvf.a than the previous build, which is quite strange. I will make a new pull request changing the compiler options. Please feel free to edit and make changes.

yuleisui:
https://github.com/SVF-tools/SVF/commit/9b5c56c922ab150eec924ece1e3fff21ac3af2d0

It reduces the binary of libSvf.a to around 6MB (a similar size as before) from your option, which generates 60MB for the release build.

Please help take a look at this fix and make it short or concise if you want.

CMakeLists.txt
MartinNowack: Yes I did.

`cmake` is already taking care of this and provides different modes: https://cmake.org/cmake/help/latest/manual/cmake-buildsystem.7.html#default-and-custom-configurations
Each mode is associated with different build flags:
* Debug: `-g`
* MinSizeRel: `-Os -DNDEBUG`
* Release: `-O3 -DNDEBUG`
* RelWithDebInfo: `-O2 -g -DNDEBUG`

There is no need to explicitly specify them but only adapt them if required.

MartinNowack: @yuleisui  You mean the release build under Linux with the provided packages, i.e. the ones that are downloaded?
I can have a look into it.

yuleisui:
Yes, you can take a look (1) the cmakelist.txt you did in your PR and (2) the one on the current upstream.

CMakeLists.txt
yuleisui: Yes, you can take a look (1) the cmakelist.txt you did in your PR and (2) the one on the current upstream.

PR #913, opened 2022-11-19
python3 perf_compare.py diff_tests/perf_history/

colindelotavo:
Run perf_compare.py in the svf-publish workflow before pushing to SVF-tools/Test-Suite.
Note: This pipeline will fail (expected behaviour) because the threshold percentage is set to 10, which is too low.


yuleisui:
There is an error in performance testing. You are only supposed to trigger an error if the performance has a large variance. It is not supposed to upload any performance data in `github-action.yml`. The upload can happen only in `svf-lib_publish.yml` after merging into the master branch.

yuleisui:
You can set a higher bar first to see whether your workflow works or not

yuleisui:
What I mean is not svf-lib-publish, but the github-action.yml. We need to catch the performance difference before publishing

colindelotavo:
I placed the `python` commands before the `git push` command in `svf-lib_publish.yml`, so an exit code on the script will make the `svf-lib_publish.yml` pipeline fail before it can run `git push` to the Test-Suite repository. But do you mean that the correct way to do this is to move the python code into the `github-action.yml` file instead? But if I do this, will it ever stop `svf-lib_publish.yml` from running `git push` to the Test-Suite repository if the `github-action.yml` pipeline fails?

yuleisui:
> I placed the `python` commands before the `git push` command in `svf-lib_publish.yml`, so an exit code on the script will make the `svf-lib_publish.yml` pipeline fail before it can run `git push` to the Test-Suite repository. But do you mean that the correct way to do this is to move the python code into the `github-action.yml` file instead? But if I do this, will it ever stop `svf-lib_publish.yml` from running `git push` to the Test-Suite repository if the `github-action.yml` pipeline fails?

Yes, remove your python script from ‘publish.yml’, place your command line in gihub-action.yml for early prevention of incorrect performance results.

yuleisui:
make a single commit pull request with only changes made on `github-action.yml`

PR #924, opened 2022-11-23
[Bug] Choosing the right label when creating the inverse edge

enochii:
When inserting the inverse label into the `VF` CFL graph, it seems the wrong label is chosen. Here is an example, the original edge added to graph is a `ret`, so the expected inverse edge should be `retbar`, but not `Abar`, as the picture shows. This picture is runtime snapshot.

The reason is that **the wrong kind is chosen**, we should use `edgeLabel`(number 2) rather than `edge->getEdgeKind()`(`VFGEdge::RetDirVF` is number 3) in line 489. Variable `edge` is a `VFGEdge`, the edge kind is not the same as CFL edge kind. The other 3 cases are similar.


yuleisui:
@TalbenXu can you take a look?

TalbenXu:
Hi, Yulei. Thank you for invitation.

The bug as enochii said is exist, and the pull request is correct.

Thank you enochii.

Regards
Pei

yuleisui:
Done.

PR #936, opened 2022-12-01
Refactor building symbol method into own function and call it from `buildSVFModule(Module &mod)` as well

MartinNowack:
If SVF is used as a library using existing modules, the symbol table would not have been build in the first place.

This fixes the issue.

### DIFF
--- a/include/SVF-LLVM/LLVMModule.h
+++ b/include/SVF-LLVM/LLVMModule.h
@@ -351,6 +351,7 @@ private:
     /// Invoke llvm passes to modify module
     void prePassSchedule();
     bool preProcessed;
+    void build_symbol_table() const;
 };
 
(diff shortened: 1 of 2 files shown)


yuleisui:
Can you take a look at the ctest. The tests in test-suite look never triggered.

MartinNowack:
I enabled the check for SVF-SE in https://github.com/SVF-tools/Test-Suite/pull/93

MartinNowack:
@yuleisui Ok, I see what you mean. Let me check what causes it.

MartinNowack:
Issue discussion here: https://github.com/SVF-tools/Test-Suite/issues/94

yuleisui:
Have to reopen this due to the update in master

PR #941, opened 2022-12-02
Make SvfCore independent from LLVM libraries

whst:


build.sh
```diff
@@ -7,11 +7,17 @@
 set -e # exit on first error
 
 jobs=4
+#if [[ -z CMAKE_BUILD_PARALLEL_LEVEL ]]; then
```
yuleisui, line 10: Can remove these comments

whst: Removed.

PR #945, opened 2022-12-04
Implement independent command line parser

mbarbar:
As title.

Some notable changes from LLVM:
* If the option is called `Foo`, to get the value, you do `Foo()`. If it is like `llvm::cl::bits` (`OptionMultiple`), you do `Foo(x)` instead of `Foo.isSet(x)`.
* In usage message, showing `=<xyz>`, might be nice to add later. Not sure how necessary.
* Options are `-foo` and `--foo` isn't supported.
* Everything requires an explicit initial value.
* Positional arguments get returned by the `parse` method in the order they came in. There is no way to set an `Option` object to be the positional arguments. Didn't fully get how it is in LLVM anyway.

Files implementing it are CommandLine.h and a little in CommandLine.cpp.

### DIFF
--- a/include/Util/CommandLine.h
+++ b/include/Util/CommandLine.h
@@ -26,7 +26,7 @@ protected:
     template <typename T>
     using OptionPossibility = std::tuple<T, std::string, std::string>;
 
-public:
+protected:
     OptionBase(std::string name, std::string description)
         : OptionBase(name, description, {})
     {
@@ -66,6 +66,7 @@ public:
     /// Can this option be set?
     virtual bool canSet(void) const = 0;
 
+public:
     /// Parse all constructed OptionBase children, returning positional arguments
     /// in the order they appeared.
     static std::vector<std::string> parseOptions(int argc, char *argv[], std::string description, std::string callFormat)
--- a/include/Util/CommandLine.h
+++ b/include/Util/CommandLine.h
@@ -40,13 +40,13 @@ protected:
     {
         assert(name[0] != '-' && "OptionBase: name starts with '-'");
         assert(!isHelpName(name) && "OptionBase: reserved help name");
-        assert(options.find(name) == options.end() && "OptionBase: duplicate option");
+        assert(getOptionsMap().find(name) == getOptionsMap().end() && "OptionBase: duplicate option");
 
         // Types with empty names (i.e., OptionMultiple) can handle things themselves.
         if (!name.empty())
         {
             // Standard name=value option.
-            options[name] = this;
+            getOptionsMap()[name] = this;
         }
     }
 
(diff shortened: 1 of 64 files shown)


yuleisui:
Build failed. Looks to me that those errors are warnings

yuleisui:
Seems that we don't need to include <assert>?
https://github.com/SVF-tools/SVF/actions/runs/3615936896/jobs/6093432312#step:6:74

mbarbar:
Wrong include turns out.

mbarbar:
What is the intended behaviour here? What is supposed to happen when a directory is given?
image

yuleisui:
@MartinNowack could you take a look at this issue in address-sanitiser check?

https://github.com/SVF-tools/SVF/actions/runs/3617136531/jobs/6095722254#step:16:28

yuleisui:
> What is the intended behaviour here? What is supposed to happen when a directory is given? image

My speculation is that your input module name (which is supposed to be a bc file but not a folder) when loading LLVM's module (https://github.com/SVF-tools/SVF/blob/master/lib/SVF-LLVM/LLVMModule.cpp#L443) after the following three lines? The `moduleNameVec` in your run is different from the original one upstream.

```
    std::vector<std::string> moduleNameVec;
    LLVMUtil::processArguments(argc, argv, arg_num, arg_value, moduleNameVec);
    cl::ParseCommandLineOptions(arg_num, arg_value,
                                "Whole Program Points-to Analysis\n");
```

yuleisui:
> I checked `processArguments` again. I think the behaviour introduced in this PR is more correct.
> 
> Before: non-IR file -> silently fail (not added to `moduleNameVec`). Also I haven't tested it but I suspect there may be possible to have misinterpretation of arguments as modules.
> 
> Now: non-IR file -> fail telling the user that it is not an IR file (just added commit for this; would fail before with LLVM's error message saying it couldn't parse the module or whatnot).
> 
> I think the test should be changed, I don't think it's doing anything and I think that's unintended (?).

I think it would be good to print out the file or the input string if it is can't parse as IR.

https://github.com/SVF-tools/SVF/actions/runs/3618309667/jobs/6098016542#step:15:720

yuleisui:
> > I checked `processArguments` again. I think the behaviour introduced in this PR is more correct.
> > Before: non-IR file -> silently fail (not added to `moduleNameVec`). Also I haven't tested it but I suspect there may be possible to have misinterpretation of arguments as modules.
> > Now: non-IR file -> fail telling the user that it is not an IR file (just added commit for this; would fail before with LLVM's error message saying it couldn't parse the module or whatnot).
> > I think the test should be changed, I don't think it's doing anything and I think that's unintended (?).
> 
> I think it would be good to print out the file or the input string if it is can't parse as IR.
> 
> https://github.com/SVF-tools/SVF/actions/runs/3618309667/jobs/6098016542#step:15:720

It is odd that `moduleName` is a folder `/home/runner/work/SVF/SVF/Test-Suite/`...

mbarbar:
Here it is in action
image

yuleisui:
> https://github.com/SVF-tools/SVF/actions/runs/3618309667/jobs/6098016542#step:15:720

I see. It is an issue in SVF-tools/Test-Suite/CMakeList.txt
libbz2.so.bc ==> libbz2.so
https://github.com/SVF-tools/Test-Suite/blob/26c4d56cd2864b79a6caa0ca221e6fa0b4b57c6f/CMakeLists.txt#L227

Let me update it and rerun this CI.

PR #946, opened 2022-12-05
POCR

TalbenXu:


yuleisui:
LLVM's option has been removed `const llvm::cl::opt<bool> `. You will need to use our own Option now, see `Options.cpp`.

For the option's name (CFLData), are you referring to hybridDataStruct or hybridDS4CFL or something explicit in our paper?

yuleisui:
> LLVM's option has been removed `const llvm::cl::opt<bool> `. You will need to use our own Option now, see `Options.cpp`.
> 
> For the option's name (CFLData), are you referring to hybridDataStruct or hybridDS4CFL or something explicit in our paper?

Maybe just make the option name as pocr.

PR #952, opened 2022-12-07
Resolve bug of getSExtValue(i1 true) = -1

bjjwwang:


lib/SVF-LLVM/LLVMModule.cpp
```diff
@@ -878,8 +878,10 @@ SVFConstantData* LLVMModuleSet::getSVFConstantData(const ConstantData* cd)
     else
     {
         SVFConstantData* svfcd = nullptr;
-        if(const ConstantInt* cint = SVFUtil::dyn_cast<ConstantInt>(cd))
-            svfcd = new SVFConstantInt(cd->getName().str(), getSVFType(cint->getType()), cint->getZExtValue(),cint->getSExtValue());
+        if(const ConstantInt* cint = SVFUtil::dyn_cast<ConstantInt>(cd)) {
+            svfcd = new SVFConstantInt(cd->getName().str(), getSVFType(cint->getType()), cint->getZExtValue(),
```
yuleisui, line 882: Make comments to explain this change

bjjwwang: OK I will do it soon

Issue #982, opened 2022-12-30
Saber partial leak checker problems


seviezhou:
I am trying to use the leak checker of saber to detect memory leak in the following code:

```C
#include <stdio.h>

char *malloc_wrapper()
{
    return (char*)malloc(16);
}

int main(int argc,char** argv) 
{
    char *ptr = malloc_wrapper();
    if (ptr) 
    {
        free(ptr);
        return 1;
    }
    return 0;
}
```

I compiled it into two versions: optimized one and unoptimized one:

```bash
clang -O0 -S -emit-llvm -g part-leak.c                       # this named part-leak-noopt.ll
clang -O2 -fno-inline -S -emit-llvm -g part-leak.c           # this named part-leak-opt.ll
```

There are two problems that confuse me. 

First, when I check `part-leak-opt.ll`, it reports a partial memory leak for the `ptr` variable that goes to line 16 as shown below. But as can be seen in the source code, the leak path is only taken when ptr is `NULL`, which should not be reported as memory leak, I wonder are there any workarounds for such issue?

```bash
$ ./saber -leak part-leak-opt.ll
	 PartialLeak : memory allocation at : (ln: 10 fl: part-leak.c)
		 conditional free path:
		  --> (ln: 11 fl: part-leak.c|False)
```

Second, for the unoptimized IR `part-leak-noopt.ll`, it does not report any bugs, there seems to be a problem other than issue #980.

```bash
$ ./saber -leak part-leak-noopt.ll
# no bug
```

seviezhou:
There are some problems in computing the Intra VFG guard. In optimized version, the guard is `not c0` while in unoptimized version, the guard is `true`.

yuleisui:
Do you have a solution to fix this?

seviezhou:
Making `isTestContainsNullAndTheValue` returns true whenever `cmp` contains a null value could fix the testcases, but I am not sure if it will affect other things because this changes the meaning of `isTestContainsNullAndTheValue`.

seviezhou:
A possible fix:

```diff
bool SaberCondAllocator::isTestContainsNullAndTheValue(const CmpStmt *cmp) const
{
     const SVFValue* op0 = cmp->getOpVar(0)->getValue();
     const SVFValue* op1 = cmp->getOpVar(1)->getValue();
-    if (SVFUtil::isa<SVFConstantNullPtr>(op1))
-    {
-        return true;
-        Set<const SVFValue* > inDirVal;
[...]

PR #984, opened 2022-12-31
Try to fix issue in isTestContainsNullAndTheValue

seviezhou:
For more details, please see issue #982.

### DIFF
--- a/svf/lib/SABER/SaberCondAllocator.cpp
+++ b/svf/lib/SABER/SaberCondAllocator.cpp
@@ -393,10 +393,7 @@ bool SaberCondAllocator::isTestContainsNullAndTheValue(const CmpStmt *cmp) const
         Set<const SVFValue* > inDirVal;
         for (const auto &it: getCurEvalSVFGNode()->getOutEdges())
         {
-            if (it->isIndirectVFGEdge())
-            {
-                inDirVal.insert(it->getDstNode()->getValue());
-            }
+            inDirVal.insert(it->getDstNode()->getValue());
         }
         return inDirVal.find(op0) != inDirVal.end();
     }
@@ -405,10 +402,7 @@ bool SaberCondAllocator::isTestContainsNullAndTheValue(const CmpStmt *cmp) const
         Set<const SVFValue* > inDirVal;
         for (const auto &it: getCurEvalSVFGNode()->getOutEdges())
         {
-            if (it->isIndirectVFGEdge())
-            {
-                inDirVal.insert(it->getDstNode()->getValue());
-            }
+            inDirVal.insert(it->getDstNode()->getValue());
         }
         return inDirVal.find(op1) != inDirVal.end();
     }
--- a/svf/lib/SABER/SaberCondAllocator.cpp
+++ b/svf/lib/SABER/SaberCondAllocator.cpp
@@ -374,14 +374,22 @@ bool SaberCondAllocator::isTestNotNullExpr(const SVFValue* test) const
 /*!
  * Return true if:
  * (1) cmp contains a null value
- * (2) there is an indirect edge from cur evaluated SVFG node to cmp operand
+ * (2) there is an indirect/direct edge from cur evaluated SVFG node to cmp operand
  *
  * e.g.,
+ * indirect edge:
  *      cur svfg node -> 1. store i32* %0, i32** %p, align 8, !dbg !157
  *      cmp operand   -> 2. %1 = load i32*, i32** %p, align 8, !dbg !159
  *                       3. %tobool = icmp ne i32* %1, null, !dbg !159
  *                       4. br i1 %tobool, label %if.end, label %if.then, !dbg !161
  *     There is an indirect edge 1->2 with value %0
+ * 
+ * direct edge:
+ *      cur svfg node -> 1. %3 = tail call i8* @malloc(i64 16), !dbg !22
+ *      (cmp operand)    2. %4 = icmp eq i8* %3, null, !dbg !28
+ *                       3. br i1 %4, label %7, label %5, !dbg !30
+ *     There is an direct edge 1->1 with value %3
+ * 
  */
 bool SaberCondAllocator::isTestContainsNullAndTheValue(const CmpStmt *cmp) const
 {
@@ -407,7 +415,6 @@ bool SaberCondAllocator::isTestContainsNullAndTheValue(const CmpStmt *cmp) const
         return inDirVal.find(op1) != inDirVal.end();
     }
     return false;
-
 }
 
 /*!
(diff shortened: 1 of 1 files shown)


svf/lib/SABER/SaberCondAllocator.cpp
```diff
@@ -393,10 +393,7 @@ bool SaberCondAllocator::isTestContainsNullAndTheValue(const CmpStmt *cmp) const
         Set<const SVFValue* > inDirVal;
         for (const auto &it: getCurEvalSVFGNode()->getOutEdges())
         {
-            if (it->isIndirectVFGEdge())
-            {
-                inDirVal.insert(it->getDstNode()->getValue());
-            }
+            inDirVal.insert(it->getDstNode()->getValue());
```
yuleisui, line 396: Can you add the description for the non-indirect value-flow case at/from line 385 and modify the comments where possible?

seviezhou: So I will modify the head comment for function `isTestContainsNullAndTheValue`, is that ok?

svf/lib/SABER/SaberCondAllocator.cpp
```diff
@@ -374,14 +374,22 @@ bool SaberCondAllocator::isTestNotNullExpr(const SVFValue* test) const
...
  *      cmp operand   -> 2. %1 = load i32*, i32** %p, align 8, !dbg !159
  *                       3. %tobool = icmp ne i32* %1, null, !dbg !159
  *                       4. br i1 %tobool, label %if.end, label %if.then, !dbg !161
  *     There is an indirect edge 1->2 with value %0
+ * 
+ * direct edge:
+ *      cur svfg node -> 1. %3 = tail call i8* @malloc(i64 16), !dbg !22
+ *      (cmp operand)    2. %4 = icmp eq i8* %3, null, !dbg !28
+ *                       3. br i1 %4, label %7, label %5, !dbg !30
+ *     There is an direct edge 1->1 with value %3
```
yuleisui, line 391: 1->2?

seviezhou: The `getCurEvalSVFGNode` returns the malloc node, and the `op1` is also the malloc node, so according to "`there is an indirect/direct edge from cur evaluated SVFG node to cmp operand`", I wrote `1->1`. Should it be changed to `1->2`?

```
getCurEvalSVFGNode(): AddrVFGNode ID: 5 AddrStmt: [Var22 <-- Var23]
   %3 = tail call i8* @malloc(i64 16), !dbg !22 ln: 10 fl: part-leak.c

op1:    %3 = tail call i8* @malloc(i64 16), !dbg !22 ln: 10 fl: part-leak.c
```

yuleisui: 1-2

seviezhou: Ok.

yuleisui: Good

svf/lib/SABER/SaberCondAllocator.cpp
jumormt: Should we add the value of the current SVFG Node (%3 in this example) if it is directly passed as a CMP operand? I think `inDirVal.insert(getCurEvalSVFGNode()->getValue());` should be inserted before iterating the out edges of the current VFG node.

yuleisui: @seviezhou can you have a look at @jumormt's comment?

seviezhou: From my debug result, both `op0` and `getCurEvalSVFGNode` return `%3`, and it has 2 out edges, both of them also return `%3`:

```
op0:    %3 = tail call i8* @malloc(i64 16), !dbg !22 ln: 10 fl: part-leak.c
getCurEvalSVFGNode: AddrVFGNode ID: 5 AddrStmt: [Var22 <-- Var23]
   %3 = tail call i8* @malloc(i64 16), !dbg !22 ln: 10 fl: part-leak.c

num edge: 2
dstNode: ActualParmVFGNode ID: 9 CS[ln: 13 fl: part-leak.c]ValVar ID: 22
   %3 = tail call i8* @malloc(i64 16), !dbg !22 ln: 10 fl: part-leak.c
insert dstNode:    %3 = tail call i8* @malloc(i64 16), !dbg !22 ln: 10 fl: part-leak.c

dstNode: ActualParmVFGNode ID: 21 CS[ln: 13 fl: part-leak.c]ValVar ID: 22
   %3 = tail call i8* @malloc(i64 16), !dbg !22 ln: 10 fl: part-leak.c
insert dstNode:    %3 = tail call i8* @malloc(i64 16), !dbg !22 ln: 10 fl: part-leak.c

inDirVal.find(op0) != inDirVal.end(): true
```

I am not very familiar with the internal of SVF, so I am not sure how the out edges are constructed. But based on this results, maybe we do not need to add `inDirVal.insert(getCurEvalSVFGNode()->getValue())`, but for insurance, we can add it.

jumormt: @seviezhou Thanks for your detailed answer. It is correct that these SVFG nodes contain the same variable (both `%3`) because the direct value flow is a top-level variable in the LLVM partial SSA form. I think it is necessary to add this line in case there are no direct out edges of the current SVFG node.

seviezhou: Ok, I will make a commit.

seviezhou: Patch like this would be ok:

```diff
@@ -391,6 +391,7 @@ bool SaberCondAllocator::isTestContainsNullAndTheValue(const CmpStmt *cmp) const
    if (SVFUtil::isa<SVFConstantNullPtr>(op1))
    {
        Set<const SVFValue* > inDirVal;
+       inDirVal.insert(getCurEvalSVFGNode()->getValue());
        for (const auto &it: getCurEvalSVFGNode()->getOutEdges())
        {
            if (it->isIndirectVFGEdge())
@@ -403,6 +404,7 @@ bool SaberCondAllocator::isTestContainsNullAndTheValue(const CmpStmt *cmp) const
    else if (SVFUtil::isa<SVFConstantNullPtr>(op0))
    {
        Set<const SVFValue* > inDirVal;
+       inDirVal.insert(getCurEvalSVFGNode()->getValue());
        for (const auto &it: getCurEvalSVFGNode()->getOutEdges())
        {
            if (it->isIndirectVFGEdge())
```

yuleisui:
@jumormt let me know if this one can merge

jumormt:
@yuleisui It looks good to me.

PR #992, opened 2023-01-08
Fix a assertion failure cause i128 type

ChoKyuWon:
When switch-case expression takes i128 type on case destination, getSExtValue raises assertion failure. This patch checks the bit width of ConstantInt and runs getSExtValue only if ConstantInt has a bit width smaller than 64.

Signed-off-by: ChoKyuWon <kyuwoncho18@gmail.com>


yuleisui:
A good fx, thanks, wiil merge

Issue #991, opened 2023-01-08
Problems of Handling Global SVFG Nodes


seviezhou:
I think there are some imperfections in terms of how SVF handles global SVFG Nodes. 

The usage of global nodes starts from the SVFG building process. During the `SaberSVFGBuilder`, two kinds of SVFG nodes are considered to be global nodes:

1. The node is a `StoreSVFGNode` and its destination's points-to set contains a global objects.
2. The node is a `LoadSVFGNode` and its source's points-to set contains a global objects.

During the source-sink analysis, a forward analysis is first performed for each slice. During the traversal, if the destination node of an edge is a global node then the slice is not analyzed anymore since SVF assumes that the client uses the memory until the program exits. 

I observe that this filtering is important because it eliminates the analysis for a large number of slice. For example, in OpenSSL, I found that over half of all the slicing is abandoned because of this global node checking. Disable this will significantly increase the analysis time. 

However, I found the global node identification is too coarse-grained because of the imprecise pointer analysis, which marks too many incorrect global nodes and potentially cause missing bugs. I made a simple example below:

```C
#include <stdlib.h>

struct data_st* global_st;

struct data_st 
{
    char *tmp_store;
};

void* malloc_wrapper(size_t len)
{
    return malloc(len);
}

void foo(int c)
{
    struct data_st *st = (struct data_st*)malloc_wrapper(sizeof(struct data_st));
    void* ptr = malloc_wrapper(16);

    st->tmp_store = ptr;    // delete this leak will be detected
    
    free(st);               // no leak
    if (c > 10) free(ptr);  // partial leak
}

int main(int argc,char** argv) 
{
    struct data_st *sptr = (struct data_st*)malloc_wrapper(sizeof(struct data_st));
    global_st = sptr;       // no leak, poison malloc_wrapper to be global
    foo(argc);
    return 0;
}
```

Compile it and check with saber shows not bugs:

```bash
clang -S -emit-llvm -g leak.c
saber -leak leak.ll
# no bugs
```

But the partial leak at line 23 is supposed to be reported. The reason for this result is:

1. In `main` function, the returned pointer of `malloc_wrapper` is stored to a global pointer, which makes the points-to set of `malloc_wrapper` contains a global variable, resulting all stores or loads of returned pointer from `malloc_wrapper` to be marked as global nodes. Therefore, `st->tmp_store = ptr;` is marked as a global node.
[...]

PR #997, opened 2023-01-10
Add An Option to Ignore External Functions During Global Collection

seviezhou:
Please refer to issue #991

### DIFF
--- a/svf/include/Util/Options.h
+++ b/svf/include/Util/Options.h
@@ -224,7 +224,7 @@ public:
     static const Option<bool> PrintPathCond;
 
     // SaberSVFGBuilder.cpp
-    static const Option<bool> IgnoreExtFun;
+    static const Option<bool> CollectExtRetGlobals;
 
     // SVFUtil.cpp
     static const Option<bool> DisableWarn;
--- a/svf/include/Util/Options.h
+++ b/svf/include/Util/Options.h
@@ -223,6 +223,9 @@ public:
     // SaberCondAllocator.cpp
     static const Option<bool> PrintPathCond;
 
+    // SaberSVFGBuilder.cpp
+    static const Option<bool> IgnoreExtFun;
+
     // SVFUtil.cpp
     static const Option<bool> DisableWarn;
 
--- a/svf/lib/SABER/SaberSVFGBuilder.cpp
+++ b/svf/lib/SABER/SaberSVFGBuilder.cpp
@@ -99,7 +99,7 @@ void SaberSVFGBuilder::collectGlobals(BVDataPTAImpl* pta)
     }
 }
 
-PointsTo& SaberSVFGBuilder::CollectPtsChain(BVDataPTAImpl* pta,NodeID id, NodeToPTSSMap& cachedPtsMap)
+PointsTo& SaberSVFGBuilder::CollectPtsChain(BVDataPTAImpl* pta, NodeID id, NodeToPTSSMap& cachedPtsMap)
 {
     SVFIR* pag = svfg->getPAG();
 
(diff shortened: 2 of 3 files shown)


seviezhou:
Should I make a commit to test SVF when the new option is enabled?

svf/lib/Util/Options.cpp
```diff
@@ -679,6 +679,14 @@ const Option<bool> Options::PrintPathCond(
 );
 
 
+// SaberSVFGBuilder.cpp
+const Option<bool> Options::IgnoreExtFun(
+    "saber-ignore-ext-fun",
```
yuleisui, line 684: saber-collect-extret-globals

svf/lib/Util/Options.cpp
```diff
@@ -679,6 +679,14 @@ const Option<bool> Options::PrintPathCond(
...
+    "Don't include pointers returned by external function during collecting globals",
+    false
```
yuleisui, line 686: make it true as default.

svf/lib/SABER/SaberSVFGBuilder.cpp
```diff
@@ -110,6 +111,21 @@ PointsTo& SaberSVFGBuilder::CollectPtsChain(BVDataPTAImpl* pta,NodeID id, NodeTo
     else
     {
         PointsTo& pts = cachedPtsMap[baseId];
+        // base object
+        if (Options::IgnoreExtFun())
+        {
+            if (pta->isFIObjNode(baseId))
+            {
+                PAGNode* pn = pag->getGNode(baseId);
+                const SVFValue* value = pn->getValue();
```
yuleisui, line 120: pn may not have a value, it could be a dummy node.

if(pta->isFIObjNode(baseId) && pag->getGNode(baseId)->hasValue())
{
   const SVFCallInst* inst = SVFUtil::dyn_cast<SVFCallInst>( pag->getGNode(baseId)->getValue());
   if(inst ...)
   ....

}

yuleisui (commented):
Please see my comments

svf/include/Util/Options.h
```diff
@@ -223,6 +223,9 @@ class Options
     // SaberCondAllocator.cpp
     static const Option<bool> PrintPathCond;
 
+    // SaberSVFGBuilder.cpp
+    static const Option<bool> IgnoreExtFun;
```
yuleisui, line 227: CollectExtRetGlobals

svf/lib/Util/Options.cpp
seviezhou: Making it true as default will cost performance downgrade, should we do some profiling first?

yuleisui: See my new name. The name should be used the other way around to make it clear. `True` is your current false.

seviezhou: Ok

svf/lib/SABER/SaberSVFGBuilder.cpp
```diff
@@ -99,17 +100,32 @@ void SaberSVFGBuilder::collectGlobals(BVDataPTAImpl* pta)
     }
 }
 
-PointsTo& SaberSVFGBuilder::CollectPtsChain(BVDataPTAImpl* pta,NodeID id, NodeToPTSSMap& cachedPtsMap)
+PointsTo& SaberSVFGBuilder::CollectPtsChain(BVDataPTAImpl* pta, NodeID id, NodeToPTSSMap& cachedPtsMap)
```
yuleisui, line 103: Could make some comments to explain this design choice based on your issue and also leave the issue link here too (https://github.com/SVF-tools/SVF/issues/991)?

Thanks,

seviezhou: Ok.

yuleisui:
Looks good and merged.

Issue #1002, opened 2023-01-23
FunptrDDAClient doesn't analyze some pointers


sinotca529:
Hi! Thanks for the great tool!

In `FunptrDDAClient::answerQueries`, I found a problem where necessary queries were not issued.
The problem is caused because the condition of `SVFIR::isValidPtr` is too strict.

# The problem I found
`FunptrDDAClient` does not analyze the 0th argument of lv1a and lv1b (the variable `cb`).

Step to reproduce:
1. Run `clang-12 -g -S -emit-llvm -O2 main.c`
2. Run `dvf -query=funptr -dfs -print-query-pts main.ll`

I tested on [the commit](https://github.com/SVF-tools/SVF/tree/17193ff5a3765e565824570c97a55cdbdf92772b)

main.c :
```c
typedef void (*CB)(void);
typedef void (*LV1)(CB);

void a(void) {}

__attribute__((noinline))
void lv1a(CB cb) { cb(); }

__attribute__((noinline))
void lv1b(CB cb) { cb(); }

int main(int argc, char **argv) {
    LV1 lv1 = argc % 2 ? lv1a : lv1b;
    lv1(a);
    return 0;
}
```

<details>
<summary>main.ll</summary>

```llvm
; ModuleID = 'main.c'
source_filename = "main.c"
target datalayout = "e-m:e-p270:32:32-p271:32:32-p272:64:64-i64:64-f80:128-n8:16:32:64-S128"
target triple = "x86_64-pc-linux-gnu"

; Function Attrs: norecurse nounwind readnone uwtable willreturn
define dso_local void @a() #0 !dbg !7 {
  ret void, !dbg !10
}

; Function Attrs: noinline nounwind uwtable
define dso_local void @lv1a(void ()* nocapture %0) local_unnamed_addr #1 !dbg !11 {
  call void @llvm.dbg.value(metadata void ()* %0, metadata !17, metadata !DIExpression()), !dbg !18
  tail call void %0() #4, !dbg !19
  ret void, !dbg !20
}

; Function Attrs: noinline nounwind uwtable
define dso_local void @lv1b(void ()* nocapture %0) local_unnamed_addr #1 !dbg !21 {
  call void @llvm.dbg.value(metadata void ()* %0, metadata !23, metadata !DIExpression()), !dbg !24
  tail call void %0() #4, !dbg !25
  ret void, !dbg !26
}

; Function Attrs: nounwind uwtable
[...]

PR #1003, opened 2023-01-23
fix the condition of SVFIR::isValidPointer

sinotca529:
Fix the issue #1002

### DIFF
--- a/svf/lib/SVFIR/SVFIR.cpp
+++ b/svf/lib/SVFIR/SVFIR.cpp
@@ -690,11 +690,24 @@ bool SVFIR::isNonPointerObj(NodeID id) const
     }
 }
 /*
- * If this is a dummy node we assume it is not a pointer here
+ * If this is a dummy node or node does not have incoming edges and outcoming edges we assume it is not a pointer here.
+ * However, if it is a pointer and it is an argument of a function definition, we assume it is a pointer here.
  */
 bool SVFIR::isValidPointer(NodeID nodeId) const
 {
     SVFVar* node = pag->getGNode(nodeId);
+
+    if (node->hasValue() && node->isPointer())
+    {
+        if(const SVFArgument* arg = SVFUtil::dyn_cast<SVFArgument>(node->getValue()))
+        {
+            if (!(arg->getParent()->isDeclaration()))
+                return true;
+        }
+    }
+
+    if ((node->getInEdges().empty() && node->getOutEdges().empty()))
+        return false;
     return node->isPointer();
 }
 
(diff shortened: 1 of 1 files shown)


yuleisui:
It looks this will affect quite a few other cases. Would be good to dig it a bit or make it just for DDA.

sinotca529:
Observation of errors in the regression test and testing with a hand-made code revealed that errors occur when passing string literals to `puts`/`printf`.

This is the hand-made code that causes an error:
```c
#include <stdio.h>
int main(void) {
    puts("foo");
    return 0;
}
```

The error occured when SVF tries to analyze the 0th arg of `puts`.
Because `puts` is a declaration and `SVFIRBuilder::handleExtCall` does not make call edges from actual parameters to formal parameters, the 0th arg of `puts` has no incoming edges and no outcoming edges.
So, it satisfies `SVFIR::isValidPointer` wrongly.

To avoid this exception, I updated the condition of `SVFIR::isValidPointer` as:
```c
/*
 * If this is a dummy node or node does not have incoming edges and outcoming edges we assume it is not a pointer here.
 * However, if it is a pointer and it is an argument of a function definition, we assume it is a pointer here.
 */
bool SVFIR::isValidPointer(NodeID nodeId) const
{
    SVFVar* node = pag->getGNode(nodeId);

    if (node->hasValue() && node->isPointer())
    {
        if(const SVFArgument* arg = SVFUtil::dyn_cast<SVFArgument>(node->getValue()))
        {
            if (!(arg->getParent()->isDeclaration()))
                return true;
        }
    }

    if ((node->getInEdges().empty() && node->getOutEdges().empty()))
        return false;
    return node->isPointer();
}
```

sinotca529:
Observation of errors in the regression test and testing with a hand-made code revealed that errors occur when passing string literals to `puts`/`printf`.

This is the hand-made code that causes an error:
```c
#include <stdio.h>
int main(void) {
    puts("foo");
    return 0;
}
```

The error occured when SVF tries to analyze the 0th arg of `puts`.
Because `puts` is a declaration and `SVFIRBuilder::handleExtCall` does not make call edges from actual parameters to formal parameters, the 0th arg of `puts` has no incoming edges and no outcoming edges.
So, it satisfies `SVFIR::isValidPointer` wrongly.

To avoid this exception, I updated the condition of `SVFIR::isValidPointer` as:
```c
/*
 * If this is a dummy node or node does not have incoming edges and outcoming edges we assume it is not a pointer here.
 * However, if it is a pointer and it is an argument of a function definition, we assume it is a pointer here.
 */
bool SVFIR::isValidPointer(NodeID nodeId) const
{
    SVFVar* node = pag->getGNode(nodeId);

    if (node->hasValue() && node->isPointer())
    {
        if(const SVFArgument* arg = SVFUtil::dyn_cast<SVFArgument>(node->getValue()))
        {
            if (!(arg->getParent()->isDeclaration()))
                return true;
        }
    }

    if ((node->getInEdges().empty() && node->getOutEdges().empty()))
        return false;
    return node->isPointer();
}
```

PR #1008, opened 2023-02-06
Json module Read & Write

xudon9:
Note that we also need SVFIR read/write for full analysis from JSON file.
Coverage for SVFModuleRW temporarily disabled.

### DIFF
--- a/svf-llvm/lib/LLVMModule.cpp
+++ b/svf-llvm/lib/LLVMModule.cpp
@@ -123,13 +123,9 @@ SVFModule* LLVMModuleSet::buildSVFModule(const std::vector<std::string> &moduleN
 
     build_symbol_table();
 
-    auto* rawSvfModule = svfModule.get();
+    svfModule->writeToJson(dumpJson());
 
-    std::string jsonPath = dumpJson();
-    if (!jsonPath.empty())
-        SVFModuleWrite(rawSvfModule, jsonPath);
-
-    return rawSvfModule;
+    return svfModule.get();
 }
 
 void LLVMModuleSet::build_symbol_table() const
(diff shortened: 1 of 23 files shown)


svf-llvm/lib/LLVMModule.cpp
```diff
@@ -119,8 +123,15 @@ SVFModule* LLVMModuleSet::buildSVFModule(const std::vector<std::string> &moduleN
 
     build_symbol_table();
 
-    return svfModule.get();
+    auto* rawSvfModule = svfModule.get();
```
yuleisui, line 126: put into SVFModule::writeToJason()

svf/lib/SVFIR/SVFModuleRW.cpp
```diff
@@ -0,0 +1,1206 @@
...
+}
+
+SVFValue* createValue(SVFValue::SVFValKind kind)
+{
+    switch (kind)
+    {
+    default:
+        SVFUtil::errs() << "Impossible SVFValue kind " << kind
+                        << " in createValue()\n";
+        abort();
```
yuleisui, line 697: put an assertion();

svf/lib/SVFIR/SVFModuleRW.cpp
```diff
@@ -0,0 +1,1206 @@
...
+    } while (0)
+
+SVFType* createType(SVFType::SVFTyKind kind)
+{
+    switch (kind)
+    {
+    default:
+        SVFUtil::errs() << "Impossible SVFTyKind " << kind
+                        << " in createType()\n";
+        abort();
```
yuleisui, line 671: make sure every abort() has an assertion

PR #1017, opened 2023-02-20
SVFIR Json Writer

xudon9:
- Create template class `GenericGraphWriter/Reader` so that we can reuse the code for IRGraph, ICFG, CHGraph, etc.
- JSON are generated by two formats: human readable form (more redundant but easy to read), disk saving form

### DIFF
--- a/svf/include/SVFIR/SVFModuleRW.h
+++ b/svf/include/SVFIR/SVFModuleRW.h
@@ -76,18 +76,15 @@ class SVFModuleWrite
 {
 private:
     const SVFModule* module; ///< Borrowed pointer to the SVFModule.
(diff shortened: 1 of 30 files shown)


svf/include/SVFIR/SVFModuleRW.h
```diff
@@ -195,6 +192,15 @@ class SVFModuleRead
     cJSON* readJson(cJSON* iter, SVFMetadataAsValue* value);
 };
 
+// class SVFIRWriter
```
yuleisui, line 195: Remove this commented code?

xudon9: Ok. This is commented out because I haven't fully implement it.

yuleisui:
should ignore SVFIRRW for code coverage.

PR #1069, opened 2023-04-27
improve event management

JoelYYoung:


svf/include/Util/SVFBugReport.h
```diff
@@ -141,7 +131,7 @@ class GenericBug
 public:
     /// note: should be initialized with a bugEventStack
     GenericBug(BugType bugType, EventStack bugEventStack):
```
yuleisui, line 133: EventStack bugEventStack => const EventStack& bugEventStack

JoelYYoung: fixed right now

PR #1075, opened 2023-05-04
Extra information in SVF File system 

xudon9:
- Add `IdAllocator`
- Add `typeLocSetsMap`
- Remove legacy `SVFModuleRead` & `SVFModuleWrite`
- Fix `SparseBitVector` serialization / deserialization bug.
- Rename SVFIRRW.h/.cpp as SVFFileSystem.h/.cpp


yuleisui:
We need to change the names SVFIRRW to SVFFileSystem

PR #1082, opened 2023-05-08
cmake: use paths relative to project source, not cmake top level

adriaanjacobs:
We use SVF as a submodule and build it as part of a larger CMake project. `CMAKE_SOURCE_DIR` evaluates to the source dir where the top level CMake for the current build was invoked (dir of our top-level CMakeLists.txt). Instead, `PROJECT_SOURCE_DIR` evaluates to the top-level dir of the current (sub-)project. When building SVF normally, this makes no difference, but in our case, it does. 

The same goes for the `*_BINARY_DIR` variables: we ideally want SVF's generated `config.h` to be under it's own subdir in our build folder, not globally (to avoid messing with our own generated include paths, for instance). 

After this change, we're able to continue using SVF as a submodule by simply doing:
```
set(SVF_DIR ${CMAKE_CURRENT_SOURCE_DIR}/SVF/)
set(svf_libs SvfLLVM SvfCore)
include_directories(${SVF_DIR}/svf/include/)
include_directories(${SVF_DIR}/svf-llvm/include/)
include_directories(${PROJECT_BINARY_DIR}/SVF/include/)
add_subdirectory(${SVF_DIR})
```

from a higher-level CMakeLists.txt

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,8 +2,8 @@ cmake_minimum_required(VERSION 3.13.4)
 
 project("SVF")
 
-configure_file(${CMAKE_SOURCE_DIR}/.config.in
-               ${CMAKE_BINARY_DIR}/include/Util/config.h)
+configure_file(${PROJECT_SOURCE_DIR}/.config.in
+               ${PROJECT_BINARY_DIR}/include/Util/config.h)
 
 # We need to match the build environment for LLVM: In particular, we need C++14
 # and the -fno-rtti flag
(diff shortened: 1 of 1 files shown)


yuleisui:
After merging, It looks that the CI failed because the libSvfCore.a is much larger than before (typically around 10 MB). Do you have any idea about this?

https://github.com/SVF-tools/SVF/actions/runs/4942584365/jobs/8836248588#step:10:40

yuleisui:
@adriaanjacobs Did that mean we generate the debug libs rather than the released ones?

adriaanjacobs:
Apologies, looks like my unrelated -ggdb compiler options snuck in. Thanks for fixing!

Issue #1078, opened 2023-05-05
ExtAPI: dubious handling of some POSIX functions


adriaanjacobs:
Debugging an unsound points-to set led me on a wild goose chase of ExtAPI functions that do not seem to be modeled entirely according to the semantics of the man pages.

### Unsoundly modeled dynamic allocations
- `asprintf` and  `vasprintf` allocate memory (should be freed later by the caller) and stores a pointer to it in the first argument (output argument). SVF's current extapi.json treats this function as EFT_NOOP however. Should these not be an EFT_A0R_NEW?  -> PR: #1084 
- `realpath` can allocate memory for its return value, but ExtAPI currently only models a copy edge between its arguments, not the return value being a (possibly) newly allocated object (it depends on the arguments being non-null). 
- `getaddrinfo` allocates a linked list of structures and returns it via output argument. It's quite unique: the allocated object is initialized and can refer to other, new allocations. I don't think there's a good way to model this with the current ExtAPI grammar.

### Pointers to static buffers modeled as EFT_NOSTRUCT_ALLOC
Many functions in the ExtAPI that are modeled as EFT_NOSTRUCT_ALLOC return a pointer to a statically allocated buffer instead. The return values of multiple calls to these functions should alias, but, if my understanding is right, they currently don't. If I understand correctly, EFT_STAT would be more appropriate? Examples include:
- `getenv`, `getlogin`, `getpass`. These are all specified by the man pages to return a pointer into a static buffer, so it's not an implementation detail. For the `strerror` and `strsignal` family of functions, this is not (exactly) specified by the man pages, but it's implied and most sane implementations do so.  
In fact, most of the EFT_NOSTRUCT_ALLOC functions in the extapi seem to actually be EFT_STATs?

### Missing ExtAPI entries
- `pmtraceerrstr` is also one of these error-handling string functions, but not modeled by ExtAPI right now
- `open_memstream`/`open_wmemstream` are absent, but they allocate dynamic memory
- `wcsdup` is similar to `strdup`, but not modeled

### Methodology
There's more than the ones I listed here. I obtained some by just stumbling on them, but others by systematically scanning the man pages for tell-tale terminology. Most man pages of functions that dynamically allocate memory mention that the returned pointer can be "passed to free". Most man pages of functions that deal with statically allocated buffers mention "statically allocated" or "static buffer". I've found these queries to produce pretty precise results, with little noise. 

[...]

PR #1084, opened 2023-05-08
ExtAPI: update unsoundly modeled functions

adriaanjacobs:
See #1078. This starting patch aims to fix `asprintf` and `vasprintf`, others to follow. We inferred the right grammar for these from `posix_memalign`.

### DIFF
--- a/svf/include/Util/ExtAPI.json
+++ b/svf/include/Util/ExtAPI.json
@@ -518,8 +518,16 @@
     "asprintf": {
         "return":  "int",
         "arguments":  "(char **restrict, const char *restrict, ...)",
-        "type": "EFT_NOOP",
-        "overwrite_app_function": 1
+        "type": "EFT_A0R_NEW",
+        "overwrite_app_function": 1,
+        "AddrStmt": {
+            "src": "Obj",
+            "dst": "Dummy"
+        },
+        "StoreStmt": {
+            "src": "Dummy",
+            "dst": "Arg0"
+        }
     },
     "atexit":   {
         "return":  "int",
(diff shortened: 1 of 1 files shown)


adriaanjacobs:
Looks like some of the bitcode test files are triggering an assertion in VFG.h:
```
inline NodeID getDef(const PAGNode* pagNode) const
{
    PAGNodeToDefMapTy::const_iterator it = PAGNodeToDefMap.find(pagNode);
    assert(it!=PAGNodeToDefMap.end() && "SVFVar does not have a definition??");
    return it->second;
}
```
Not sufficiently familiar with SVF's internals to figure this one out. @yuleisui is this caused by this ExtAPI patch? I would expect that the similarity with `posix_memalign` would not cause issues for these functions. Does the fact that these are vararg functions have something to do with this?

yuleisui:
> Looks like some of the bitcode test files are triggering an assertion in VFG.h:
> 
> ```
> inline NodeID getDef(const PAGNode* pagNode) const
> {
>     PAGNodeToDefMapTy::const_iterator it = PAGNodeToDefMap.find(pagNode);
>     assert(it!=PAGNodeToDefMap.end() && "SVFVar does not have a definition??");
>     return it->second;
> }
> ```
> 
> Not sufficiently familiar with SVF's internals to figure this one out. @yuleisui is this caused by this ExtAPI patch? I would expect that the similarity with `posix_memalign` would not cause issues for these functions. Does the fact that these are vararg functions have something to do with this?

Let us have a look. @shuangxiangkan could you take a look at this assertion error?

adriaanjacobs:
I wrongfully modeled them as EFT_NOSTRUCT_ALLOCs instead of EFT_A0R_NEW (fixed in recent force push). This change might fix the test failures.

PR #1093, opened 2023-05-16
tiny improvement

joeyjiaojg:
1. increase build speed
2. fix no llvm:None


build.sh
```diff
@@ -6,7 +6,7 @@
 # Dependencies include: build-essential libncurses5 libncurses-dev cmake zlib1g-dev
 set -e # exit on first error
 
-jobs=4
+jobs=$(nproc)
```
yuleisui, line 9: Would be good to keep it as 4 since some machines might be preemptive and struggling with more cores as per our experience.

build.sh
```diff
@@ -221,10 +221,12 @@ BUILD_DIR="./${BUILD_TYPE}-build"
 
 rm -rf "${BUILD_DIR}"
 mkdir "${BUILD_DIR}"
+EXTRA_FLAGS=""
+[ "${BUILD_WITH_CLANG}" != "" ] && EXTRA_FLAGS="-DCMAKE_CXX_COMPILER=$LLVM_DIR/bin/clang++ -DLLVM_DIR=$LLVM_DIR"
```
yuleisui, line 225: I would keep it as it is as some machines might not have clang installed.

svf/include/Util/Annotator.h
```diff
@@ -112,7 +112,7 @@ class Annotator
             values.push_back(val);
             //llvm::ArrayRef<llvm::Metadata*> ar(metavalues);
             // FIXME: delete the old MDNode
-            inst->setMetadata(str, MDNode::get(inst->getContext(), llvm::None));
+            inst->setMetadata(str, MDNode::get(inst->getContext(), nullptr));
```
yuleisui, line 115: good fix!

build.sh
joeyjiaojg: limit to 4 at least

build.sh
joeyjiaojg: Added BUILD_WITH_CLANG

joeyjiaojg (commented):
updated, please check again?

svf/include/Util/Annotator.h
yuleisui: Could you make this fix as a separate pull request?

build.sh
yuleisui: Do we need to specify clang build? SVF is supposed to be built with g++ and others as long as it can compile

joeyjiaojg: I tried by using g++ which doesn't compile if I'm using android llvm-17, it only compiles clang, so I prefer to set BUILD_WITH_CLANG and use clang if perfered, it won't influence if using g++ by default.

svf/include/Util/Annotator.h
yuleisui: Could you make another PR with this line of fix only, so that I could merge it first?

joeyjiaojg: https://github.com/SVF-tools/SVF/pull/1094

build.sh
joeyjiaojg: > supposed to be built with g++ an

There are the flags I used
```
export LLVM_DIR=/usr/src/clang-r487747
export LLVM_CONFIG=$LLVM_DIR/bin/llvm-config
export PATH=$LLVM_DIR/bin:$PATH
```
And build cmd
```
BUILD_WITH_CLANG=true source build.sh
```

yuleisui: I see. Can this be configured in a nicer way and make sure you have the comments at the top of this file to explain the option too

yuleisui: Why not call it compiler flag rather than extra flag and make it earlier so that people can see it together with other options for maintenance purposes

build.sh
yuleisui: Limited to 4 at most, or let us just leave it as 4 or make it configurable. Sometimes the compilation seems to be quite intense and will not be able to finish when competing with each other

joeyjiaojg (commented):
done, pls review again.

build.sh
```diff
@@ -6,11 +6,13 @@
 # Dependencies include: build-essential libncurses5 libncurses-dev cmake zlib1g-dev
 set -e # exit on first error
 
-jobs=4
+jobs=$(echo $(nproc) || echo $(sysctl -n hw.physicalcpu))
```
yuleisui, line 9: Can we make it one line rather than a if condition

build.sh
```diff
@@ -6,11 +6,13 @@
...
+[ $jobs -lt 4 ] && jobs=4
 
 #########
 # VARs and Links
 ########
+CMAKE_EXTRA_FLAGS=""
```
yuleisui, line 15: Remove it

build.sh
```diff
@@ -221,10 +223,11 @@ BUILD_DIR="./${BUILD_TYPE}-build"
 
 rm -rf "${BUILD_DIR}"
 mkdir "${BUILD_DIR}"
+[ "${BUILD_WITH_CLANG}" != "" ] && CMAKE_EXTRA_FLAGS="-DCMAKE_CXX_COMPILER=$LLVM_DIR/bin/clang++ -DLLVM_DIR=$LLVM_DIR"
 cmake -D CMAKE_BUILD_TYPE:STRING="${BUILD_TYPE}" \
```
yuleisui, line 227: Can we add “ DCMAKE_CXX_COMPILER” here similar to this line? The basic principle is to make sure all macros are not ad hoc and should be as consistent as possible and reduce arbitrary insertions and try to keep it as simple as possible

joeyjiaojg: I might not quite follow, are u talking about the way https://github.com/SVF-tools/SVF/pull/1093/commits/f723d9de8925ee7cc4e9de57b4d720f4cfe535bd?

build.sh
```diff
@@ -224,6 +225,8 @@ mkdir "${BUILD_DIR}"
...
     -DSVF_ENABLE_ASSERTIONS:BOOL=true            \
     -DSVF_SANITIZE="${SVF_SANITIZER}"            \
+    -DBUILD_WITH_CLANG:BOOL=${BUILD_WITH_CLANG} \
+    -DLLVM_DIR=${LLVM_DIR} \
```
yuleisui, line 229: Do we need DLLVM_DIR here when building?

build.sh
yuleisui: Would it be faster when you were trying to build svf using more than 4 threads? I think we can leave it as 4 and people can simply adjust it by themselves

build.sh
joeyjiaojg: yes, otherwise build failure

build.sh
joeyjiaojg: much faster when I use blade with higher jobs.

joeyjiaojg:
The new patch should be better now.

joeyjiaojg:
```
CMake Error at CMakeLists.txt:3 (project):
  No CMAKE_CXX_COMPILER could be found.

  Tell CMake where to find the compiler by setting either the environment
  variable "CXX" or the CMake cache entry CMAKE_CXX_COMPILER to the full path
  to the compiler, or to the compiler name if it is in the PATH.
```
OK, no better way I know without using any bash condition variable.

joeyjiaojg:
@yuleisui ok, this should be final?

people can choose to build with clang and other JOBS value
```
export LLVM_DIR=/usr/src/clang-android && export LLVM_CONFIG=$LLVM_DIR/bin/llvm-config && JOBS=$(nproc) COMPILE_FLAGS="-DCMAKE_CXX_COMPILER=${LLVM_DIR}/bin/clang++ -DLLVM_DIR=${LLVM_DIR}" ./build.sh
```

PR #1098, opened 2023-05-21
match args of callee and callsite

jumormt:


svf/lib/Graphs/VFG.cpp
```diff
@@ -965,12 +965,27 @@ void VFG::updateCallGraph(PointerAnalysis* pta)
     }
 }
 
+/*!
+ * Match arguments for callsite at caller and callee
+ */
+bool VFG::matchArgs(const CallICFGNode* cs, const SVFFunction* callee)
```
yuleisui, line 971: Can we reuse the method in pointer analysis or share it in the SVFUtil?

svf/lib/Graphs/VFG.cpp
```diff
@@ -971,6 +971,10 @@ void VFG::updateCallGraph(PointerAnalysis* pta)
  */
 void VFG::connectCallerAndCallee(const CallICFGNode* callBlockNode, const SVFFunction* callee, VFGEdgeSetTy& edges)
 {
+    /// if the arg size does not match then we do not need to connect this parameter
```
yuleisui, line 974: This comments should go to SVFUtil.

svf/lib/Graphs/VFG.cpp
```diff
@@ -971,6 +971,10 @@ void VFG::updateCallGraph(PointerAnalysis* pta)
...
+    /// unless the callee is a variadic function (the first parameter of variadic function is its paramter number)
+    if(SVFUtil::matchArgs(callBlockNode->getCallSite(), callee) == false)
```
yuleisui, line 976: Can we make this at the assertion point? Otherwise, we may miss some of the caller-callee edges

PR #1116, opened 2023-06-03
fix: support programs built w/ ASan (#1114)

### LINKED ISSUE #1114: Support for C++ virtual methods with AddressSanitizer
I'm running the Andersen pointer analysis on [poppler](https://gitlab.freedesktop.org/poppler/poppler/-/tree/1d23101ccebe14261c6afc024ea14f29d209e760) and it fails this assertion
https://github.com/SVF-tools/SVF/blob/6d52d36cf9ec48584b167ba028fe9f79721f674c/svf-llvm/lib/CHGBuilder.cpp#L102

Apparently it finds a vtable that does not have the expected format. Indeed (one of) the offending global variables looks like:
```
@_ZTV16AnnotCalloutLine = hidden constant { { [4 x i8*] }, [32 x i8] } { { [4 x i8*] } { [4 x i8*] [i8* null, i8* bitcast ({ { i8*, i8* }, [16 x i8] }* @_ZTI16AnnotCalloutLine to i8*), i8* bitcast (void (%class.AnnotCalloutLine*)* @_ZN16A
nnotCalloutLineD1Ev to i8*), i8* bitcast (void (%class.AnnotCalloutLine*)* @_ZN16AnnotCalloutLineD0Ev to i8*)] }, [32 x i8] zeroinitializer }, align 32, !type !67155
```
Note the type `{ { [4 x i8*] }, [32 x i8] }` instead of `{ [4 x i8*] }`.

I'm using LLVM 15 (w/o opaque pointers) and AddressSanitizer.

I was able to make a minimal reproducer (https://godbolt.org/z/x6T711aYh) and the issue seems to be the use of virtual methods and AddressSanitizer. The additional "padding" might be the [redzones added to global variables by AddressSanitizer](https://github.com/llvm/llvm-project/blob/8dfdcc7b7bf66834a761bd8de445840ef68e4d1a/llvm/lib/Transforms/Instrumentation/AddressSanitizer.cpp#L2187-L2189).

Using `-mllvm -asan-globals=0` solves the issue with poppler, but disables AddressSanitizer for global variables.

I could write a PR to fix this. What do you think is the best solution?

acidghost:
Fixes #1114.

### DIFF
--- a/svf-llvm/lib/CHGBuilder.cpp
+++ b/svf-llvm/lib/CHGBuilder.cpp
@@ -90,8 +90,7 @@ void CHGBuilder::buildCHGNodes(const GlobalValue *globalvalue)
 {
     if (LLVMUtil::isValVtbl(globalvalue) && globalvalue->getNumOperands() > 0)
     {
-        const ConstantStruct *vtblStruct = SVFUtil::dyn_cast<ConstantStruct>(globalvalue->getOperand(0));
-        assert(vtblStruct && "Initializer of a vtable not a struct?");
+        const ConstantStruct *vtblStruct = LLVMUtil::getVtblStruct(globalvalue);
         string className = getClassNameFromVtblObj(globalvalue->getName().str());
         if (!chg->getNode(className))
             createNode(className);
@@ -367,9 +366,7 @@ void CHGBuilder::analyzeVTables(const Module &M)
         const GlobalValue *globalvalue = SVFUtil::dyn_cast<const GlobalValue>(&(*I));
         if (LLVMUtil::isValVtbl(globalvalue) && globalvalue->getNumOperands() > 0)
         {
-            const ConstantStruct *vtblStruct =
-                SVFUtil::dyn_cast<ConstantStruct>(globalvalue->getOperand(0));
-            assert(vtblStruct && "Initializer of a vtable not a struct?");
+            const ConstantStruct *vtblStruct = LLVMUtil::getVtblStruct(globalvalue);
 
             string vtblClassName = getClassNameFromVtblObj(globalvalue->getName().str());
             CHNode *node = chg->getNode(vtblClassName);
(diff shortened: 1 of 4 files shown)


svf-llvm/lib/CHGBuilder.cpp
```diff
@@ -86,6 +86,17 @@ void CHGBuilder::buildCHG()
         chg->dump("cha");
 }
 
+inline static const ConstantStruct *maybeGetInnerVtblStruct(const ConstantStruct *vtblStruct)
```
yuleisui, line 89: Can you rename this to be AddressSanVtblStruct?

svf-llvm/lib/CHGBuilder.cpp
```diff
@@ -96,6 +107,8 @@ void CHGBuilder::buildCHGNodes(const GlobalValue *globalvalue)
         if (!chg->getNode(className))
             createNode(className);
 
+        vtblStruct = maybeGetInnerVtblStruct(vtblStruct);
```
yuleisui, line 110: I would suggest that you make an option here to use AddressSanVtblStruct. There are few other places where `const ConstantStruct *vtblStruct = SVFUtil::dyn_cast<ConstantStruct>(globalvalue->getOperand(0));`, you may need to make sure we all call AddressSanVtblStruct. Why not wrap up `const ConstantStruct *vtblStruct = getVtblStruct()`; and put everything in `getVtblStruct`, including your method and the assert.

acidghost:
I'm not entirely sure I've got all the locations where it needs to be replaced with `getVtblStruct`.

yuleisui:
> I'm not entirely sure I've got all the locations where it needs to be replaced with `getVtblStruct`.

Looks to me like all have been replaced. Could you put a comment to explain the function`getVtblStruct `?Maybe give a tiny llvm bc code fragment example with and without ASan. I will merge it after your comments.

acidghost:
It should be ready. Let me know if it needs any further update.

PR #1140, opened 2023-07-04
collect removed functions and erase it from Modules.

bjjwwang:


svf-llvm/lib/LLVMModule.cpp
```diff
@@ -115,6 +115,7 @@ SVFModule* LLVMModuleSet::buildSVFModule(const std::vector<std::string> &moduleN
     LLVMModuleSet* mset = getLLVMModuleSet();
 
     mset->loadModules(moduleNameVec);
+    mset->loadExtModules();
```
yuleisui, line 118: loadExtAPIModules

PR #1141, opened 2023-07-05
Replace extapi.json with extapi.bc

shuangxiangkan:


extapi.cmake
```diff
@@ -0,0 +1,2 @@
+execute_process(COMMAND clang -S -c -Xclang -disable-O0-optnone -fno-discard-value-names -emit-llvm ${PROJECT_SOURCE_DIR}/svf/lib/Util/extapi.c -o ${PROJECT_BINARY_DIR}/include/Util/extapi.bc)
```
yuleisui, line 1: make sure clang you are using is from $LLVMHOME

svf-llvm/include/SVF-LLVM/LLVMModule.h
```diff
@@ -87,6 +87,9 @@ class LLVMModuleSet
     LLVMType2SVFTypeMap LLVMType2SVFType;
     Type2TypeInfoMap Type2TypeInfo;
 
+    FunDeclToDefMapTy AppFuncDef2SVFExternFuncDef;
```
yuleisui, line 90: try to use FunDeclToDefMap in LLVMModule.cpp

svf/lib/Util/extapi.c
```diff
@@ -0,0 +1,983 @@
...
+#include <netdb.h>
+#include <signal.h>
+#include <curses.h>
+#include <langinfo.h>
+#include <sys/types.h>
+#include <unistd.h>
+#include <termios.h>
+
+
+int svf_asprintf_ALLOCHEAPVIAARG0_EXTERN(char **restrict strp, const char *restrict fmt, ...)
```
yuleisui, line 19: extern int svf_asprintf_xxx

make comments to explain the name of a method, including three parts: svf_methodname_methodproperties

/*
EFT_NOOP = 0,        // no effect on pointers
        EFT_ALLOC,           // returns a ptr to a newly allocated object
        EFT_REALLOC,         // like L_A0 if arg0 is a non-null ptr, else ALLOC
        EFT_FREE,            // free memory arg0 and all pointers passing into free function
        EFT_FREE_MULTILEVEL, // any argument with 2-level pointer passing too a free wrapper function e.g., XFree(void**) which frees memory for void* and void**
        EFT_NOSTRUCT_ALLOC,  // like ALLOC but only allocates non-struct data
        EFT_STAT,            // retval points to an unknown static var X
        EFT_STAT2,           // ret -> X -> Y (X, Y - external static vars)
        EFT_L_A0,            // copies arg0, arg1, or arg2 into LHS
        EFT_L_A1,
        EFT_L_A2,
        EFT_L_A8,
        EFT_L_A0__A0R_A1,  // stores arg1 into *arg0 and returns arg0 (currently only for memset)
        EFT_L_A0__A0R_A1R, // copies the data that arg1 points to into the location
        EFT_L_A1__FunPtr,  // obtain the address of a symbol based on the arg1 (char*) and parse a function to LHS (e.g., void *dlsym(void *handle, char *funname);)
        //  arg0 points to; note that several fields may be
        //  copied at once if both point to structs.
        //  Returns arg0.
        EFT_A1R_A0R,      // copies *arg0 into *arg1, with non-ptr return
        EFT_A3R_A1R_NS,   // copies *arg1 into *arg3 (non-struct copy only)
        EFT_A1R_A0,       // stores arg0 into *arg1
        EFT_A2R_A1,       // stores arg1 into *arg2
        EFT_A4R_A1,       // stores arg1 into *arg4
        EFT_L_A0__A2R_A0, // stores arg0 into *arg2 and returns it
        EFT_L_A0__A1_A0,  // store arg1 into arg0's base and returns arg0
        EFT_A0R_NEW,      // stores a pointer to an allocated object in *arg0
        EFT_A1R_NEW,      // as above, into *arg1, etc.
        EFT_A2R_NEW,
        EFT_A4R_NEW,
        EFT_A11R_NEW,
        EFT_STD_RB_TREE_INSERT_AND_REBALANCE, // Some complex effects
        EFT_STD_RB_TREE_INCREMENT,            // Some complex effects
        EFT_STD_LIST_HOOK,                    // Some complex effects

        CPP_EFT_A0R_A1,       // stores arg1 into *arg0
        CPP_EFT_A0R_A1R,      // copies *arg1 into *arg0
        CPP_EFT_A1R,          // load arg1
        EFT_CXA_BEGIN_CATCH,  //__cxa_begin_catch
        CPP_EFT_DYNAMIC_CAST, // dynamic_cast
        EFT_NULL              // not found in the list
*/

svf-llvm/include/SVF-LLVM/LLVMModule.h
```diff
@@ -87,6 +87,9 @@ class LLVMModuleSet
...
+    FunDeclToDefMapTy SVFExternFuncDef2AppFuncDef;
```
yuleisui, line 91: Remove this map. Take a look at createSVFDataStructure in LLVMModule.cpp. If you have two def functions with the same name "foo"  in app and "svf_foo" in ext.c. We will map both llvm functions to the same svf function named "svf_foo" if we would like use "svf_foo" to overwrite "foo"

yuleisui: make sure both llvm function maps to the same svf function in this map: LLVMFunc2SVFFunc

PR #1172, opened 2023-08-21
fix read and write to file err

JasonZhongZexin:
write the gepObjVarMap to the file and load it from the file


svf/lib/MemoryModel/PointerAnalysisImpl.cpp
```diff
@@ -355,6 +371,25 @@ bool BVDataPTAImpl::readFromFile(const string& filename)
...
+        NodeID id;
+        ss >> base >> offset >>id;
+        SVFIR::NodeOffsetMap::const_iterator iter = gepObjVarMap.find(std::make_pair(base, offset));
+        if (iter == gepObjVarMap.end())
+        {
+            pag->readGepObjNodeFromFile(base, offset, id);
+        }
+        
+    }
+
```
JasonZhongZexin, line 392: ln256 - ln266 and ln 393 - ln 408 may be able to be removed since we are now writing and reading all GepObjVarNode to the file via the GepObjVarMap.

svf/lib/Util/NodeIDAllocator.cpp
```diff
@@ -84,6 +84,12 @@ NodeID NodeIDAllocator::allocateObjectId(void)
     return id;
 }
 
+void NodeIDAllocator::updateNumObjAndNodes()
```
yuleisui, line 87: updateNumObjAndNodes => increaseNumOfObjAndNodes

svf/include/Util/NodeIDAllocator.h
```diff
@@ -82,6 +82,7 @@ class NodeIDAllocator
         return numObjects;
     }
 
+    void updateNumObjAndNodes();
```
yuleisui, line 85: inline void updateNumObjAndNodes() =》inline void increaseNumOfObjAndNodes(){
    ++numObjects;
    ++numNodes;
}

PR #1184, opened 2023-09-05
fix getgepoffset and accumulateconstantOffset

bjjwwang:
## support byte size
- add SVFType2ByteSize in LLVMUtil
- add u32 elem_bytesize and bool getByteOffset to support byte size

### DIFF
--- a/svf/include/MemoryModel/AccessPath.h
+++ b/svf/include/MemoryModel/AccessPath.h
@@ -110,6 +110,10 @@ public:
     //@}
 
     /// Return accumulated constant byte offset given OffsetVarVec and elemByteSize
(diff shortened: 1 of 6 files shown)


svf/include/MemoryModel/AccessPath.h
```diff
@@ -110,7 +110,7 @@ class AccessPath
     //@}
 
     /// Return accumulated constant offset given OffsetVarVec
-    APOffset computeConstantOffset() const;
+    APOffset computeConstantOffset(u32_t elem_bytesize, bool getByteOffset) const;
```
yuleisui, line 113: can we make it default parameters

Issue #1186, opened 2023-09-05
ExtAPI: questions about new format


adriaanjacobs:
Hi

We've submitted patches in the past to allow extending the ExtAPI programmatically with new entries to describe allocation wrappers (#821, #1083), and would like to continue updating that functionality to support the new format. To that purpose, we have a few questions about the direction of the new format, and how you would like extensibility to be handled:

* After reading [the wiki](https://github.com/SVF-tools/SVF/wiki/Handling-External-APIs-with-extapi.c), it seems that the `OVERWRITE` attribute will now always clear the function definition in the app module (app_def). This seems different from the previous behaviour of the `OVERWRITE_APP` flag: we used to be able to keep the original definition in the app, while SVF simply ignored the app definition and used the behaviour provided by the extapi database (now extapi.bc, previously extapi.json) to handle it. 
  * Is there a reason for this change? We'd prefer the old way, since we wouldn't have to re-insert the allocation wrapper definitions to actually compile the module.
  * In our understanding, deleting the app body without merging in the extapi definition of the function would result in linker errors by default? 
* We are currently considering to add the option to specify an llvm::Module as the extApiModule instead of a filename, and let the user parse the .bc file themselves. That would allow us to insert our allocation wrappers before SVF ever sees the module. If that's desirable for you, we can send those patches in.
* The ExtAPI class is now fully empty, and all the relevant parts are contained in LLVMModuleSet instead. Could we refactor this back into ExtAPI to maintain more separation? Ideally, we'd like the parsing of extapi.c & the population of the relevant maps to be done separately from building the LLVMModule(Set), so we can analyze the module up front & generate the necessary changes to extapi.c before we ever invoke the "real" SVF machinery. 

As always, happy to send PRs for these changes. Just wanted to confirm whether they are welcome :)

yuleisui:
Your proposal seems good to me. Could you please make a pr?

@shuangxiangkan could you take a look at this? Make any comments if you have.

adriaanjacobs:
Another quick question: it looks like all functions that are annotated with STATIC in extapi.c are also annotated with ALLOC_RET; is that intentional? The aliasing behaviors of those annotations seem contradictory: ALLOC_RET always returns a new object, whereas all calls to a STATIC function should alias.

yuleisui:
Though calling the same static function, it will return different new objects. You could have a try to validate it.

If a function in extapi.c has one or more annotations, the function body will be ignored and the side-effect is based on the annotations.

[...]

PR #1188, opened 2023-09-07
Remove "STATIC" annotation in extapi.bc

shuangxiangkan:


svf-llvm/lib/extapi.c
```diff
@@ -23,291 +22,209 @@ void *SyGetmem(unsigned long size)
     return NULL;
 }
 
-__attribute__((annotate("ALLOC_RET")))
-__attribute__((annotate("STATIC")))
 const unsigned short **__ctype_b_loc(void)
 {
-    return NULL;
+    return malloc(10);
```
yuleisui, line 27: why malloc(10)? better to have a macro rather than a concrete number.

If you remove all STATIC functions, how about static functions mentioned here:
https://github.com/SVF-tools/SVF/issues/1186

svf-llvm/lib/extapi.c
```diff
@@ -1193,3 +902,208 @@ void _ZNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEC1EPKcRKS3_(void **arg
...
+}
+
+char * dngettext(const char * domainname, const char * msgid, const char * msgid_plural, unsigned long int n)
+{
+    return STATIC_OBJECT;
+}
+
+void *fdopen(int fd, const char *mode)
+{
+    return STATIC_OBJECT;
```
adriaanjacobs, line 973: As far as I know, fdopen will (just like fopen & family) return a new object every time it's called. Should this be ALLOC_RET instead?

adriaanjacobs:
Looks good to me. To be clear: this resolves the issue of contradicting ALLOC_RET and STATIC annotations, but as far as I understand, we would still need a change to the format to support some of the more complex extapi behavior discussed in #1186, is that correct?

Regardless, this looks good. The `fdopen` thing has been present for a long(er) time, just never thought to comment on it.

PR #1191, opened 2023-09-08
fix a bug that leads to get many incorrect switch target case value -1

canliture:
When I use SVF to do some research on static (sparse) symbolic execution, e.g., translate SVFG/VFGNode to Z3 Formula. I think the the following code will lead to some incorrect formulas.
 
Given the code:
```c
#include <stdio.h>


void test(int a) {
    switch (a)
    {
        case 0: printf("0\n"); break;
        case 1:
        case 2:
        case 3: printf("a >=1 && a <= 3\n"); break;
        case 4:
        case 6:
        case 7:  printf("a >= 4 && a <=7\n"); break;
        default: printf("a < 0 || a > 7"); break;
    }
}
```
SVF will get many incorrect target case value -1 `due to use of API SwitchInst::findCaseDest`. Here is comments about SwitchInst::findCaseDest:
```c
/// Finds the unique case value for a given successor. Returns null if the
/// successor is not found, not unique, or is the default case.
ConstantInt *findCaseDest(BasicBlock *BB) {
  // ...
}
```
Find the `unique` case value .... .`Return null if ... , not unique ` ...

### DIFF
--- a/svf-llvm/lib/SVFIRBuilder.cpp
+++ b/svf-llvm/lib/SVFIRBuilder.cpp
@@ -946,29 +946,20 @@ void SVFIRBuilder::visitSwitchInst(SwitchInst &inst)
     NodeID cond = getValueNode(inst.getCondition());
 
     BranchStmt::SuccAndCondPairVec successors;
-    for (u32_t i = 0; i < inst.getNumSuccessors(); ++i)
+
+    // get case successor basic block and related case value
+    SuccBBAndCondValPairVec succBB2CondValPairVec;
+    LLVMUtil::getSuccBBandCondValPairVec(inst, succBB2CondValPairVec);
+    for (auto &succBB2CaseValue : succBB2CondValPairVec)
     {
-        BasicBlock *succBB = inst.getSuccessor(i);
-        /// default case is set to -1;
-        s64_t val = -1;
-        /// Equivalent to: succBB != inst.getDefaultDest()
-        if (i != 0)
-        {
-            // According to llvm::SwitchInst class comments:
-            // Operand[0]    = Value to switch on
-            // Operand[1]    = Default basic block destination
-            // Operand[2n  ] = Value to match
-            // Operand[2n+1] = BasicBlock to go to on match
-            ConstantInt *condVal = SVFUtil::dyn_cast<ConstantInt>(inst.getOperand(i * 2));
-            if (condVal && condVal->getBitWidth() <= 64)
-                val = condVal->getSExtValue();
-        }
+        s64_t val = LLVMUtil::getCaseValue(inst, succBB2CaseValue);
+        const BasicBlock *succBB = succBB2CaseValue.first;
         const Instruction* succInst = &succBB->front();
         const SVFInstruction* svfSuccInst = LLVMModuleSet::getLLVMModuleSet()->getSVFInstruction(succInst);
         const ICFGNode* icfgNode = pag->getICFG()->getICFGNode(svfSuccInst);
-        successors.push_back(std::make_pair(icfgNode,val));
+        successors.push_back(std::make_pair(icfgNode, val));
     }
-    addBranchStmt(brinst, cond,successors);
+    addBranchStmt(brinst, cond, successors);
 }
 
 ///   %ap = alloca %struct.va_list
--- a/svf-llvm/lib/SVFIRBuilder.cpp
+++ b/svf-llvm/lib/SVFIRBuilder.cpp
@@ -948,12 +948,22 @@ void SVFIRBuilder::visitSwitchInst(SwitchInst &inst)
     BranchStmt::SuccAndCondPairVec successors;
     for (u32_t i = 0; i < inst.getNumSuccessors(); ++i)
     {
-        const Instruction* succInst = &inst.getSuccessor(i)->front();
-        const ConstantInt* condVal = inst.findCaseDest(inst.getSuccessor(i));
+        BasicBlock *succBB = inst.getSuccessor(i);
         /// default case is set to -1;
         s64_t val = -1;
-        if (condVal && condVal->getBitWidth() <= 64)
-            val = condVal->getSExtValue();
+        /// Equivalent to: succBB != inst.getDefaultDest()
+        if (i != 0)
+        {
+            // According to llvm::SwitchInst class comments:
+            // Operand[0]    = Value to switch on
+            // Operand[1]    = Default basic block destination
+            // Operand[2n  ] = Value to match
+            // Operand[2n+1] = BasicBlock to go to on match
+            ConstantInt *condVal = SVFUtil::dyn_cast<ConstantInt>(inst.getOperand(i * 2));
+            if (condVal && condVal->getBitWidth() <= 64)
+                val = condVal->getSExtValue();
+        }
+        const Instruction* succInst = &succBB->front();
         const SVFInstruction* svfSuccInst = LLVMModuleSet::getLLVMModuleSet()->getSVFInstruction(succInst);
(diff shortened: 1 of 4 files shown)


svf-llvm/lib/SVFIRBuilder.cpp
```diff
@@ -948,12 +948,22 @@ void SVFIRBuilder::visitSwitchInst(SwitchInst &inst)
...
-            val = condVal->getSExtValue();
+        /// Equivalent to: succBB != inst.getDefaultDest()
+        if (i != 0)
+        {
+            // According to llvm::SwitchInst class comments:
+            // Operand[0]    = Value to switch on
+            // Operand[1]    = Default basic block destination
+            // Operand[2n  ] = Value to match
+            // Operand[2n+1] = BasicBlock to go to on match
+            ConstantInt *condVal = SVFUtil::dyn_cast<ConstantInt>(inst.getOperand(i * 2));
```
yuleisui, line 962: what does this mean here, why i * 2? Can we use `findCaseDest` + something else?

canliture: In my mind, 'condVal' is target `case value`. And in LLVM IR:

Operand 0 is `condition of switch`, 
Operand 1 is `default` BasicBlock Value , and 

Operand 2 is target Value of `the first` Case
Operand 3 is target succussor BasicBlock Value of `the first` Case's 
...
Operand 2n      is target Value of `the n-th` Case
Operand 2n+1  is target succussor BasicBlock Value of `the n-th` Case's

<hr/>

Before we use findCaseDest, source code of findCaseDest is below: 
```c
  /// Finds the unique case value for a given successor. Returns null if the
  /// successor is not found, not unique, or is the default case.
  ConstantInt *findCaseDest(BasicBlock *BB) {
    if (BB == getDefaultDest())
      return nullptr;

    ConstantInt *CI = nullptr;
    for (auto Case : cases()) {
      if (Case.getCaseSuccessor() != BB)
        continue;

      if (CI)
        return nullptr; // Multiple cases lead to BB.

      CI = Case.getCaseValue();
    }

    return CI;
  }
```
we will get -1 findCaseDest of default case, and also get -1 when encounter continuous case without break stmt, the LLVM IR related the given code is 
```c
// ... 
  switch i32 %0, label %sw.default [
    i32 0, label %sw.bb                                ; get -1 due to default branch
    i32 1, label %sw.bb1                               ; get -1 without break
    i32 2, label %sw.bb1                              ; get -1  without break
    i32 3, label %sw.bb1                             
    i32 4, label %sw.bb3                             ; get -1 without break
    i32 6, label %sw.bb3                             ; get -1 without break
    i32 7, label %sw.bb3
  ], !dbg !18
// ... 
```

if we use `findCaseDest` + something else, let me thought a few moments. I can't give the answer at once...

canliture:
I give another answer:
```c
void SVFIRBuilder::visitSwitchInst(SwitchInst &inst)
{
    NodeID brinst = getValueNode(&inst);
    NodeID cond = getValueNode(inst.getCondition());

    BranchStmt::SuccAndCondPairVec successors;

    std::vector<std::pair<ConstantInt*, BasicBlock*>> caseValue2SuccBBVec;
    // default case
    caseValue2SuccBBVec.push_back({nullptr, inst.getDefaultDest()});
    // normal case
    for (const SwitchInst::CaseHandle& cs : inst.cases())
    {
        caseValue2SuccBBVec.push_back({cs.getCaseValue(), cs.getCaseSuccessor()});
    }
    for (const auto& caseValue2SuccBB : caseValue2SuccBBVec)
    {
        ConstantInt *caseValue = caseValue2SuccBB.first;
        BasicBlock *succBB = caseValue2SuccBB.second;
        /// default case is set to -1;
        s64_t val = -1;
        if (caseValue != nullptr && succBB != inst.getDefaultDest())
        {
            if (caseValue->getBitWidth() <= 64)
                val = caseValue->getSExtValue();
        }
        const Instruction* succInst = &succBB->front();
        const SVFInstruction* svfSuccInst = LLVMModuleSet::getLLVMModuleSet()->getSVFInstruction(succInst);
        const ICFGNode* icfgNode = pag->getICFG()->getICFGNode(svfSuccInst);
        SVFUtil::outs() << val << ": " << succBB->getName().str() << std::endl;
        successors.push_back(std::make_pair(icfgNode,val));
    }
    addBranchStmt(brinst, cond,successors);
}
```
output is:
```
-1: sw.default
0: sw.bb
1: sw.bb1
2: sw.bb1
3: sw.bb1
4: sw.bb3
6: sw.bb3
7: sw.bb3
```

yuleisui:
> 

This look good but code is quite a lot. Could we try to simplify it and any helper functions or make the naming more explicit and easy understandable

canliture:
> > 
> 
> This look good but code is quite a lot. Could we try to simplify it and any helper functions or make the naming more explicit and easy understandable

Yeah, I think I have make the code more easily understandable by [f691c36](https://github.com/SVF-tools/SVF/pull/1191/commits/f691c36f5e08dfce2e2ea6d84ef83ba9074a232d)

yuleisui:
Thanks and will merge.

yuleisui:
Are there any test cases you could provide here to showcase the difference between current buggy implementation and your fix.

canliture:
> Are there any test cases you could provide here to showcase the difference between current buggy implementation and your fix.

given the testcase:
```
#include <stdio.h>

void test(int a) {
    switch (a)
    {
        case 0: printf("0\n"); break;
        case 1:
        case 2:
        case 3: printf("a >=1 && a <= 3\n"); break;
        case 4:
        case 6:
        case 7:  printf("a >= 4 && a <=7\n"); break;
        default: printf("a < 0 || a > 7"); break;
    }
}
```
generate IR part related switch:
```
...
  switch i32 %0, label %sw.default [
    i32 0, label %sw.bb
    i32 1, label %sw.bb1
    i32 2, label %sw.bb1
    i32 3, label %sw.bb1
    i32 4, label %sw.bb3
    i32 6, label %sw.bb3
    i32 7, label %sw.bb3
  ], !dbg !18
...
```
insert a line code into the original codes here: 
https://github.com/SVF-tools/SVF/blob/0d9f7cf00dfc2075582243f3b2e622efbed57b68/svf-llvm/lib/SVFIRBuilder.cpp#L958-L959
inserted code like this:
```c
const ICFGNode* icfgNode = pag->getICFG()->getICFGNode(svfSuccInst);
SVFUtil::outs() << "case value " << val << ": " << inst.getSuccessor(i)->getName().str() << std::endl;
successors.push_back(std::make_pair(icfgNode,val));
```
run wpa or other tools that will call into SVFIRBuilder::visitSwitchInst to build SVFIR, output like this:
```
case value -1: sw.default
case value 0: sw.bb
case value -1: sw.bb1
case value -1: sw.bb1
case value -1: sw.bb1
case value -1: sw.bb3
case value -1: sw.bb3
case value -1: sw.bb3
```
Apparently, the output is not correct. After applying the patch, and insert code at the same place:
```c
const ICFGNode* icfgNode = pag->getICFG()->getICFGNode(svfSuccInst);
SVFUtil::outs() << "case value " << val << ": " << succBB->getName().str() << std::endl;
successors.push_back(std::make_pair(icfgNode, val));
```
run wpa or other tools that will call into SVFIRBuilder::visitSwitchInst to build SVFIR, new output like this:
```
case value -1: sw.default
case value 0: sw.bb
case value 1: sw.bb1
case value 2: sw.bb1
case value 3: sw.bb1
case value 4: sw.bb3
case value 6: sw.bb3
case value 7: sw.bb3
```
Apparently, the output is correct.

PR #1212, opened 2023-10-05
Fix the extapi.bc path not found issue

shuangxiangkan:


svf/include/Util/ExtAPI.h
```diff
@@ -38,7 +38,7 @@
 
 /// For a more detailed explanation of how External APIs are handled in SVF, please refer to the SVF Wiki: https://github.com/SVF-tools/SVF/wiki/Handling-External-APIs-with-extapi.c
 
-#define EXTAPI_BC_PATH "Release-build/svf-llvm/extapi.bc"
+#define DEFUALT_EXTAPI_BC_PATH "/svf-llvm/extapi.bc" // Default path to extapi.bc through the SVF build method(e.g. source ./build.sh)
```
251, line 41: Typo: "DEFUALT"

svf/lib/Util/ExtAPI.cpp
```diff
@@ -80,55 +88,77 @@ static std::string GetStdoutFromCommand(const std::string& command)
...
-    bcFilePath = getFilePath("npm root");
-    if (!stat(bcFilePath.c_str(), &statbuf))
-        return bcFilePath;
-
-    SVFUtil::errs() << "No extapi.bc found at " << bcFilePath << " for getExtAPI(); The default path for extapi.bc is: SVF_IR_PATH/CMAKE_BUILD_TYPE-build/svf-llvm/extapi.bc !\n";
+    struct stat statbuf;
+    // 1. Set extapi.bc path through setExtBcPath() method
+    if (!extBcPath.empty() && !stat(extBcPath.c_str(), &statbuf))
+        return extBcPath;
+
```
251, line 130: You could simplify that function and make it much faster:

1. add a short circuit in the beginning:
```C
if (!extBcPath.empty())
    return extBcPath;
```

2. just assemble the paths via getenv, getFilePath, ... as below but than forward them to `setExtBcPath` and do the stat call there once!

This should also remove the code duplication.

PR #1227, opened 2023-10-16
Split callsite and return site in a CFBBNode

jumormt:
- Split callsite and return site in a CFBBNode
- WTO is updated based on the new structure of CFBBG


svf/include/Graphs/CFBasicBlockG.h
```diff
@@ -493,6 +498,9 @@ class CFBasicBlockGraph : public GenericCFBasicBlockGTy
 class CFBasicBlockGBuilder
 {
 
+public:
+    typedef Map<const SVFBasicBlock*, std::vector<CFBasicBlockNode*>> SVFBBToCFBBNodes;
```
yuleisui, line 502: Since the input of CFBasicBlock is ICFG,  we need to remove "SVFBasicBlock" from CFBasicBlockG. Clean up code and remove unnecessary data structures of SVF.

PR #1238, opened 2023-10-31
Suppress warnings for unused variables when SVF_ENABLE_ASSERTIONS is …

hjjandy:
…not defined.

When SVF_ENABLE_ASSERTIONS is not defined when running cmake, assert in two .h files will be removed, causing unused variable warning. It then leads to compilation errors due to "-Werror -Wunused-variable".

Fixing it by replacing variables in assert with the target expressions. For two functions that cannot be fixed in this way, i.e., CFBasicBlockG.h/add CFBBEdge() and CDG.h/addCDGEDge(), put the two variables in the return statements.


yuleisui:
Thanks for the fix!

PR #1240, opened 2023-11-07
fix getByteOffsets in GepStmt and SVFIR2ItvExeState

bjjwwang:
- add getLLVMByteSize in SVFType
- fix getByteOffset in SVFIR2ExeState and GepStmt

### DIFF
--- a/svf/include/MemoryModel/AccessPath.h
+++ b/svf/include/MemoryModel/AccessPath.h
@@ -109,12 +109,31 @@ public:
     }
     //@}
 
-    /// Return accumulated constant byte offset given OffsetVarVec
-    /// e.g. GepStmt* gep = [i32*4]*, 2
+    /**
+     * Computes the total constant byte offset of an access path.
+     * This function iterates over the offset-variable-type pairs in reverse order,
+     * accumulating the total byte offset for constant offsets. For each pair,
+     * it retrieves the corresponding SVFValue and determines the type of offset
+     * (whether it's an array, pointer, or structure). If the offset corresponds
+     * to a structure, it further resolves the actual element type based on the
+     * offset value. It then multiplies the offset value by the size of the type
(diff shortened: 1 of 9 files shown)


svf/lib/SVFIR/SVFType.cpp
```diff
@@ -4,6 +4,11 @@
 namespace SVF
 {
 
+__attribute__((weak))
+u32_t SVFType::getLLVMByteSize() const {
+    return 0;
```
yuleisui, line 9: Put an assertion to mention implementation and add an abort() method. See SVFValue::toString()

svf/include/MemoryModel/AccessPath.h
```diff
@@ -109,12 +109,11 @@ class AccessPath
...
-    /// elemBytesize is the element byte size of an static alloc or heap alloc array
-    /// e.g. GepStmt* gep = **,
-    /// s32_t elemBytesize = LLVMUtil::SVFType2ByteSize(gep->getRHSVar()->getValue()->getType());
-    /// APOffset byteOffset = gep->accumulateConstantByteOffset(elemBytesize);
-    APOffset computeConstantByteOffset(u32_t elemBytesize) const;
+    /// Return accumulated constant byte offset given OffsetVarVec
+    /// e.g. GepStmt* gep = [i32*4]*, 2
+    /// APOffset byteOffset = gep->accumulateConstantByteOffset();
+    /// byteOffset should be 8 since i32 is 4 bytes and index is 2.
+    APOffset computeConstantByteOffset() const;
```
yuleisui, line 116: Need to explain a bit more about this method since the argument has been removed.

yuleisui: I guess the argument is no longer needed.

Issue #12, opened 2017-03-20
Assertion triggered after modifying code


ktran:
Hello!

I'm currently using the SVF analysis for a project and I run into an issue using the analysis
for newly inserted instructions. In short, I get the following error whenever I try to check
whether the newly inserted instruction (AllocaInst) aliases with another value :

SVF/include/MemoryModel/MemModel.h:554:
SymID SymbolTableInfo::getValSym(const llvm::Value*): Assertion `iter!=valSymMap.end() &&"value sym not found"' failed.

Below is a test pass that should throw this error for any loop containing a load.
Weirdly, this **only** happens if I compile Svf & the pass in Debug mode, but not if I build it in
MinSizeRel mode.

In general I believe that I need to update the pointer analysis after having modified the code, however, I'm not sure how to properly update the analysis. The test pass below contains some of my failed trials of naively re-running it (see commented lines).

What I am doing wrong? How do I properly update/re-run the analysis? And, do you know why this is
only happening if compiled in Debug mode?

```c++
#include "llvm/Analysis/LoopPass.h"
#include "llvm/Support/raw_ostream.h"
#include "llvm/IR/IRBuilder.h"

#include "MemoryModel/PointerAnalysis.h"
#include "WPA/Andersen.h"

using namespace llvm;

namespace {
  struct TestSvf : public LoopPass {
    static char ID;

    TestSvf() : LoopPass(ID) {}

    virtual void getAnalysisUsage(AnalysisUsage &AU) const {
      AU.addRequired<LoopInfoWrapperPass>();
    }
    virtual bool runOnLoop(Loop *L, LPPassManager &LPM);
  };
}

bool TestSvf::runOnLoop(Loop *L, LPPassManager &LPM) {
  // Create AA before changing loop
  Module &M = *L->getHeader()->getParent()->getParent();
  AndersenWaveDiff *AA = AndersenWaveDiff::createAndersenWaveDiff(M);

  // Change loop: add an alloca instruction
  BasicBlock *H = L->getHeader();
  IRBuilder<> Builder(&*(H->getFirstInsertionPt()));
  AllocaInst *Alloca = Builder.CreateAlloca(Type::getInt1Ty(getGlobalContext()), 0, "test_alloca");

  // Try to recalculate AA
  // 1st try: AA->analyze(M);

  // 2nd try:
  // AndersenWaveDiff::releaseAndersenWaveDiff();
  // AA = AndersenWaveDiff::createAndersenWaveDiff(M);

[...]

Issue #11, opened 2017-02-27
Buffer index sensitivity with non variant GEP instructions


davidtr1037:
In the following code:
`
int f() {
  char buf[10] = {0,};
  int x = buf[0];
  int y = buf[1];
  return x + y;
}
`
The GEP instructions which correspond to x and y are both constant,
but when we check aliasing, we get MayAlias.
Actually, in both cases, the LocationSet of the GepObjPN is constant and equal to 0.
Is this a bug or SVF handles buffers always in an index insensitive way?

rockysui:
Hi David,

Variant GEP is used to handle pointer arithmetic when iterating fields of a struct.
Given ”p = q + i", if q points to an object o, then p may point any of o's fields.

Arrays (like your case) are always handled insensitively in the default memory model.

I believe you can use the following option to get what you wanted:

wpa -ander -locMM xx.bc

"-locMM" uses our fine-grained location-based memory model to distinguish constant array indexes. (see link https://github.com/unsw-corg/SVF/blob/master/include/MemoryModel/LocMemModel.h)

Note that "-locMM" is not the default option as it may be very costly and unscalable for large-size programs.

PR #1262, opened 2023-11-29
fix a bug: 'getExitBB' of SVFFunction may get incorrect exit block.

canliture:
reproduce the bug:
given the code in file bb.c:
```c
int print();

void test()
{
    for (int i = 0 ; i < 100 ; i++)
    {
        print();
    }
}
```
compile and optimize the code with the following commands:
```shell
clang -S -c -Xclang -disable-O0-optnone -fno-discard-value-names -emit-llvm -o bb.ll bb.c
opt -loop-unroll -unroll-count=3 -S -o bb.ll.ll bb.ll
```
we get the final optimized bitcode file `bb.ll.ll` and get the code of test function:
```c
; Function Attrs: noinline nounwind ssp uwtable
define void @test() #0 {
entry:
  %i = alloca i32, align 4
  store i32 0, i32* %i, align 4
  br label %for.cond

for.cond:                                         ; preds = %for.inc.2, %entry
  %0 = load i32, i32* %i, align 4
  %cmp = icmp slt i32 %0, 100
  br i1 %cmp, label %for.body, label %for.end

for.body:                                         ; preds = %for.cond
  %call = call i32 (...) @print()
  br label %for.inc

for.inc:                                          ; preds = %for.body
  %1 = load i32, i32* %i, align 4
  %inc = add nsw i32 %1, 1
  store i32 %inc, i32* %i, align 4
  %2 = load i32, i32* %i, align 4
  %cmp.1 = icmp slt i32 %2, 100
  br i1 %cmp.1, label %for.body.1, label %for.end

for.end:                                          ; preds = %for.inc.1, %for.inc, %for.cond
  ret void

for.body.1:                                       ; preds = %for.inc
  %call.1 = call i32 (...) @print()
  br label %for.inc.1

for.inc.1:                                        ; preds = %for.body.1
  %3 = load i32, i32* %i, align 4
  %inc.1 = add nsw i32 %3, 1
  store i32 %inc.1, i32* %i, align 4
  %4 = load i32, i32* %i, align 4
  %cmp.2 = icmp slt i32 %4, 100
  br i1 %cmp.2, label %for.body.2, label %for.end

for.body.2:                                       ; preds = %for.inc.1
  %call.2 = call i32 (...) @print()
  br label %for.inc.2

for.inc.2:                                        ; preds = %for.body.2
  %5 = load i32, i32* %i, align 4
  %inc.2 = add nsw i32 %5, 1
  store i32 %inc.2, i32* %i, align 4
  br label %for.cond, !llvm.loop !6
}
```
so, the `basic block vector` is:
```c
[
  entry,
  for.cond,
  for.body,
  for.inc,
  for.end,
  for.body.1,
  for.inc.1,
  for.body.2,
  for.inc.2,
]
```
in the code above, we know: `for.end` basic block is the exit basic block of the function test, so I think we can't  get exit basic block by simply retrieving `the last element` of the `basic block vector`

by pulling request, I unify to set entry and exit basic block according to the definition of the entry/exit basic block of a function: no preds/succs.

### DIFF
--- a/svf-llvm/lib/LLVMModule.cpp
+++ b/svf-llvm/lib/LLVMModule.cpp
@@ -321,14 +321,11 @@ void LLVMModuleSet::initSVFBasicBlock(const Function* func)
             svfbb->addPredBasicBlock(svf_pred_bb);
         }
 
-        // set exit block
-        if (svfbb->getSuccessors().empty() && !bb->getInstList().empty())
+        /// set exit block
+        if (svfbb->getSuccessors().empty())
         {
-            const Instruction &lastInst = bb->getInstList().back();
-            if (SVFUtil::dyn_cast<ReturnInst>(&lastInst))
-            {
-                svfFun->setExitBlock(svfbb);
-            }
+            assert(LLVMUtil::basicBlockHasRet(bb) && "exit basic block must have no successors and have a return instruction");
+            svfFun->setExitBlock(svfbb);
         }
 
         for (BasicBlock::const_iterator iit = bb->begin(), eiit = bb->end(); iit != eiit; ++iit)
--- a/svf-llvm/lib/LLVMModule.cpp
+++ b/svf-llvm/lib/LLVMModule.cpp
@@ -321,12 +321,11 @@ void LLVMModuleSet::initSVFBasicBlock(const Function* func)
             svfbb->addPredBasicBlock(svf_pred_bb);
         }
 
-        /// set exit block
+        /// set exit block: exit basic block must have no successors and have a return instruction
         if (svfbb->getSuccessors().empty())
         {
-            if (LLVMUtil::basicBlockHasRet(bb))
+            if (LLVMUtil::basicBlockHasRetInst(bb))
             {
-                // exit basic block must have no successors and have a return instruction
                 svfFun->setExitBlock(svfbb);
             }
         }
--- a/svf-llvm/lib/LLVMModule.cpp
+++ b/svf-llvm/lib/LLVMModule.cpp
@@ -321,14 +321,14 @@ void LLVMModuleSet::initSVFBasicBlock(const Function* func)
             svfbb->addPredBasicBlock(svf_pred_bb);
         }
 
-        // set entry / exit block
-        if (svfbb->getPredecessors().empty())
+        // set exit block
+        if (svfbb->getSuccessors().empty() && !bb->getInstList().empty())
         {
-            svfFun->setEntryBlock(svfbb);
-        }
-        if (svfbb->getSuccessors().empty())
-        {
-            svfFun->setExitBlock(svfbb);
+            const Instruction &lastInst = bb->getInstList().back();
+            if (SVFUtil::dyn_cast<ReturnInst>(&lastInst))
+            {
+                svfFun->setExitBlock(svfbb);
+            }
         }
 
         for (BasicBlock::const_iterator iit = bb->begin(), eiit = bb->end(); iit != eiit; ++iit)
--- a/svf-llvm/lib/LLVMModule.cpp
+++ b/svf-llvm/lib/LLVMModule.cpp
@@ -324,8 +324,11 @@ void LLVMModuleSet::initSVFBasicBlock(const Function* func)
         /// set exit block
         if (svfbb->getSuccessors().empty())
         {
-            assert(LLVMUtil::basicBlockHasRet(bb) && "exit basic block must have no successors and have a return instruction");
-            svfFun->setExitBlock(svfbb);
+            if (LLVMUtil::basicBlockHasRet(bb))
+            {
+                // exit basic block must have no successors and have a return instruction
+                svfFun->setExitBlock(svfbb);
+            }
         }
 
         for (BasicBlock::const_iterator iit = bb->begin(), eiit = bb->end(); iit != eiit; ++iit)
--- a/svf-llvm/lib/LLVMModule.cpp
+++ b/svf-llvm/lib/LLVMModule.cpp
@@ -305,6 +305,7 @@ void LLVMModuleSet::initSVFFunction()
 
 void LLVMModuleSet::initSVFBasicBlock(const Function* func)
 {
+    SVFFunction *svfFun = getSVFFunction(func);
     for (Function::const_iterator bit = func->begin(), ebit = func->end(); bit != ebit; ++bit)
     {
         const BasicBlock* bb = &*bit;
@@ -319,6 +320,17 @@ void LLVMModuleSet::initSVFBasicBlock(const Function* func)
             const SVFBasicBlock* svf_pred_bb = getSVFBasicBlock(*pred_it);
             svfbb->addPredBasicBlock(svf_pred_bb);
         }
+
+        // set entry / exit block
+        if (svfbb->getPredecessors().empty())
+        {
+            svfFun->setEntryBlock(svfbb);
+        }
+        if (svfbb->getSuccessors().empty())
+        {
+            svfFun->setExitBlock(svfbb);
+        }
+
         for (BasicBlock::const_iterator iit = bb->begin(), eiit = bb->end(); iit != eiit; ++iit)
         {
             const Instruction* inst = &*iit;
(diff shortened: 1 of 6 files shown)


canliture:
maybe multi exit blocks...

yuleisui:
> maybe multi exit blocks leads to assert crashes

Could you help fix this?

canliture:
> > maybe multi exit blocks leads to assert crashes
> 
> Could you help fix this?

Yes, I currently work on this.

jumormt:
Hi @canliture , which LLVM version are you using to compile this case?

svf/include/SVFIR/SVFValue.h
```diff
@@ -320,6 +320,15 @@ class SVFFunction : public SVFValue
     std::vector<const SVFBasicBlock*> allBBs;   /// all BasicBlocks of this function
     std::vector<const SVFArgument*> allArgs;    /// all formal arguments of this function
     std::vector<std::string> annotations; /// annotations of this function
+    SVFBasicBlock *entryBlock;                  /// entry BasicBlock of this function
```
jumormt, line 323: I think we don't need an entryBlock because it is always the first block in allBBs. This is consistent with LLVM's API called getEntryBlock(), which simply returns the front of its basic blocks.

svf-llvm/lib/LLVMModule.cpp
```diff
@@ -319,6 +320,17 @@ void LLVMModuleSet::initSVFBasicBlock(const Function* func)
...
         }
+
+        // set entry / exit block
+        if (svfbb->getPredecessors().empty())
+        {
+            svfFun->setEntryBlock(svfbb);
+        }
+        if (svfbb->getSuccessors().empty())
+        {
+            svfFun->setExitBlock(svfbb);
```
jumormt, line 331: We should consider the unreachable blocks introduced by instructions such as exit(0). These blocks are not function exit blocks and cannot be connected to the unique exit block. A valid function exit should end with a return instruction.

jumormt: To my knowledge,  each llvm function can only have one return instruction. This means there should be only one function exit block? In that way, no extra unique function exit block is needed.

canliture:
> Hi @canliture , which LLVM version are you using to compile this case?

I use opt command of llvm 13
```
opt -loop-unroll -unroll-count=3 -S -o bb.ll.ll bb.ll
```

svf/include/SVFIR/SVFValue.h
canliture: Thank you for your review. Yeah, here is redundant.

svf-llvm/lib/LLVMModule.cpp
canliture: Yeah, we do not need an extra unique exit block `theoretically`. 

I would like to improve robustness of SVF: when someone generate a `valid` bitcode file like above, we should handle it correctly.

And I have a question, why my first pr can't pass the testcases. I guess the following assert leads to fail to pass the testcase (`maybe multi exit basic block`)
```
inline void setExitBlock(SVFBasicBlock *bb)
{
    assert(!exitBlock && "function already has a exit Basicblock.");
    exitBlock = bb;
}
```

Because I have no testcases, so I can only guess the reason...

canliture:
I find another testcase here:

https://github.com/SVF-tools/Test-Suite/blob/ac3865cafc6f5ab3af86e5e923aa4f1462704583/test_cases_bc/basic_cpp_tests/vector-4.cpp.bc#L339

jumormt:
> I find another testcase here:
> 
> https://github.com/SVF-tools/Test-Suite/blob/ac3865cafc6f5ab3af86e5e923aa4f1462704583/test_cases_bc/basic_cpp_tests/vector-4.cpp.bc#L339

Thank you for your time. That's a good catch! It is not rigorous to return the last block as the exit block. 

It seems that some cases fail the CI because of no return instructions residing in the function (e.g., https://github.com/SVF-tools/Test-Suite/blob/ac3865cafc6f5ab3af86e5e923aa4f1462704583/test_cases_bc/basic_cpp_tests/deque-1.cpp.bc#L981).

jumormt:
Confirming that all 27 failed cases are caused by the Assertion `exitBlock && "have not yet set exit Basicblock?"'. This happens when analyzing the following functions:

define linkonce_odr hidden void @__clang_call_terminate(i8* %0) #6 comdat {
  %2 = call i8* @__cxa_begin_catch(i8* %0) #11
  call void @_ZSt9terminatev() #12
  unreachable
}

I think there should be no exit BBs for this function. Also, there should be no return edge from the callee back to the caller. Therefore, I suppose that the assertion in getExitBB() for checking null for exitBB can be removed.

yuleisui:
Need to revise the assert to be: 
  
assert((function.hasReturn() && exitBlock) && "have not yet set exit Basicblock?");

svf-llvm/lib/LLVMModule.cpp
```diff
@@ -319,6 +320,17 @@ void LLVMModuleSet::initSVFBasicBlock(const Function* func)
             const SVFBasicBlock* svf_pred_bb = getSVFBasicBlock(*pred_it);
             svfbb->addPredBasicBlock(svf_pred_bb);
         }
+
+        // set exit block
+        if (svfbb->getSuccessors().empty() && !bb->getInstList().empty())
```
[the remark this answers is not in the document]
canliture: Thanks for review. 

Yeah, an exit basic block is a 'single' basic block in a function:
1. having no successors 
2. containing return instruction in a function, and  
3. ret instruction is not neccessarily the last instruction in the basic block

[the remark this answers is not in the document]
canliture: > Looks to me the crests are failed. Please help fix them.

Yes, I have fixed all the test failures.

canliture:
I think maybe there are some potential crashes of all callers of getExitBB().

But we have guard assert in getExitBB(), these crashes may be left for more real projects test, but currently, It's very nice.😄

svf-llvm/lib/LLVMModule.cpp
```diff
@@ -319,6 +320,17 @@ void LLVMModuleSet::initSVFBasicBlock(const Function* func)
             const SVFBasicBlock* svf_pred_bb = getSVFBasicBlock(*pred_it);
             svfbb->addPredBasicBlock(svf_pred_bb);
         }
+
+        /// set exit block
+        if (svfbb->getSuccessors().empty())
+        {
+            if (LLVMUtil::basicBlockHasRet(bb))
```
yuleisui, line 327: basicBlockHasRet=>basicBlockHasRetInst

svf-llvm/lib/LLVMModule.cpp
```diff
@@ -319,6 +320,17 @@ void LLVMModuleSet::initSVFBasicBlock(const Function* func)
...
+            {
+                // exit basic block must have no successors and have a return instruction
```
yuleisui, line 329: move this comment to lin324

svf/include/SVFIR/SVFValue.h
```diff
@@ -458,6 +468,11 @@ class SVFFunction : public SVFValue
         return isNotRet;
     }
 
+    inline bool hasReturn() const
```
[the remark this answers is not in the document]
yuleisui: could you keep hasReturn and remove the other one?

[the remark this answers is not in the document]
canliture: Thanks for reviewing. Yeah, I think so.

svf/lib/Graphs/ICFG.cpp
```diff
@@ -54,7 +54,11 @@ FunExitICFGNode::FunExitICFGNode(NodeID id, const SVFFunction* f)
     // if function is implemented
     if (f->begin() != f->end())
     {
-        bb = f->getExitBB();
+        // ensure the enclosing function has exit basic block
+        if (!f->isNotRetFunction())
```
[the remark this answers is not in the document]
canliture: sorry for late to commit. 😂 now Is it necessary to commit?

PR #1280, opened 2023-12-12
merge the isStrongUpdate in to SVFUtil

jumormt:


svf/lib/Util/SVFUtil.cpp
```diff
@@ -324,3 +324,22 @@ bool SVFUtil::matchArgs(const SVFInstruction* cs, const SVFFunction* callee)
     else
         return getSVFCallSite(cs).arg_size() == callee->arg_size();
 }
+
+bool SVFUtil::isStrongUpdate(const PointsTo &dstCPSet, NodeID &singleton, BVDataPTAImpl *pta) {
```
yuleisui, line 328: Put this method in pointeranalysis class and rename it to be ‘isSingleton’ with only one argument pts

Issue #1279, opened 2023-12-11
SVFG does not use indirect call edges to join formal and actual parameters


acidghost:
I'm trying to use SVFG to compute a def-use graph in order to compute a thin backward interprocedural slice from a `Value`.

The following case produces an SVFG as I would expect:
```c
#include <stddef.h>
#include <stdint.h>

int myFunc(int A) {
  // target
  return A + 42;
}

int LLVMFuzzerTestOneInput(const uint8_t *Data, size_t Size) {
  if (Size < 1)
    return 1;
  int Type = Data[0];

  myFunc(Type);

  return 0;
}
```

The SVFG for the above example clearly shows a path from, e.g., `LoadSVFGNode ID 18` (bottom left in figure) to the 0th argument of `LLVMFuzzerTestOneInput`: ![svfg-direct](https://github.com/SVF-tools/SVF/assets/1787979/8898da10-ee66-4595-99bd-34ddfab34ceb)

If I instead call the function indirectly:
```c
  // myFunc(Type);
  int (*Fn)(int) = myFunc;
  Fn(Type);
```

Then the SVFG for this version does not link the actual and formal parameter of the indirectly called function: 
![svfg-indir](https://github.com/SVF-tools/SVF/assets/1787979/a18fd1f7-5ef9-4650-91cf-da43fc426ea1)

I verified that the indirect call gets resolved correctly by PTA. What am I missing?

Bitcode: https://godbolt.org/z/rjM9ban1r.

The code I'm using to build the SVFG is the same as the example pass.

yuleisui:
I think you might have to update the svfg by calling its ‘updatecallgraph’ method to connect indirect edges

yuleisui:
Could you change ‘ LLVMFuzzerTestOneInput’ to be ‘main’?

acidghost:
I tried with the following but the result is the same.

```c
#include <stddef.h>
#include <stdint.h>

int myFunc(int A) {
  // target
  return A + 42;
}

// int LLVMFuzzerTestOneInput(const uint8_t *Data, size_t Size) {
int main(int argc, char **argv) {
  // if (Size < 1)
  //   return 1;
  // int Type = Data[0];
  if (argc < 2)
    return 1;
  int Type = argv[1][0];

  // myFunc(Type);

  int (*Fn)(int) = myFunc;
  Fn(Type);

  return 0;
}
```

yuleisui:
[...]

PR #1282, opened 2023-12-12
(fix issue #1279) connect actual to formal vfgnode for full svfg

jumormt:
connect actual to formal vfgnode for full svfg


jumormt:
fix issue #1279

PR #1291, opened 2023-12-19
Add `-fno-rtti`/`fno-exceptions` based on LLVM's configuration

Johanmyst:
Currently the top-level `CMakeLists.txt` script force-adds the flags `-fno-rtti` and `-fno-exceptions` using the `add_compile_options()` function from CMake; adding the flags for all modules built thereafter. This causes SVF to disregard LLVM's configuration and always build `SvfLLVM` without RTTI or exceptions.

This pull request moves the loading of LLVM to the top-level configuration file, where `-fno-rtti` and `-fno-exceptions` are added for both `SvfCore` and `SvfLLVM` iff the LLVM instance being used is configured that way.

Additionally, the way the compile definitions now match the way recommended by LLVM for including LLVM from a `CMakeLists.txt` script.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,15 +5,77 @@ project("SVF")
 configure_file(${PROJECT_SOURCE_DIR}/.config.in
                ${PROJECT_BINARY_DIR}/include/Util/config.h)
 
-# We need to match the build environment for LLVM: In particular, we need C++14
-# and the -fno-rtti flag
-set(CMAKE_CXX_STANDARD 14)
-set(CMAKE_CXX_STANDARD_REQUIRED ON)
-# add -std=gnu++14
-set(CMAKE_CXX_EXTENSIONS ON)
+# Match configuration used to build LLVM (match C++ standard; match
+# runtime typing information (RTTI); match exception handling; etc)
+if(COMMAND add_llvm_library)
+  message(STATUS "Detected in-tree build configuration; skipping LLVM fetching")
+  set(IN_SOURCE_BUILD 1)
+else()
+  message(STATUS "Detected out-of-tree build configuration; fetching LLVM")
 
-add_compile_options("-fno-rtti")
-add_compile_options("-fno-exceptions")
+  find_package(LLVM REQUIRED CONFIG HINTS ${LLVM_DIR} $ENV{LLVM_DIR})
+  message(STATUS "LLVM STATUS:
+    Version       ${LLVM_PACKAGE_VERSION}
+    Definitions   ${LLVM_DEFINITIONS}
+    Includes      ${LLVM_INCLUDE_DIRS}
+    Libraries     ${LLVM_LIBRARY_DIRS}
+    Targets       ${LLVM_TARGETS_TO_BUILD}
+    Build type    ${LLVM_BUILD_TYPE}
+    Exceptions    ${LLVM_ENABLE_EH}
+    RTTI          ${LLVM_ENABLE_RTTI}
+    Dynamic lib   ${LLVM_LINK_LLVM_DYLIB}"
+  )
+
+  if("${LLVM_PACKAGE_VERSION_MAJOR}" VERSION_GREATER "15")
+    message(FATAL_ERROR "Unsupported LLVM version (need <= 15.0.7; got ${LLVM_PACKAGE_VERSION})")
+  endif()
+
+  if(CMAKE_BUILD_TYPE STREQUAL "Debug" AND NOT ${LLVM_BUILD_TYPE} STREQUAL "Debug")
+    message(FATAL_ERROR "Got build type 'Debug' but LLVM is not built in Debug mode!")
+  endif()
+
+  if(NOT LLVM_ENABLE_EH)
+    message(STATUS "Building SVF without exception handling")
+    add_compile_options("-fno-exceptions")
+  endif()
+
+  if(NOT LLVM_ENABLE_RTTI)
+    message(STATUS "Building SVF without RTII")
+    add_compile_options("-fno-rtti")
+  endif()
+
+  # Load the LLVM definitions & include directories
+  separate_arguments(LLVM_DEFINITIONS_LIST NATIVE_COMMAND ${LLVM_DEFINITIONS})
+
+  if(NOT ${LLVM_LINK_LLVM_DYLIB})
+    message(STATUS "Linking to separate LLVM static libraries")
+    llvm_map_components_to_libnames(
+      llvm_libs
+      analysis
+      bitwriter
+      core
+      instcombine
+      instrumentation
+      ipo
+      irreader
+      linker
+      scalaropts
+      support
+      target
+      transformutils)
+  else()
+    message(STATUS "Linking to LLVM dynamic shared library object")
+    set(llvm_libs LLVM)
+  endif()
+
+  # Make the "add_llvm_library()" command available
+  include(AddLLVM)
+endif()
+
+# Set C++ standard; use C++14 as LLVM 15 defaults to that (doesn't need to match)
+set(CMAKE_CXX_STANDARD 14)
+set(CMAKE_CXX_EXTENSIONS ON)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
 
 # Treat compiler warnings as errors
 add_compile_options("-Werror" "-Wall")
@@ -80,11 +142,18 @@ endif()
 add_subdirectory(svf)
 add_subdirectory(svf-llvm)
 
+include(GNUInstallDirs)
+install(
+  TARGETS SvfCore SvfLLVM
+  RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR}
+  LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
+  ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}
+)
 install(
   DIRECTORY ${PROJECT_SOURCE_DIR}/svf/include/
             ${PROJECT_SOURCE_DIR}/svf-llvm/include/
   COMPONENT devel
-  DESTINATION include/svf
+  DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/include/svf
   FILES_MATCHING
   PATTERN "**/*.h")
 
--- a/svf-llvm/CMakeLists.txt
+++ b/svf-llvm/CMakeLists.txt
@@ -1,56 +1,12 @@
-# To support both in- and out-of-source builds, we check for the presence of the
-# add_llvm_loadable_module command. - if this command is not present, we are
-# building out-of-source
-if(NOT COMMAND add_llvm_library)
-  find_package(LLVM REQUIRED CONFIG HINTS "${LLVM_DIR}")
-  set(LLVM_DIR ${LLVM_BINARY_DIR} PARENT_SCOPE)
-
-  message(STATUS "Found LLVM ${LLVM_PACKAGE_VERSION}")
-  message(STATUS "Using LLVMConfig.cmake in: ${LLVM_DIR}")
-
-  if(NOT LLVM_ENABLE_RTTI)
-    add_compile_options("-fno-rtti")
-    message(STATUS "Disable RTTI")
-  endif()
-
-  if(NOT LLVM_ENABLE_EH)
-    add_compile_options("-fno-exceptions")
-    message(STATUS "Disable exceptions")
-  endif()
-
-  list(APPEND CMAKE_MODULE_PATH "${LLVM_CMAKE_DIR}")
-  include(AddLLVM)
-
-  add_definitions(${LLVM_DEFINITIONS})
-  # include_directories(SYSTEM ${LLVM_INCLUDE_DIRS})
-
-  if(LLVM_LINK_LLVM_DYLIB)
-    set(llvm_libs LLVM)
-  else()
-    llvm_map_components_to_libnames(
-      llvm_libs
-      analysis
-      bitwriter
-      core
-      instcombine
-      instrumentation
-      ipo
-      irreader
-      linker
-      scalaropts
-      support
-      target
-      transformutils)
-  endif()
-else()
-  set(IN_SOURCE_BUILD 1)
-endif()
-
 # SVF-LLVM contains LLVM Libs
 file(GLOB SVFLLVM_SOURCES lib/*.cpp)
 
 add_llvm_library(SvfLLVM ${SVFLLVM_SOURCES})
+
 target_include_directories(SvfLLVM SYSTEM PUBLIC ${LLVM_INCLUDE_DIRS})
+target_link_directories(SvfLLVM PUBLIC ${LLVM_LIBRARY_DIRS})
+target_compile_definitions(SvfLLVM PUBLIC ${LLVM_DEFINITIONS})
+
 target_include_directories(SvfLLVM PUBLIC include)
 target_link_libraries(SvfLLVM PUBLIC ${llvm_libs} ${Z3_LIBRARIES} SvfCore)
 


yuleisui:
There seems a build error here. https://github.com/SVF-tools/SVF/actions/runs/7263779025/job/19806338524?pr=1291#step:6:197

CMakeLists.txt
```diff
@@ -5,15 +5,77 @@ project("SVF")
...
+    RTTI          ${LLVM_ENABLE_RTTI}
+    Dynamic lib   ${LLVM_LINK_LLVM_DYLIB}"
+  )
+
+  if("${LLVM_PACKAGE_VERSION_MAJOR}" VERSION_GREATER "15")
+    message(FATAL_ERROR "Unsupported LLVM version (need <= 15.0.7; got ${LLVM_PACKAGE_VERSION})")
+  endif()
+
+  if(CMAKE_BUILD_TYPE STREQUAL "Debug" AND NOT ${LLVM_BUILD_TYPE} STREQUAL "Debug")
+    message(FATAL_ERROR "Got build type 'Debug' but LLVM is not built in Debug mode!")
```
yuleisui, line 34: We don't need LLVM to be built in debug mode if SVF is built under debug

Johanmyst: Ah, alright, fair enough. I added this check because of this remark [from this setup guide](https://github.com/SVF-tools/SVF/wiki/Setup-Guide-(Autoconf)):

> Build the debug version (make sure you have also installed the debug version of LLVM)

That page hasn't been updated since 2018 though, so I assume this is either no longer the case or was written erroneously and just never updated? Either way, I'll remove this check from the CMake script and update the setup guide page!

CMakeLists.txt
```diff
@@ -5,15 +5,87 @@ project("SVF")
...
+  # and up to (and including) version 15 the default was C++14. From versions 16 and onwards, the
+  # default C++ standard has been C++17. Build SVF with C++14 unless the LLVM version is below 6.
+  if (LLVM_VERSION_MAJOR VERSION_LESS 6)
+    set(CMAKE_CXX_STANDARD 98)
+    message(STATUS "Got LLVM version ${LLVM_VERSION}; using C++ standard: ${CMAKE_CXX_STANDARD}")
+  elseif(LLVM_VERSION_MAJOR VERSION_LESS 16)
+    set(CMAKE_CXX_STANDARD 14)
+    message(STATUS "Got LLVM version ${LLVM_VERSION}; using C++ standard: ${CMAKE_CXX_STANDARD}")
+  else()
+    message(FATAL_ERROR "Found unsupported LLVM version; need <= 15, got: ${LLVM_VERSION}")
```
yuleisui, line 81: We will update to LLVM-16 in the coming weeks. Would be good to have this `else` branch open with C++14 for now?

Johanmyst: Interesting; I thought upgrading to LLVM-16 or higher was not planned due to the switch to opaque pointers? Either way, I'll modify this check to a warning for versions greater than 16. :+1:

yuleisui: Thanks.

Johanmyst: Done; for versions greater than 15 the C++17 standard will now be used (a warning will be printed for now, feel free to remove that when the switch to version LLVM 16 is released).

Johanmyst:
Pushed another small fix; because the LLVM definitions/include dirs/link dirs were added directly to the `SvfCore` target rather than globally the tools in `svf-llvm/tools/...` could not find the `libLLVM.so` library file, so this last commit adds the definitions/dirs globally so that any targets defined after/in subdirectories of `SvfLLVM` can also find this correctly.

yuleisui:
build fails again, looks to be the same issue when finding llvm.

svf-llvm/tools/CFL/CMakeLists.txt
```diff
@@ -3,7 +3,13 @@ if(DEFINED IN_SOURCE_BUILD)
...
   add_executable(cfl cfl.cpp)
 
-  target_link_libraries(cfl SvfLLVM ${llvm_libs})
+  separate_arguments(LLVM_DEFINITIONS_LIST NATIVE_COMMAND ${LLVM_DEFINITIONS})
+  target_include_directories(cfl PUBLIC include)
+  target_include_directories(cfl SYSTEM PUBLIC ${LLVM_INCLUDE_DIRS})
+  target_link_directories(cfl PUBLIC ${LLVM_LIBRARY_DIRS})
+  target_compile_definitions(cfl PUBLIC ${LLVM_DEFINITIONS})
+
+  target_link_libraries(cfl SvfCore SvfLLVM)
```
yuleisui, line 12: It looks to me like every tool has these repetitive lines. Could we move them to parent cmakelist?

CMakeLists.txt
```diff
@@ -5,15 +5,86 @@ project("SVF")
...
+    Version       ${LLVM_VERSION}
+    Definitions   ${LLVM_DEFINITIONS}
+    Includes      ${LLVM_INCLUDE_DIRS}
+    Libraries     ${LLVM_LIBRARY_DIRS}
+    Targets       ${LLVM_TARGETS_TO_BUILD}
+    Build type    ${LLVM_BUILD_TYPE}
+    Exceptions    ${LLVM_ENABLE_EH}
+    RTTI          ${LLVM_ENABLE_RTTI}
+    Dynamic lib   ${LLVM_LINK_LLVM_DYLIB}"
+  )
```
yuleisui, line 27: After a second look, the current cmakelist in the root folder was not supposed to have anything related to LLVM and those LLVM options should be moved to svf-llvm/CMakelist.txt. I assume this change should be straight-forward?

SVF itself is now designed language-independent, and we may have another module later (e.g., svf-xxx) which may not rely on llvm-based language, so a clean root cmakelist would be helpful.

svf-llvm/tools/CFL/CMakeLists.txt
Johanmyst: Of course; I did this because I thought the linking was going wrong but on my system this is difficult to reproduce as my LLVM instance's library directory is also in `$LD_LIBRARY_PATH`, so I cannot reproduce the issue with finding `libLLVM.[so|a]`. Especially since the CMake output during the git workflow don't include the build command itself, so I can't see if the issue is due to the `rpath` not being set right, or if somehow `link_directories()` is not propagating to targets defined in subdirectories.

Either way, it is not necessary to explicitly link against the LLVM library anyway since the library is defined using `add_llvm_library`, so I removed this part. Moreover, since SVF aims to support building in- and out-of-tree, the SVF-LLVM tools can/should also be defined using `add_llvm_executable` or even `add_llvm_tool` (I didn't do the latter for now as that means they only get built when specific build flags are set for LLVM). This simplifies the linking & the makefiles a bit since there is no longer a need for in-tree detection and what not.

Note that I've also modified `build.sh` and `setup.sh` to not only append LLVM's binary directory (for local installations by the build script) to the `$PATH` variable, but also its library directories to `$LD_LIBRARY_PATH` so that the linker can reliable find the libraries (idem ditto for Z3). I also did the same for the library files produced by SVF so targets linking against SVF can find them easier. Please let me know if you think this should be done differently!

CMakeLists.txt
Johanmyst: Alright; I moved this check to the top-level `CMakeLists.txt` because that was where flags like `-fno-exceptions` or `-fno-rtti` were added. I've moved this check back to the script in the `svf-llvm` directory; the flags for `-fno-rtti` or `-fno-exceptions` are now only added to `SvfLLVM` (if users want the main library to be built with it they can just specify it through `$CFLAGS` or `$CXXFLAGS`).

build.sh
```diff
@@ -205,7 +205,10 @@ if [[ ! -d "$Z3_DIR" ]]; then
     export Z3_DIR="$SVFHOME/$Z3Home"
 fi
 
-export PATH=$LLVM_DIR/bin:$PATH
+# Add LLVM & Z3 to $PATH and $LD_LIBRARY_PATH (prepend so that selected instances will be used first)
+export PATH=$LLVM_DIR/bin:$Z3_DIR/bin:$PATH
+export LD_LIBRARY_PATH=$LLVM_DIR/lib:$Z3_BIN/lib:$LD_LIBRARY_PATH
+
```
yuleisui, line 211: There is no need to export since the end of `build.sh` will call `setup.sh` to establish the environmental variables.

CMakeLists.txt
```diff
@@ -80,11 +72,18 @@ endif()
...
+  ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}
+)
 install(
   DIRECTORY ${PROJECT_SOURCE_DIR}/svf/include/
             ${PROJECT_SOURCE_DIR}/svf-llvm/include/
   COMPONENT devel
-  DESTINATION include/svf
+  DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/include/svf
   FILES_MATCHING
   PATTERN "**/*.h")
```
yuleisui, line 89: Could we also move the following to svf-llvm/CMakeList.txt?

# Compile extapi.c to extapi.bc
find_path(LLVM_CLANG_DIR
        NAMES clang llvm
        HINTS ${LLVM_DIR} ENV LLVM_DIR
        PATH_SUFFIXES bin)
add_custom_target(extapi_ir ALL
    COMMAND ${LLVM_CLANG_DIR}/clang -w -S -c -Xclang -disable-O0-optnone -fno-discard-value-names -emit-llvm ${PROJECT_SOURCE_DIR}/svf-llvm/lib/extapi.c -o ${PROJECT_BINARY_DIR}/svf-llvm/extapi.bc
    DEPENDS ${PROJECT_SOURCE_DIR}/svf-llvm/lib/extapi.c
)

CMakeLists.txt
```diff
@@ -2,19 +2,11 @@ cmake_minimum_required(VERSION 3.13.4)
...
 
-# We need to match the build environment for LLVM: In particular, we need C++14
-# and the -fno-rtti flag
-set(CMAKE_CXX_STANDARD 14)
-set(CMAKE_CXX_STANDARD_REQUIRED ON)
-# add -std=gnu++14
-set(CMAKE_CXX_EXTENSIONS ON)
-
-add_compile_options("-fno-rtti")
-add_compile_options("-fno-exceptions")
```
yuleisui, line 16: If we want to enable svfcore as `-fno-rtti/fno-exceptions`, these two options should also be in the root/CMakeList or svf/CMakeList?

svf-llvm/CMakeLists.txt
```diff
@@ -1,60 +1,83 @@
...
-  # include_directories(SYSTEM ${LLVM_INCLUDE_DIRS})
+if(NOT LLVM_ENABLE_EH)
+  message(STATUS "Building SVF-llvm without exception handling")
+  add_compile_options("-fno-exceptions")
+endif()
 
-  if(LLVM_LINK_LLVM_DYLIB)
-    set(llvm_libs LLVM)
+if(NOT LLVM_ENABLE_RTTI)
+  message(STATUS "Building SVF-llvm without RTII")
```
xudon9, line 32: `RTTI` is misspelled in the message.

build.sh
Johanmyst: Alright, I'll remove those references.

CMakeLists.txt
Johanmyst: Done.

CMakeLists.txt
Johanmyst: Sure, this was why I initially moved the LLVM finding into the top-level `CMakeLists.txt` since settings these options is based on whether the found LLVM instance supports them. I changed it now so that the `CMakeLists.txt` in `svf-llvm` sets a variable in its parent scope, which is then checked in the top-level `CMakeLists.txt` to conditionally set the flags for all targets.

CMakeLists.txt
Johanmyst: I also moved the installation commands to the subdirectory so the `SvfLLVM` target isn't referenced at all in the top-level `CMakeLists.txt` anymore.

While doing that I also fixed the remaining installation predicates:
1. `<svf_root>/svf/include` (i.e. its contents) is installed to `<install_prefix>/include/svf/`
2. `<svf_root>/svf-llvm/include` (i.e. the `SVF-LLVM` directory therein) is installed to `<install_prefix>/include/svf/`
3. `<svf_root/[Debug|Release]-Build/include/Util/config.h` is installed to `<install_prefix>/include/svf/Util/config.h`
4. `<svf_root/[Debug|Release]-Build/svf/libSvfCore.[a|so]` is installed to `<install_prefix>/lib/libSvfCore.[a|so]`
5. `<svf_root/[Debug|Release]-Build/svf-llvm/libSvfLLVM.[a|so]` is installed to `<install_prefix>/lib/libSvfLLVM[a|so]`
6. `<svf_root>/[Debug|Release]-Build/svf-llvm/extapi.bc` is installed to `<install_prefix>/include/svf/SVF-LLVM/extapi.bc` (I thought this was the best place as placing it by itself in `<install_prefix>/lib/extapi.bc` so that it's placed alongside `libSVF.[a|so]` seemed inconvenient)
7. Tools from `<svf_root>/svf-llvm/tools/...` (built to `<svf_root>/[Debug|Release]-Build/bin/...`) are installed to `<install_prefix>/bin/...`

This should be unnecessary for the way SVF is currently designed to be used (i.e. through the `build.sh` and `setup.sh` scripts), but it allows users who want to install SVF to a specific location to not have to manually link/include the different directories/library files so explicitly.

yuleisui:
Thanks and waiting for your push and will have another look.

svf-llvm/CMakeLists.txt
```diff
@@ -1,61 +1,129 @@
...
 add_llvm_library(SvfLLVM ${SVFLLVM_SOURCES})
-target_include_directories(SvfLLVM SYSTEM PUBLIC ${LLVM_INCLUDE_DIRS})
 target_include_directories(SvfLLVM PUBLIC include)
 target_link_libraries(SvfLLVM PUBLIC ${llvm_libs} ${Z3_LIBRARIES} SvfCore)
 
-if(DEFINED IN_SOURCE_BUILD)
+# Build SVF with C++ standard C++17
+set_property(TARGET SvfLLVM PROPERTY CXX_STANDARD 17)
+set_property(TARGET SvfLLVM PROPERTY CXX_EXTENSIONS ON)
+set_property(TARGET SvfLLVM PROPERTY CXX_STANDARD_REQUIRED ON)
```
xudon9, line 78: Any better reason to set the C++ standard for SVFcore and SVF-LLVM separately? I think the original way is less repetitive.

xudon9 (approved):
Other changes look good to me.

CMakeLists.txt
```diff
@@ -2,18 +2,9 @@ cmake_minimum_required(VERSION 3.13.4)
...
 project("SVF")
 
-configure_file(${PROJECT_SOURCE_DIR}/.config.in
-               ${PROJECT_BINARY_DIR}/include/Util/config.h)
+message(STATUS "Using CMake with generator: ${CMAKE_GENERATOR}")
 
-# We need to match the build environment for LLVM: In particular, we need C++14
-# and the -fno-rtti flag
-set(CMAKE_CXX_STANDARD 14)
-set(CMAKE_CXX_STANDARD_REQUIRED ON)
```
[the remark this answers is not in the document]
yuleisui: @Johanmyst could you add the options here and remove `set_property` in svf-llvm/CMakLiist.txt and svf/CMakeList.txt

svf-llvm/CMakeLists.txt
Johanmyst: I think CMake generally recommends applying options/properties to specific targets rather than using global options (e.g. the documentation for [`link_libraries()`](https://cmake.org/cmake/help/latest/command/link_libraries.html) states: "*The [target_link_libraries()](https://cmake.org/cmake/help/latest/command/target_link_libraries.html#command:target_link_libraries) command should be preferred whenever possible*".

However, I think the `set_property()` command -- by default -- applies the properties to the current directory and any subdirectories included later. I applied the option to each target separately to avoid clobbering the build configuration of any projects that might include SVF through `add_subdirectory()`, but I think this is indeed unnecessary. I'll move this back to the top-level `CMakeLists.txt`.

CMakeLists.txt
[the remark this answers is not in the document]
Johanmyst: Done. :+1:

yuleisui:
Merged and thanks @Johanmyst .

Johanmyst:
No problem!

PR #1302, opened 2023-12-28
bugfix in AndersenPWC

egnchen:
pretty obvious as it seems.


egnchen:
similar to https://github.com/SVF-tools/SVF/commit/a1cfc668092f549e60b8a462abd1fe067f364a05

yuleisui:
Thanks for the patch.

PR #1303, opened 2023-12-28
make the pointee type related to gep more robust

jumormt:
1. assert that we can have at most one pointer type (normally the first index that simply pieces the base pointer) when bridging a gep instruction. 

2. remove the assertion checking in getelementNum because the pointee may be of a pointer type for the multi-level pointer, e.g., `%2 = getelementptr inbounds i32*, i32** %1, ... ` the corresponding source code is like:

```cpp
int getValue(int** arr, int x, int y) {
    return arr[x][y];
}
```

This can be transformed to:  `%2 = getelementptr inbounds [N x i32], [N x i32]* %1, ...`
However, we do not know `N` without context information. `int**` implies **non-contiguous blocks of memory**. In this case, we conservatively return the max field limit.

### DIFF
--- a/svf-llvm/lib/SVFIRBuilder.cpp
+++ b/svf-llvm/lib/SVFIRBuilder.cpp
@@ -282,15 +282,16 @@ bool SVFIRBuilder::computeGepOffset(const User *V, AccessPath& ap)
 
     bool isConst = true;
 
-    bool addedPointerType = false;
+    bool prevPtrOperand = false;
     for (bridge_gep_iterator gi = bridge_gep_begin(*V), ge = bridge_gep_end(*V);
             gi != ge; ++gi)
     {
         const Type* gepTy = *gi;
         const SVFType* svfGepTy = LLVMModuleSet::getLLVMModuleSet()->getSVFType(gepTy);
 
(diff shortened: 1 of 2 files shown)


svf-llvm/lib/SVFIRBuilder.cpp
```diff
@@ -282,11 +282,15 @@ bool SVFIRBuilder::computeGepOffset(const User *V, AccessPath& ap)
...
     bool isConst = true;
 
+    bool addedPointerType = false;
     for (bridge_gep_iterator gi = bridge_gep_begin(*V), ge = bridge_gep_end(*V);
             gi != ge; ++gi)
     {
         const Type* gepTy = *gi;
         const SVFType* svfGepTy = LLVMModuleSet::getLLVMModuleSet()->getSVFType(gepTy);
+
+        assert((!addedPointerType || !svfGepTy->isPointerTy()) && "multiple pointer type?");
```
yuleisui, line 292: (addedPointerType && svfGepTy->isPointerTy()) == false

addedPointerType => prevPtrOperand

"multiple pointer type?" => "Expect no more than gep operand to be of a pointer type"

PR #1304, opened 2023-12-28
opaque pointer stage 2-1

jumormt:
infer object type based on instructions or globals


svf-llvm/lib/LLVMUtil.cpp
```diff
@@ -288,20 +282,15 @@ Type *LLVMUtil::getPointeeType(const Value *value)
...
-    {
-        // get the pointee type of return value
-        // e.g., for `%call = call ptr @_Znwm(i64 noundef 8)`, we return i64
-        return func->getFunctionType();
+        // get the pointee type of the global pointer (begins with @ symbol in llvm)
+        return global->getValueType();
     }
     else
     {
-        assert(false && (LLVMUtil::dumpValue(value) + "Unknown llvm Type, cannot get Ptr Element Type").c_str());
```
yuleisui, line 304: better to dump this unknown value.

svf-llvm/lib/SymbolTableBuilder.cpp
```diff
@@ -234,12 +234,15 @@ void SymbolTableBuilder::buildMemModel(SVFModule* svfModule)
...
-        // TODO: getPtrElementType to be removed
-        const Type* objtype = LLVMUtil::getPtrElementType(ptrType);
-        (void)getOrAddSVFTypeInfo(objtype);
+    Type *valType = val->getType();
+    (void)getOrAddSVFTypeInfo(valType);
+    if (SVFUtil::dyn_cast<PointerType>(valType))
+    {
+        // collect object type in instructions (alloc, heap) and globals
+        if (LLVMUtil::isObject(val)) {
+            const Type *objtype = LLVMUtil::getPointeeType(val);
```
yuleisui, line 243: would void* x = malloc() work here?

svf-llvm/lib/SymbolTableBuilder.cpp
```diff
@@ -590,8 +592,8 @@ ObjTypeInfo* SymbolTableBuilder::createObjTypeInfo(const Value* val)
     // (2) Other objects (e.g., alloca, global, etc.)
     else
     {
-        if(const PointerType* refTy = SVFUtil::dyn_cast<PointerType>(val->getType()))
-            objTy = getPtrElementType(refTy);
+        if(SVFUtil::isa<PointerType>(val->getType()))
```
yuleisui, line 595: It looks to me that this pointee type has to be inferred if this val is an object from a malloc site.

svf-llvm/lib/SymbolTableBuilder.cpp
yuleisui: Object type has to be inferred here?

svf-llvm/lib/SymbolTableBuilder.cpp
jumormt: > It looks to me that this pointee type has to be inferred if this val is an object from a malloc site.

This branch cannot be a malloc site. The object is alloca, global, etc. The malloc site is not yet done (to be done in inferTypeOfHeapObjOrStaticObj).

yuleisui: I see.

svf-llvm/lib/SymbolTableBuilder.cpp
```diff
@@ -598,6 +590,8 @@ ObjTypeInfo* SymbolTableBuilder::createObjTypeInfo(const Value* val)
 
     if (objTy)
     {
+        // collect object type information
+        (void) getOrAddSVFTypeInfo(objTy);
```
yuleisui, line 594: should return an SVFType ad use it at line 596.

jumormt:
[bash-opaque.diff.txt](https://github.com/SVF-tools/SVF/files/13824493/bash-opaque.diff.txt)
[redis-opaque.diff.txt](https://github.com/SVF-tools/SVF/files/13824494/redis-opaque.diff.txt)

PR #1305, opened 2023-12-28
Modernise CMake script

Johanmyst:
This pull request contains a number of optimisations/modernisations to the CMake infrastructure. The main improvement is that SVF can now be included in other projects using CMake's [`find_package(SVF)`](https://cmake.org/cmake/help/latest/command/find_package.html) command (including specifying version requirements). The CMake SVF package exports the following symbols/targets (available after calling `find_package(SVF)`):

- `SvfCore` and `SvfLLVM` shared library targets; can be linked directly through `[target_]link_libraries([target] SvfCore|SvfLLVM)`
- The SVF-LLVM tools `cfl`, `dvf`, `llvm2svf`, `mta`, `saber`, `svf-ex`, and `wpa` binaries
- The symbols defined in `.config.cmake.in`:
  - `SVF_INSTALL_ROOT`: installation prefix where SVF was installed
  - `SVF_INSTALL_[BIN|LIB|INCLUDE]_DIR`: respective subdirectories in installation
  - `SVF_INSTALL_EXTAPI_[DIR|FILE]`: directory containing/path to `extapi.bc`
  - These build options used to build SVF:
    - `SVF_SANITIZE` (string): sanitiser used when building SVF
    - `SVF_COVERAGE` (boolean): whether SVF was built with runtime instrumentation
    - `SVF_WARN_AS_ERROR` (boolean): whether SVF was compiled with compiler warnings treated as errors
    - `SVF_EXPORT_DYNAMIC` (boolean): whether (unused) dynamic symbols were exported to the dynamic symbol table
    - `SVF_ENABLE_ASSERTIONS` (boolean): whether assertions/debug things were enabled

Additionally, the following small fixes/changes were applied:
- CMake project definition now includes version number, docstring, and link to SVF homepage
- Output location of `extapi.bc` is more easily configurable (not yet through command line option)
- Enabling/disabling treating compiler warnings as errors is now a command-line option
- Adding of compiler options/definitions is now done through CMake's generator expressions (modernisation)
- `SvfCore` and `SvfLLVM` are added as export targets of SVF package
- Include headers are now included through current recommended method (i.e. using `FILE_SET HEADERS`)
- Use `GLOB_RECURSE` to find source files/headers instead of applying `GLOB` in every subdirectory of `svf`/`svf-llvm` to ensure new source directories are automatically included
- Generate `extapi.bc` using `add_custom_command()` rather than `add_custom_target()` as the latter is intended for commands without any output, while the prior is intended for commands that generate output files. A dependency on the generated output `extapi.bc` is defined from a custom target to the new custom command, thereby ensuring `extapi.bc` is correctly interpreted by CMake (e.g. to include it in the default `clean` target)

I've tested this on Linux using LLVM 15.0.7 and all seems functional both using the old approach as well as including SVF in projects through `find_package(SVF)`. Please let me know what you think/if you encounter any issues!


CMakeLists.txt
```diff
@@ -1,38 +1,93 @@
...
-  add_compile_options("-UNDEBUG")
-endif()
+add_compile_options($<$<BOOL:${SVF_ENABLE_ASSERTIONS}>:-UNDEBUG>)
 
-# Turn this on if you need symbols (e.g., use them for backtrace debugging)
-# add_link_options("-rdynamic")
+# Export dynamic symbols if requested (adds "-export-dynamic" to linkers that support it to enable backtraces)
+add_compile_options($<$<BOOL:${SVF_EXPORT_DYNAMIC}>:-rdynamic>)
 
-option(SVF_COVERAGE "Create coverage build")
```
yuleisui, line 26: Coverage file looks missing: 
https://github.com/SVF-tools/SVF/actions/runs/7349649179/job/20021425836?pr=1305#step:17:19

Johanmyst: My bad, I think this was because I forgot to check if `SVF_COVERAGE` is defined as an environment variable (I only checked for it as a command-line argument).

Johanmyst:
@yuleisui did you have another chance to take a look at these changes? I think it should all be properly functional now! :)

yuleisui:
This looks to be a major change to the cmake. I would like to keep this posted for sometime so that anyone else could review before I merge it. @xudon9 could you also take a look?

yuleisui:
I am not an expert in CMake. After your change, I am wondering about how an SVF subproject could utilize `find_package(SVF)`. Could you help with the CMake configuration for this SVF example project?"

Johanmyst:
Of course! I submitted the changes [as a pull request](https://github.com/SVF-tools/SVF-example/pull/62), please let me know what you think!

Note that I also changed some of the symbols that were exported through the CMake package (e.g. now SVF's build type is accessible when importing the SVF package), as well as enabling LTO & compiling with Position Independent Code so that transitive references are properly locatable.

yuleisui:
Thanks for the update.  Once the main SVF repo is updated, a [new SVF-npm is uploaded](https://github.com/SVF-tools/SVF/blob/b8c148c9b058d921333d86d9cb74c3e6a8fe946a/.github/workflows/svf-lib_publish.yml#L102).  In SVF-example and many other subprojects, SVF is [installed via npm](https://github.com/SVF-tools/SVF-example/blob/14ccc1738a40e98445a7f7e71a4581113806e528/.github/workflows/svf-example-build.yml#L39). I assume there is no need to change the folder layout of SVF-npm or any environment variables here: https://github.com/SVF-tools/SVF-npm?

Johanmyst:
No problem!

> I assume there is no need to change the folder layout of SVF-npm or any environment variables here (...)

I don't think so, no. Nothing has been changed about the way the files are laid out during the build phase. The only difference is that, with these changes, people are able to use SVF as a regular CMake package (in the same way that you'd include something like LLVM in a project) if they want to. The old way of using SVF (i.e. explicitly using directories like `Release-build` for finding the library files, or the `svf/include` and `svf-llvm/include` directories for the headers) still works the exact same way!

yuleisui:
> No problem!
> 
> > I assume there is no need to change the folder layout of SVF-npm or any environment variables here (...)
> 
> I don't think so, no. Nothing has been changed about the way the files are laid out during the build phase. The only difference is that, with these changes, people are able to use SVF as a regular CMake package (in the same way that you'd include something like LLVM in a project) if they want to. The old way of using SVF (i.e. explicitly using directories like `Release-build` for finding the library files, or the `svf/include` and `svf-llvm/include` directories for the headers) still works the exact same way!

Thanks and will merge towards end of this week or early next.

yuleisui:
@Johanmyst I have just merged your cmake patch. It has a problem with this error https://github.com/SVF-tools/SVF/actions/runs/7537407715/job/20516284438#step:5:14006

It looks to me that the FILE_SET variable is not defined: https://github.com/SVF-tools/SVF/blob/520a08f6c61f7272c2e5253d649b3a6233df76fe/svf-llvm/CMakeLists.txt#L132

yuleisui:
> FILE_SET

FILE_SET is only supported after cmake 3.2.3 which most ubuntu and windows do not support. Could you remove FILE_SET and use something else?

yuleisui:
> > FILE_SET
> 
> FILE_SET is only supported after cmake 3.2.3 which most ubuntu and windows do not support. Could you remove FILE_SET and use something else?

This patch looks to be a workaround solution: https://github.com/SVF-tools/SVF/pull/1328

Johanmyst:
@yuleisui Can I assume this is resolved then?

yuleisui:
Yes, that issue is resolved. However, assistance is still needed to modify CMake so that the folder layout generated in both Release-Build and Debug-Build is consistent with the layout in the installation directory. This would be good for any developer would like to debug locally compiled SVF without modifying the environmental variables. It is also good for us to copy the Release-Build folder directly to npm without installing... 

Could you help ensure that the folder layout in Release-Build mirrors that of the installation directory? This includes placing 'svfllvm.a' and 'svfcore.a' under the 'lib' directory, as well as putting the *.cmake files under 'cmake/SVF'.

### **Layout under Release-Build**:
Screen Shot 2024-01-22 at 8 01 54 pm

### **Layout under the Install directory**:

Johanmyst:
I've made some changes to the build system (see commit), but I don't really see how modifying the default build directory layout would be beneficial. Of course, some files could be relocated during the build phase (e.g. similar to how `svf-llvm/tools/...` are built to `[Debug|Release]-build/bin`), but that would still leave the header files separate (they are not copied during the build phase).

If anything, I'd argue no longer building the tools in `svf-llvm/tools` to `<build_dir>/bin` and just following the default build layout makes more sense, and to use the installation directory layout when desired by just using the installed version. For example, wouldn't it be easier to actually install the CMake package to a directory (e.g. with `cmake --install <build_dir> --prefix <install_dir>`) so that the "normal" layout with `install/bin`, `install/lib`, and `install/include` exist, and to copy that directory for creating the npm package rather than copying the build directory, some parts of the source directories, etc?

yuleisui:
> I've made some changes to the build system (see commit [#47c14a0](https://github.com/Johanmyst/SVF/commit/47c14a09269e830b9eca9167d1c302a70d252fd0)), but I don't really see how modifying the default build directory layout would be beneficial. Of course, some files could be relocated during the build phase (e.g. similar to how `svf-llvm/tools/...` are built to `[Debug|Release]-build/bin`), but that would still leave the header files separate (they are not copied during the build phase).
> 
> If anything, I'd argue no longer building the tools in `svf-llvm/tools` to `<build_dir>/bin` and just following the default build layout makes more sense, and to use the installation directory layout when desired by just using the installed version. For example, wouldn't it be easier to actually install the CMake package to a directory (e.g. with `cmake --install <build_dir> --prefix <install_dir>`) so that the "normal" layout with `install/bin`, `install/lib`, and `install/include` exist, and to copy that directory for creating the npm package rather than copying the build directory, some parts of the source directories, etc?

The reason for maintaining a consistent layout is that when the SVF repository and a sub-project coexist on someone's local machine, differing layouts can lead to confusion about which executable to debug. This is because you have to run 'make install' each time you want to debug your svfllvm library through a subproject, which significantly hampers local debugging efficiency. Personally, I prefer to have the same layout, where the binaries are already in place, requiring only minor adjustments for libraries. As for the headers, they should not pose an issue, as we can easily upload them to SVF's npm by copying the ones from the root of the source folder.

Issue #1306, opened 2023-12-29
Question: the path of SVFG for branches


Zhong0Lei:
When using Saber to test this test code, the expected results were not obtained
my codes：
```
#include<stdlib.h>
typedef struct Node {
    int data;           
    struct Node *next;  
} Node;
int main() {
    Node* head1 = (Node*)malloc(sizeof(Node)); 
    Node* head2 = (Node*)malloc(sizeof(Node)); 
    int a = 1;
    int b = 1;
    Node *p = head1;
    if(a == b){
        p = head2;
    }
    free(p);
    return 0;
}
```
The correct result should be that both head1 and head2 report `PartialLeaf`

But the test results show that head1 has no memory leak, while head2 reports PartialLeak

![image](https://github.com/SVF-tools/SVF/assets/47477662/306441d1-d460-4625-8eca-7a9997c12df4)

I checked the slice diagram and found that the path from the malloc node of head1 to the free node did not pass through `if...then`  block, may I ask why this is?

yuleisui:
@jumormt could you have a look?

jumormt:
This case is tricky. Different from head2, the conditional free path for head1 should be the if-else branch. It seems that the value-flow guard between p@Line 11 and p@Line 15 should be `a != b` instead of a true value.

Regarding why head1 does not pass through the if-then block, the points-to set of head1@Line 11 does not pass to p@Line 13 (i.e., no indirect value-flow between p@Line 11 and p@Line 13) because it is a strong update at Line 13. Therefore, saber removes that redundant indirect value-flow.

yuleisui:
@ZcoderL could you upload the c file and its svfg? @jumormt the strong update should remove 11->13 but should keep 11->15 and the value-flow guard should have a condition a!=b?

jumormt:
Yes, 11->13 is removed due to a strong update. 11->15 is a normal direct def-use and is kept in SVFG. The source code with line number:

![image](https://github.com/SVF-tools/SVF/assets/44386076/c8d7e923-5253-48fe-a280-38b47b98a27c)

And the SVFG (relevant part):

![image](https://github.com/SVF-tools/SVF/assets/44386076/67340c0c-a67a-4bca-97f9-e0373c4e0c90)

The entire SVFG:

![svfg](https://github.com/SVF-tools/SVF/assets/44386076/62184723-ae50-4b1d-bc51-8e591ac24ae0)

yuleisui:
[...]

PR #1309, opened 2023-12-30
fix issue #1306: compute vf guard for strong update in another branch

jumormt:
fix issue #1306: we compute the vf guard of invalid branches stemming from removed strong-update value-flow, and conjunct the negation of that.

### DIFF
--- a/svf/include/SABER/ProgSlice.h
+++ b/svf/include/SABER/ProgSlice.h
@@ -58,10 +58,13 @@ public:
     typedef FIFOWorkList<const SVFGNode*> VFWorkList;		    ///< worklist for value-flow guard computation
     typedef FIFOWorkList<const SVFBasicBlock*> CFWorkList;	///< worklist for control-flow guard computation
 
+    typedef SaberCondAllocator::SVFGNodeToSVFGNodeSetMap SVFGNodeToSVFGNodeSetMap;
+
+
     /// Constructor
-    ProgSlice(const SVFGNode* src, SaberCondAllocator* pa, const SVFG* graph, const Map<const SVFGNode*, SVFGNodeSet>& edges):
+    ProgSlice(const SVFGNode* src, SaberCondAllocator* pa, const SVFG* graph):
         root(src), partialReachable(false), fullReachable(false), reachGlob(false),
-        pathAllocator(pa), _curSVFGNode(nullptr), finalCond(pa->getFalseCond()), svfg(graph), removedSUVFEdges(edges)
+        pathAllocator(pa), _curSVFGNode(nullptr), finalCond(pa->getFalseCond()), svfg(graph)
     {
     }
 
@@ -299,6 +302,10 @@ protected:
     /// Compute invalid branch condition stemming from removed strong update value-flow edges
     Condition computeInvalidCondFromRemovedSUVFEdge(const SVFGNode * cur);
 
+    const SVFGNodeToSVFGNodeSetMap& getRemovedSUVFEdges() const {
+        return pathAllocator->getRemovedSUVFEdges();
+    }
+
 private:
     SVFGNodeSet forwardslice;				///<  the forward slice
     SVFGNodeSet backwardslice;				///<  the backward slice
@@ -312,7 +319,6 @@ private:
     const SVFGNode* _curSVFGNode;			///<  current svfg node during guard computation
     Condition finalCond;					///<  final condition
     const SVFG* svfg;						///<  SVFG
-    const Map<const SVFGNode*, SVFGNodeSet>& removedSUVFEdges;
 };
 
 } // End namespace SVF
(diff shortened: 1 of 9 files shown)


svf/lib/SABER/SaberSVFGBuilder.cpp
```diff
@@ -276,6 +276,8 @@ void SaberSVFGBuilder::rmIncomingEdgeForSUStore(BVDataPTAImpl* pta)
                     }
                     for (SVFGEdge* edge: toRemove)
                     {
+                        if (isa<StoreSVFGNode>(edge->getSrcNode()))
```
yuleisui, line 279: Do we need this condition? Not all edges in toRemove?

svf/lib/SABER/SrcSnkDDA.cpp
```diff
@@ -97,6 +97,7 @@ void SrcSnkDDA::analyze(SVFModule* module)
             if(Options::DumpSlice())
                 annotateSlice(_curSlice);
 
+            memSSA.getRemovedSUVFEdges();
```
yuleisui, line 100: why call this method here?

svf/include/SABER/SaberSVFGBuilder.h
```diff
@@ -97,6 +101,8 @@ class SaberSVFGBuilder : public SVFGBuilder
     PointsTo globs;
     /// Store all global SVFG nodes
     SVFGNodeSet globSVFGNodes;
+    /// Removed strong update edge
+    Map<const SVFGNode*, SVFGNodeSet> removedSUVFEdges;
```
yuleisui, line 105: better to move this map to SaberCondAllocator

svf/include/SABER/SaberSVFGBuilder.h
```diff
@@ -62,6 +62,10 @@ class SaberSVFGBuilder : public SVFGBuilder
         svfg->addActualParmVFGNode(pagNode, cs);
     }
 
+    const Map<const SVFGNode*, SVFGNodeSet>& getRemovedSUVFEdges() const {
```
yuleisui, line 65: better to move this method under SaberCondAllocator

svf/include/SABER/ProgSlice.h
```diff
@@ -309,6 +312,7 @@ class ProgSlice
     const SVFGNode* _curSVFGNode;			///<  current svfg node during guard computation
     Condition finalCond;					///<  final condition
     const SVFG* svfg;						///<  SVFG
+    const Map<const SVFGNode*, SVFGNodeSet>& removedSUVFEdges;
```
yuleisui, line 315: better to move this map to SaberCondAllocator

svf/include/SABER/ProgSlice.h
```diff
@@ -59,9 +59,9 @@ class ProgSlice
     typedef FIFOWorkList<const SVFBasicBlock*> CFWorkList;	///< worklist for control-flow guard computation
 
     /// Constructor
-    ProgSlice(const SVFGNode* src, SaberCondAllocator* pa, const SVFG* graph):
+    ProgSlice(const SVFGNode* src, SaberCondAllocator* pa, const SVFG* graph, const Map<const SVFGNode*, SVFGNodeSet>& edges):
```
yuleisui, line 62: keep the constructor unchanged and move the map to SaberCondAllocator

svf/lib/SABER/SaberSVFGBuilder.cpp
jumormt: Yes, the condition should be removed. But [this case](https://github.com/SVF-tools/Test-Suite/blob/master/src/mem_leak/malloc43.c) won't pass because of an IntraMSSAPhinode to strong update. I'm afraid there will be more false positives.

svf/include/SABER/SaberSVFGBuilder.h
jumormt: I'm not sure. removedSUVFEdges is derived from saber svfg builder. The Sabercondallocator is not yet initialized at the moment.

svf/include/SABER/SaberSVFGBuilder.h
```diff
@@ -33,6 +33,8 @@
 #include "MSSA/SVFGBuilder.h"
 #include "SVFIR/SVFValue.h"
 #include "Util/WorkList.h"
+#include "SABER/SaberCondAllocator.h"
```
yuleisui, line 36: Put this include to cpp and use forward declaration instead.

yuleisui:
CFL cases failed, looks to be assertion failures.

jumormt:
> CFL cases failed, looks to be assertion failures.

Yes, CFL uses sabersvfgbuilder.

jumormt:
> > CFL cases failed, looks to be assertion failures.
> 
> Yes, CFL uses sabersvfgbuilder.

https://github.com/SVF-tools/SVF/blob/master/svf/include/CFL/CFLVF.h#L62

yuleisui:
> > > CFL cases failed, looks to be assertion failures.
> > 
> > 
> > Yes, CFL uses sabersvfgbuilder.
> 
> https://github.com/SVF-tools/SVF/blob/master/svf/include/CFL/CFLVF.h#L62

I see, we will then need to set the condition allocator.

PR #1310, opened 2023-12-31
Fix bugs for big program

bjjwwang:
1. in SVFStat.h. 
```
typedef OrderedMap<std::string, u32_t> NUMStatMap;
```
The key type should be string, since `const char*` cannot support NUMStatMap["$KEYNAME"]

2. in SVFIR2ItvExeState.cpp
add SVFOtherType handler, since float/double is one of SVFOtherType, the intervalValue range of double/float is s32_t's range.

### DIFF
--- a/svf/include/Util/SVFStat.h
+++ b/svf/include/Util/SVFStat.h
@@ -30,6 +30,8 @@
 #ifndef SVF_SVFSTAT_H
 #define SVF_SVFSTAT_H
 
+#include <string>
+
 namespace SVF
 {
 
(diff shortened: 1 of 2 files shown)


svf/include/Util/SVFStat.h
```diff
@@ -30,6 +30,8 @@
 #ifndef SVF_SVFSTAT_H
 #define SVF_SVFSTAT_H
 
+#include <string>
```
yuleisui, line 33: Do we need to include the string header here? It would be good to put in the cpp file.

PR #1315, opened 2024-01-03
remove isNonPointerObj API

yuleisui:


jumormt:
[bash-objptr.diff.txt](https://github.com/SVF-tools/SVF/files/13824975/bash-objptr.diff.txt)
[redis-objptr.diff.txt](https://github.com/SVF-tools/SVF/files/13824976/redis-objptr.diff.txt)

PR #1356, opened 2024-02-02
remove getbasevalvar in getgepval

jumormt:


svf/lib/SVFIR/SVFIR.cpp
```diff
@@ -385,7 +385,7 @@ GepStmt* SVFIR::addVariantGepStmt(NodeID src, NodeID dst, const AccessPath& ap)
  */
 NodeID SVFIR::addGepValNode(const SVFValue* curInst,const SVFValue* gepVal, const AccessPath& ap, NodeID i, const SVFType* type)
 {
-    NodeID base = getBaseValVar(getValueNode(gepVal));
```
yuleisui, line 388: Remove this function if it will never be used later.

Issue #1314, opened 2024-01-03
The Call Graph Problem of Virtual Functions


ZcoderL:
I recently attempted to analyze the entire project using SVF and encountered a problem with identifying virtual function calls.
Here is a simple case I reproduced
```
#include <stdio.h>
namespace name1
{
    class name1_base
    {
    public:
        /* pure virtual function */
        virtual void action(int * data);
    };
    class int_name1 : public name1_base
    {
    public:
        void action(int * data);
    };
    void int_name1::action(int * data)
    {
        printf("%d",*data);
    }
    void test()
    {
        int * data;
        data = nullptr;
        name1_base* baseObject = new int_name1;
        baseObject->action(data);
        delete baseObject;
    }
}
```
```
#include <stdio.h>
namespace name2
{
    class name2_base
    {
    public:
        virtual void action(char * data);
    };
    class long_name2 : public name2_base
    {
    public:
        void action(char * data);
    };
    void long_name2::action(char * data)
    {
        printf("%d",*data);
    }
    void test()
    {
        char * data;
        data = nullptr;
        name2_base* baseObject = new long_name2;
        baseObject->action(data);
        delete baseObject;
    }
}
```
Two namespaces were used in each of the two CPP files; The content is basically consistent (except for class and parameter names that are different), then compile and link them separately with `llvm-link name1.bc name2.bc -o test.bc`. Generate a call graph for the linked BC file, and it is found that there is a missing call to a function.
![image](https://github.com/SVF-tools/SVF/assets/47477662/907a6822-404c-43f6-ac29-10127356ebbe)

But if all the content is written in one file, then the result is normal.
![image](https://github.com/SVF-tools/SVF/assets/47477662/4a294f98-ea34-4382-8ea6-dd4c367b7e5d)

Can you give me some help? Thank you！
The test files and callgraph are as follows：
[test24-1-3.zip](https://github.com/SVF-tools/SVF/files/13815407/test24-1-3.zip)

Issue #1301, opened 2023-12-25
SVF seems to be incomplete when dealing with virtual calls.


Lqs66:
I'm using SVF version 2.1.

The omission occurs when I write SVF as a lib to get the callsite of all the virtual functions and their targets. As far as I know this should be complete.

My core code is as follows:

```
//
// Created by lqs66 on 23-12-24.
//

/**
 *   A driver program of SVF including usages of SVF APIs.
 *   vcall used to get all virtual callsites and targets
*/

#include "SVF-FE/LLVMUtil.h"
#include "Graphs/SVFG.h"
#include "WPA/Andersen.h"
#include "SVF-FE/PAGBuilder.h"
#include "MemoryModel/PointerAnalysis.h"
#include <chrono>

using namespace llvm;
using namespace std;
using namespace SVF;

static llvm::cl::opt<std::string> InputFilename(cl::Positional,
                                                llvm::cl::desc("<input bitcode>"), llvm::cl::init("-"));

int main(int argc, char ** argv)
{
    auto start_time = std::chrono::high_resolution_clock::now();
    int arg_num = 0;
    char **arg_value = new char*[argc];
    std::vector<std::string> moduleNameVec;
    SVFUtil::processArguments(argc, argv, arg_num, arg_value, moduleNameVec);
    cl::ParseCommandLineOptions(arg_num, arg_value,
                                "virtual callsites Analysis\n");

    //parse ir file
    SVFModule* svfModule = LLVMModuleSet::getLLVMModuleSet()->buildSVFModule(moduleNameVec);

    /// Build Program Assignment Graph (PAG)
    PAGBuilder builder;
    PAG* pag = builder.build(svfModule);

    /// Create Andersen's pointer analysis. The analyze() function has been called at creation time. 
    Andersen* ander = AndersenWaveDiff::createAndersenWaveDiff(pag);

    //get pta call graph
    PTACallGraph* pta = ander->getPTACallGraph();

    //get all virtual callsites and targets
    PointerAnalysis::CallEdgeMap temp = pta->getIndCallMap();
    for(auto it = temp.begin(); it != temp.end(); it++)
    {
        const CallBlockNode* vcall_node = it->first;
        const Instruction *vcall_site = vcall_node->getCallSite();
        PointerAnalysis::FunctionSet& targets = it->second;

        if (const CallBase *callInst = dyn_cast<CallBase>(vcall_site)) {
            string valueStr;
            raw_string_ostream valueStream(valueStr);
[...]

PR #1369, opened 2024-02-12
Fix incomplete/incorrect PTA when using C++ namespaces & virtual functions

Johanmyst:
This pull request fixes a bug that caused call graphs to be incomplete/incorrect when using C++ namespaces and virtual functions.

Currently, whenever namespaces and virtual functions are used with C++ classes, the underlying object's name is used to find the VTable for a called function. However, the name used therein is incorrect (the namespace component is stripped off when it shouldn't be), and causes incomplete and incorrect results to be returned, leading to an incomplete callgraph.

For example, take the following snippets:

Example 1:
```
#include <cstdio>
#include <cstring>

class Parent
{
    public:
        Parent( )          = default;
        virtual ~Parent( ) = default;

        virtual void action( ) = 0;
};

class Child : public Parent
{
    public:
        Child( )  = default;
        ~Child( ) = default;

        void action( ) { printf("Doing stuff...\n"); }
};

int main(int argc, char *argv[])
{
    Child *obj = new Child;
    obj->action( );
    delete obj;
    return 0;
}
```
This corresponds to the following — expected — callgraph:
correct callgraph dot

Example 2:
```
#include <cstdio>
#include <cstring>

namespace n1
{

class Parent
{
    public:
        Parent( )          = default;
        virtual ~Parent( ) = default;

        virtual void action( ) = 0;
};

class Child : public Parent
{
    public:
        Child( )  = default;
        ~Child( ) = default;

        void action( ) { printf("Doing stuff...\n"); }
};

}    // namespace n1

int main(int argc, char *argv[])
{
    n1::Child *obj = new n1::Child;
    obj->action( );
    delete obj;
    return 0;
}
```

This results in the following, incomplete callgraph:
incorrect callgraph dot

The introduction of the namespaces causes the virtual call not to be found. This is because the set of indirect call edges found in the `AndersenBase::solveConstraints()` function ([line 118](https://github.com/SVF-tools/SVF/blob/dda7fe5da77043bc09dc0a25cc3d83fd7a617764/svf/lib/WPA/Andersen.cpp#L118)) is empty in the second example.

This set is retrieved from the PAG's indirect callsites `SVFIR::indCallSiteToFunPtrMap`, which is then passed to [`Andersen::updateCallGraph()`](https://github.com/SVF-tools/SVF/blob/dda7fe5da77043bc09dc0a25cc3d83fd7a617764/svf/lib/WPA/Andersen.cpp#L654), which again calls [`BVDataPTAImpl::onTheFlyCallGraphSolve`](https://github.com/SVF-tools/SVF/blob/dda7fe5da77043bc09dc0a25cc3d83fd7a617764/svf/lib/MemoryModel/PointerAnalysisImpl.cpp#L492). This calls [`PointerAnalysis::resolveCPPIndCalls`](https://github.com/SVF-tools/SVF/blob/dda7fe5da77043bc09dc0a25cc3d83fd7a617764/svf/lib/MemoryModel/PointerAnalysis.cpp#L486), which, by default, will not connect virtual calls based on the Call Hierarchy Analysis (CHA), and thus calls [`PointerAnalysis::getVFnsFromPts` ](https://github.com/SVF-tools/SVF/blob/dda7fe5da77043bc09dc0a25cc3d83fd7a617764/svf/lib/MemoryModel/PointerAnalysis.cpp#L438).

This function calls `chgraph->csHasVtblsBasedonCHA(SVFUtil::getSVFCallSite(cs->getCallSite()))`, which returns `false` in the second example, because the CHG's `CHG::csToCHAVtblsMap` map has no VTable entries for the given callsite. During construction of the CHG however this map *is* correctly built (note: **without** stripping the namespace component from the objects) and contains mappings for `n1::Child` and such.

However, during lookup, the `getClassNameOfThisPtr` function returns only `Child`, rather than `n1::Child`, causing the lookup to fail, and no call-edge to be created. This is due to `inferThisPtrClsName` calling `extractClsNamesFromFunc`, which calls `updateClassNameBeforeBrackets` on the demangled name. However, this last function call strips the `n1::` part from the demangled name incorrectly, causing the broken lookup. Since this function only gets called on this code-path, removing this stripping (which makes it consistent with other places) should be fine and produces correct call graphs for the above examples.

I believe this could also be the underlying issue for issue #1301 and possibly issue #1314.

### DIFF
--- a/svf-llvm/lib/CppUtil.cpp
+++ b/svf-llvm/lib/CppUtil.cpp
@@ -639,7 +639,6 @@ Set<std::string> cppUtil::extractClsNamesFromFunc(const Function *foo)
     {
         // c++ constructor
         DemangledName demangledName = cppUtil::demangle(name);
-        updateClassNameBeforeBrackets(demangledName);
         return {demangledName.className};
     }
     else if (isTemplateFunc(foo))


yuleisui:
Thanks for the pull request.

@jumormt could you help review this too? We will also need to add the test cases into SVF-Test-Suite for our CI.

jumormt:
> extractClsNamesFromFunc

This makes sense. The semantic is different from `isConstructor'. I will add the test case to SVF Test Suite.

PR #1371, opened 2024-02-13
add some comments for cpp type inference

jumormt:


svf-llvm/lib/CppUtil.cpp
```diff
@@ -398,7 +400,9 @@ const Argument* cppUtil::getConstructorThisPtr(const Function* fun)
     return thisPtr;
 }
 
-void updateClassNameBeforeBrackets(cppUtil::DemangledName& dname)
+/// strip brackets and namespace in classname
```
yuleisui, line 403: Strip off

yuleisui: Strip sth from sth

PR #1397, opened 2024-03-06
fix LD_LIBRARY_PATH error in setup.sh

yaoyue123:
$Z3_BIN is not defined in the setup.sh script, and libz3.so is located in the $Z3_DIR/bin directory. 
image


yaoyue123:
收到

Issue #1392, opened 2024-03-05
Assertion when using SVF on Ceph


Louise4one:
I tried to use SVF on Ceph (version 18.2.1). First I used WLLVM (LLVM version 16.0.0, Clang version 10.0.0) to generate Ceph bitcode, then I wanted to run wpa on Ceph RGW. But I countered with assertion below:

```
 /ceph/build/bin# wpa -ander radosgw.bc
wpa: /SVF/clang+llvm-16.0.0-x86_64-linux-gnu-ubuntu-18.04/include/llvm/IR/Function.h:785: llvm::Argument* llvm::Function::getArg(unsigned int) const: Assertion `i < NumArgs && "getArg() out of range!"' failed.
Aborted (core dumped)
```

I think it might be an unavoidable warning if SVF is used on such complex project, so I just ignored it. Then another assertion came out:

```
/ceph/build/bin# wpa -ander radosgw.bc
wpa: /SVF/svf/include/Util/Casting.h:198: static bool SVF::SVFUtil::isa_impl_cl<To, const From*>::doit(const From*) [with To = llvm::GetElementPtrInst; From = llvm::Value]: Assertion `Val && "SVFUtil::isa<> used on a null pointer"' failed.
Aborted (core dumped)
```

What leads to this assertion? Should I ignored this one again? If not, how can I solve this problem? 

By the way, I tried to turn off all assertions in SVF and ran Ceph RGW again. However, segmentation fault happened this time, and the gdb information was: 

```
(gdb) bt
#0  0x000055c105a75391 in SVF::ObjTypeInference::fwInferObjType(llvm::Value const*) ()
#1  0x000055c105a89116 in SVF::SymbolTableBuilder::inferTypeOfHeapObjOrStaticObj(llvm::Instruction const*) ()
#2  0x000055c105a8ac3d in SVF::SymbolTableBuilder::createObjTypeInfo(llvm::Value const*) ()
#3  0x000055c105a8b3ad in SVF::SymbolTableBuilder::collectObj(llvm::Value const*) ()
#4  0x000055c105a8be18 in SVF::SymbolTableBuilder::buildMemModel(SVF::SVFModule*) ()
#5  0x000055c105a525fe in SVF::LLVMModuleSet::buildSymbolTable() const ()
[...]

PR #1403, opened 2024-03-09
fix var arg and chgnode not found error

jumormt:


jumormt:
this should fix issue #1392

PR #1408, opened 2024-03-15
update AE library

jiawei-95:


svf/lib/AE/Core/RelationSolver.cpp
```diff
@@ -121,7 +124,7 @@ IntervalESBase RelationSolver::RSY(const IntervalESBase& domain, const Z3Expr& p
         Z3Expr rhs = !(gamma_hat(lower, domain));
         solver.push();
         solver.add(phi.getExpr() && rhs.getExpr());
-        Map<u32_t, double> solution;
+        Map<u32_t, int32_t> solution;
```
yuleisui, line 127: replace all int32_t with s32_t?

PR #1409, opened 2024-03-15
set name for each value

jumormt:


jumormt:
[redis.diff.txt](https://github.com/SVF-tools/SVF/files/14612705/redis.diff.txt)

PR #1422, opened 2024-04-02
Add llvm-17.0.6 support

adityavardhanpadala:
There were only very few changes from llvm 16 to llvm 17
-  Module::getGlobalList() was made private since insertGlobalVariable() was added 
- DbgInfoIntrinsic was used instead of DbgAddrIntrinsic, as debug.addr was removed from the IR

fixing these issues lets SVF build with llvm 17 and passes all the tests in the test suite.

### DIFF
--- a/svf-llvm/lib/LLVMUtil.cpp
+++ b/svf-llvm/lib/LLVMUtil.cpp
@@ -510,7 +510,14 @@ void LLVMUtil::removeFunAnnotations(Set<Function*>& removedFuncList)
     glob->setName("llvm.global.annotations.old");
     GlobalVariable *GV = new GlobalVariable(newCA->getType(), glob->isConstant(), glob->getLinkage(), newCA, "llvm.global.annotations");
     GV->setSection(glob->getSection());
+
+#if (LLVM_VERSION_MAJOR < 17)
+    module->getGlobalList().push_back(GV);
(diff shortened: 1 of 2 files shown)


svf-llvm/lib/LLVMUtil.cpp
```diff
@@ -511,7 +510,7 @@ void LLVMUtil::removeFunAnnotations(Set<Function*>& removedFuncList)
     glob->setName("llvm.global.annotations.old");
     GlobalVariable *GV = new GlobalVariable(newCA->getType(), glob->isConstant(), glob->getLinkage(), newCA, "llvm.global.annotations");
     GV->setSection(glob->getSection());
-    module->getGlobalList().push_back(GV);
+    module->insertGlobalVariable(GV);
```
yuleisui, line 513: Thanks for the contribution. Could you add a macro for this?

#if (LLVM_VERSION_MAJOR < 17)
    module->getGlobalList().push_back(GV);
#elif (LLVM_VERSION_MAJOR >= 17)
    module->insertGlobalVariable(GV);
#else
    assert(false && "llvm version not supported!");
#endif

adityavardhanpadala: I added the macro.

PR #1446, opened 2024-04-28
Fix bug and CI

jumormt:


svf/lib/Graphs/VFG.cpp
```diff
@@ -754,7 +754,9 @@ void VFG::connectDirectVFGEdges()
                 continue;
             /// for all other cases, like copy/gep/load/ret, connect the RHS pointer to its def
             if (stmtNode->getPAGSrcNode()->isConstDataOrAggDataButNotNullPtr() == false)
-                addIntraDirectVFEdge(getDef(stmtNode->getPAGSrcNode()), nodeId);
+                // for ptr only vfg, check whether src node is interested
```
yuleisui, line 757: Explain this is because of the src node of integer type if it is at a int2ptr copystmt

Issue #1483, opened 2024-06-11
SVF Global not found


LeoneChen:
I use https://github.com/SVF-tools/SVF-example to process two *.ll files (in attachment), but get "SVF Global not found"

Inside `SVFModule* svfModule = LLVMModuleSet::buildSVFModule(moduleNameVec);`

`llvmVal` may be a global replacement of val,
https://github.com/SVF-tools/SVF/blob/0eae65388008dbfafc0287873bfb4c8819582183/svf-llvm/lib/LLVMModule.cpp#L1221-L1223

and then in `getSVFGlobalValue`, `LLVMConst2SVFConst` may not able to find `g` in itself since `LLVMConst2SVFConst` only updated on non-replacement value, then reports "SVF Global not found!"
https://github.com/SVF-tools/SVF/blob/0eae65388008dbfafc0287873bfb4c8819582183/svf-llvm/include/SVF-LLVM/LLVMModule.h#L236

[capsule_manager-43d688dc37a73ece.zip](https://github.com/user-attachments/files/15786511/capsule_manager-43d688dc37a73ece.zip)

yuleisui:
Any idea to fix this?

LeoneChen:
I don't have a global view of svf project right now, and I don't know if it's feasible to update the replacement value or other similar solutions while adding per element to LLVMConst2SVFConst.

yuleisui:
What is the replacement value? Shall we use this value for global or use the previous one. We will need to map multiple global values to the one actually used?

LeoneChen:
Sorry I mean global replacement usually used in optimization for global value.
https://chatgpt.com/share/2d4cc8ce-9ce4-4b87-afd6-7fcffb4347ad

I'm not sure how to fix it

LeoneChen:
Any idea about this statement `const Value* llvmVal = LLVMUtil::getGlobalRep(val);`

yuleisui:
This is used to get the representation global if we have multiple global symbols referring to the same global variable, so that we could use a unique global to represent all.

looks like we didn’t set the rep global first.

LeoneChen:
Ok I got it, let me test if updating `LLVMConst2SVFConst` using the representation global works

yuleisui:
Thanks

LeoneChen:
[...]

PR #1484, opened 2024-06-12
Fix issue of global representation

LeoneChen:
https://github.com/SVF-tools/SVF/issues/1483

This patch is not widely tested, and we need to think more about how to fix the root causes

### DIFF
--- a/svf-llvm/include/SVF-LLVM/LLVMModule.h
+++ b/svf-llvm/include/SVF-LLVM/LLVMModule.h
@@ -174,6 +174,11 @@ public:
     }
     inline void addGlobalValueMap(const GlobalValue* glob, SVFGlobalValue* svfglob)
     {
+        if (auto glob_var = llvm::dyn_cast<llvm::GlobalVariable>(glob);
+            hasGlobalRep(glob_var))
+        {
+            glob = getGlobalRep(glob_var);
+        }
         LLVMConst2SVFConst[glob] = svfglob;
         setValueAttr(glob,svfglob);
     }
@@ -232,6 +237,11 @@ public:
 
     inline SVFGlobalValue* getSVFGlobalValue(const GlobalValue* g) const
     {
+        if (auto glob_var = llvm::dyn_cast<llvm::GlobalVariable>(g);
+            hasGlobalRep(glob_var))
+        {
+            g = getGlobalRep(glob_var);
+        }
         LLVMConst2SVFConstMap::const_iterator it = LLVMConst2SVFConst.find(g);
         assert(it!=LLVMConst2SVFConst.end() && "SVF Global not found!");
         assert(SVFUtil::isa<SVFGlobalValue>(it->second) && "not a SVFGlobal type!");


svf-llvm/include/SVF-LLVM/SVFIRBuilder.h
```diff
@@ -97,6 +97,7 @@ class SVFIRBuilder: public llvm::InstVisitor<SVFIRBuilder>
     /// GetObject - Return the object node (stack/global/heap/function) according to a LLVM Value
     inline NodeID getObjectNode(const Value* V)
     {
+        V = LLVMUtil::getGlobalRep(V);
```
yuleisui, line 100: Global rep should always be a global. The getObject node can return a null global obj. Why adding this line?

LeoneChen: The reason why I add `V = LLVMUtil::getGlobalRep(V);` is that `val` is converted to global representation in `collectObj` to perform `insert` on `symInfo->objSymMap`
https://github.com/SVF-tools/SVF/blob/0eae65388008dbfafc0287873bfb4c8819582183/svf-llvm/lib/SymbolTableBuilder.cpp#L318-L330

LeoneChen: Let me give you a case

LeoneChen: Stack is like this (pictures below may be more vivid), `objSymMap` is updated in above `collectObj`
image
image
image

LeoneChen: But I'm not sure whether this fix is accurate or not

svf-llvm/include/SVF-LLVM/SVFIRBuilder.h
```diff
@@ -97,6 +97,7 @@ class SVFIRBuilder: public llvm::InstVisitor<SVFIRBuilder>
...
         SVFValue* svfVal = LLVMModuleSet::getLLVMModuleSet()->getSVFValue(V);
```
yuleisui, line 101: I think it is good to put your ` V = LLVMUtil::getGlobalRep(V);` into method `getSVFValue `

LeoneChen:
Since it influences all usage about global value, does it will cause new problem?

yuleisui:
> Since it influences all usage about global value, does it will cause new problem?

It looks to be a clean fix for global. Would you test more and larger bcs before I merge this?

LeoneChen:
I agree that more testing is needed and I will report here if I find something new. Does SVF have some unit test cases and integration test cases that can perform the testing?

yuleisui:
> I agree that more testing is needed and I will report here if I find something new. Does SVF have some unit test cases and integration test cases that can perform the testing?

Your patch has already passed SVF tests in our CI. Would be good for you to test some large bcs and report back so I can merge your pull request

LeoneChen:
ok

PR #1491, opened 2024-06-20
Fix the problem that the intrinsic instruction is the successor of BranchStmt

LeoneChen:


svf-llvm/lib/SVFIRBuilder.cpp
```diff
@@ -936,7 +936,17 @@ void SVFIRBuilder::visitBranchInst(BranchInst &inst)
...
+        Instruction* succInst = nullptr;
+        for (auto& inst : *succBB)
+        {
+            if (not isIntrinsicInst(&inst))
+            {
+                succInst = &inst;
+                break;
+            }
+        }
+        assert(succInst && "no non-intrinsic instruction in the successor BB");
```
yuleisui, line 949: Better to find the next basic block if this bb has intrinsic instructions only before this assertion is made

yuleisui: I forgot whether we have built ICFG yet or not if built we would use the ICFG’s edge to find the successor instructions

LeoneChen: > Better to find the next basic block if this bb has intrinsic instructions only before this assertion is made

But this can be hard when succBB has more than 1 successor

LeoneChen: If succBB has only one successor, I think it's ok

PR #1500, opened 2024-07-14
fix "int a[2] = {2, 2};" case

bjjwwang:


svf-llvm/lib/SVFIRExtAPI.cpp
```diff
@@ -44,8 +44,27 @@ const Type* SVFIRBuilder::getBaseTypeAndFlattenedFields(const Value* V, std::vec
...
+    }
+    else
+    {
+        objTy = LLVMModuleSet::getLLVMModuleSet()->getTypeInference()->inferObjType(
+                value);
+    }
+
+
+    assert(objTy && "objTy should not be null");
+    u32_t numOfElems = pag->getSymbolInfo()->getNumOfFlattenElements(LLVMModuleSet::getLLVMModuleSet()->getSVFType(objTy));
```
yuleisui, line 67: Can we put these lines into inferObjType?

PR #1506, opened 2024-07-26
add interval constructor validation

bjjwwang:


svf/include/AE/Core/IntervalValue.h
```diff
@@ -81,22 +81,20 @@ class IntervalValue
...
     }
 
-    /// Create the bottom IntervalValue
+    /// Create the bottom IntervalValue [+inf, -inf]
     static IntervalValue bottom()
     {
-        IntervalValue res;
-        res.set_to_bottom();
-        return res;
+        return IntervalValue();
```
yuleisui, line 93: Better to use the constructor intervalValue(+inf,-inf);

PR #1522, opened 2024-08-19
Extensibility improvements for `PointsTo` data structures

cubicYYY:
Add helper macros to register different underlying data structure classes for `PointsTo`.
The corresponding function interface of the classes must be the same, to use the macros to abbreviate an operation on them.
In the future, the interfaces of the underlying classes should be verified through templates or [concepts](https://en.cppreference.com/w/cpp/language/constraints).

The reasons for not using virtual classes are backward compatibility and minimizing runtime overhead.


cubicYYY:
I wonder if there is a solution more elegant.

yuleisui:
I prefer not to use macros which make the code not readable. You could revise your patch to make it more clear.

cubicYYY:
> I prefer not to use macros which make the code not readable. You could revise your patch to make it more clear.

Seems like using macros is unavoidable unless we introduce a class as an interface.
Maybe _Curiously Recurring Template Pattern_ is a good choice better than virtual classes?

yuleisui:
What extension are you aiming for? I don't see any need for extending this.

cubicYYY:
> What extension are you aiming for? I don't see any need for extending this.

I'm trying to add more data structures, which requires manual modification of all the `if` statements involved.

PR #1524, opened 2024-08-21
Clang formatter for all source codes

cubicYYY:
All source codes are now formatted.
`clang-format.sh` is now introduced and should be run before committing.

The `.clang-format` file is updated and is used for formatting. It is now corresponding to a style like the status quo.


cubicYYY:
Some optional tasks to do in or not in this PR:

- [ ] add a workflow to check the format
- [ ] acquire contributor to run formatting script before committing in docs.

yuleisui:
Could you make the format change automatically by changing the GitHub action here: https://github.com/SVF-tools/SVF/blob/a9be281e76bb8e4f6f154414c54eb45127d6f36c/.github/workflows/svf-lib_publish.yml#L54

PR #1530, opened 2024-08-24
Clang formatter and a workflow for checking

cubicYYY:
All source codes are now formatted.
`clang-format.sh` should be run for new commits.

The `.clang-format` file is updated. It is now corresponding to a style like the status quo.
The `.clang-format-ignore` file is created for future use.

A workflow to check the format is available now.

This PR is a continuation of #1524 .

What's next:
- [ ] Resolve dependency issues in a few headers. So that we can sort includes.


clang-format.sh
```diff
@@ -0,0 +1 @@
+git ls-tree --full-tree --name-only -r HEAD | grep -e ".*\.\(c\|h\|hpp\|cc\|cpp\|hh\)\$" | grep -vf .clang-format-ignore | xargs clang-format -i -style=file --verbose
```
yuleisui, line 1: Move this line to svf-lib.publish.yml and remove this file

yuleisui:
Don not commit the formatted cpp/h files and they will be automatically done after CI.

cubicYYY:
Just wondering what is the point of formatting after CI instead of formatting the original codes.

yuleisui:
> Just wondering what is the point of formatting after CI instead of formatting the original codes.

It will be done automatically after CI. The programmer does not need to format their code beforehand.

cubicYYY:
Why bother formatting after the build and deployment? IMO formatting benefits developers, not users.

PR #1531, opened 2024-08-26
Stage2 : refactor buffer overflow

bjjwwang:


svf-llvm/tools/AE/ae.cpp
```diff
@@ -52,6 +51,9 @@ static Option<bool> AETEST(
     false
 );
 
+static Option<bool> OVERFLOW2(
```
yuleisui, line 54: move this to Options.h/cpp

svf/include/AE/Core/AbstractState.h
```diff
@@ -281,13 +283,14 @@ class AbstractState
...
-        return (s32_t) e.lb().getNumeral();
-    }
+    /**
+    * if this NodeID in SVFIR is a pointer, get the pointee type
+    * e.g  arr = (int*) malloc(10*sizeof(int))
+    *      getPointeeType(arr) -> return int
+    * we can set arr[0]='c', arr[1]='c', arr[2]='\0'
+    * @param call callnode of memset like api
+     */
+    const SVFType* getPointeeElement(NodeID id);
```
yuleisui, line 293: Do we need this? Could we get the type at the use sites?

svf-llvm/tools/AE/ae.cpp
bjjwwang: sorry. I remove it

svf/include/AE/Core/AbstractState.h
bjjwwang: yes, we need it. Maybe I can figure out a better way to get pointee type.

PR #1541, opened 2024-09-02
Nullptr dereference detector

ethanLin520:
Implemented nullptr dereference detector.


svf/lib/Util/Options.cpp
```diff
@@ -785,6 +785,8 @@ const Option<std::string> Options::OutputName(
     "output","output db file","output.db");
 const Option<bool> Options::BufferOverflowCheck(
     "overflow","Buffer Overflow Detection",false);
+const Option<bool> Options::NullPtrDereference(
+    "nullptr_dereference","Null Pointer Dereference Detection",false);
```
yuleisui, line 789: rename and simplify the option name as below.

nullptr_dereference=>nullptr

PR #1553, opened 2024-09-25
Hwg merge svf func call graph node

Geoffrey1014:
change class CallGraph & CallGraphNode to PTACallGraph &PTACallGraphNode ,
create new class CallGraph,
add callgraph field in SVFIR,
remove SVFModule::FunctionSet
CallGraphNode include a SVFFunction field


svf-llvm/lib/SVFIRBuilder.cpp
```diff
@@ -40,6 +40,8 @@
 #include "SVFIR/SVFValue.h"
 #include "Util/Options.h"
 #include "Util/SVFUtil.h"
+//#include "Graphs/CallGraph.h"
```
jumormt, line 43: delete comment

PR #1571, opened 2024-10-11
Prepare for svf var and callgraphnode downcast refactor

jumormt:


jumormt:
will upload differential test for large projects later soon

PR #1572, opened 2024-10-12
change condition value to svfvar in intracfgedge

jumormt:


svf-llvm/lib/ICFGBuilder.cpp
```diff
@@ -157,7 +157,7 @@ void ICFGBuilder::processFunBody(WorkList& worklist)
         }
         InstVec nextInsts;
         LLVMUtil::getNextInsts(inst, nextInsts);
-        u32_t branchID = 0;
+        s64_t branchID = 0;
```
yuleisui, line 160: Why s64_t here?

jumormt: I thought this would be more consistent with the field [branchCond](https://github.com/SVF-tools/SVF/pull/1572/files#diff-c2431addd23f71080f8dcaa8382750433f02885a4d2ef492aea25b107e01fb53L170) being of type s64_t.

PR #1588, opened 2024-11-05
add funcobjvar

jumormt:


jumormt:
Don't merge. PTS size of large project is not consistent with master. I'm working on it.

Issue #1591, opened 2024-11-11
`Saber`: Assertion `Val && "SVFUtil::isa<> used on a null pointer"' failed


fouzhe:
Hi, I used saber to analyze a program but encountered an assertion failure (~4min running using debug mode):

```
saber: /root/SVF/svf/include/Util/Casting.h:198: static bool SVF::SVFUtil::isa_impl_cl<To, const From*>::doit(const From*) [with To = SVF::ICFGNode; From = SVF::SVFBaseNode]: Assertion `Val && "SVFUtil:*>::doit(const From*) [with To :isa<> used on a null pointer"' failed.                    
```

I used the latest SVF and the crash happened when executing the following command:
```
saber -dfree rules_main.bc
```
The bitcode is as attached:
[rules_main.bc.zip](https://github.com/user-attachments/files/17695172/rules_main.bc.zip)

To foster debugging, I also tried to run the debug mode and got the following backtrace:

```
#0  0x00007ffff615a9fc in pthread_kill () from /lib/x86_64-linux-gnu/libc.so.6
#1  0x00007ffff6106476 in raise () from /lib/x86_64-linux-gnu/libc.so.6
#2  0x00007ffff60ec7f3 in abort () from /lib/x86_64-linux-gnu/libc.so.6
#3  0x00007ffff60ec71b in ?? () from /lib/x86_64-linux-gnu/libc.so.6
#4  0x00007ffff60fde96 in __assert_fail () from /lib/x86_64-linux-gnu/libc.so.6
#5  0x00005555559cab3e in SVF::SVFUtil::isa_impl_cl<SVF::ICFGNode, SVF::SVFBaseNode const*>::doit (Val=0x0) at /root/SVF/svf/include/Util/Casting.h:198
#6  0x00005555559c8c73 in SVF::SVFUtil::isa_impl_wrap<SVF::ICFGNode, SVF::SVFBaseNode const*, SVF::SVFBaseNode const*>::doit (Val=@0x7fffffffda50: 0x0) at /root/SVF/svf/include/Util/Casting.h:231
#7  0x00005555559c6427 in SVF::SVFUtil::isa_impl_wrap<SVF::ICFGNode, SVF::SVFBaseNode const* const, SVF::SVFBaseNode const*>::doit (Val=@0x7fffffffda98: 0x0) at /root/SVF/svf/include/Util/Casting.h:220
[...]

PR #1594, opened 2024-11-13
fix npd issue#1591

jumormt:


jumormt:
shoud fix issue #1591

Issue #1592, opened 2024-11-12
`Saber`: Assertion `SVFUtil::isa<X>(Val) && "cast<Ty>() argument of incompatible type!"' failed. 


fouzhe:
Hi, I encountered an assertion failure when executing `Saber` to analyze this bitcode file [affdicmain.bc.zip](https://github.com/user-attachments/files/17709944/affdicmain.bc.zip):

```
saber -dfree ./affdicmain.bc
```
Output:
```
...
Build Sparse Value-Flow Graph                              
        Create VFG Top Level Node                          
        Create SVFG Direct Edge                            
Build Sparse Value-Flow Graph 
        Create SVFG Addr-taken Node                                                                                   
        Create SVFG Indirect Edge                                                                                     
        Collect Global Variables
        Remove Dereference Direct SVFG Edge
        Add Sink SVFG Nodes
                             
*********SVFG Statistics***************
################ (program : affdicmain.bc)###############
----------------Time and memory stats--------------------
ATNodeTime          0.071
AvgWeight           9.55292
ConnDirEdgeTime     0   
ConnIndEdgeTime     0.161 
OptTime             0     
TLNodeTime          0                                      
TotalTime           0.301    
----------------Numbers stats----------------------------
ActualIn            11875                                  
ActualOut           5045                                   
ActualParam         16591
ActualRet           3347   
Addr                10284
AvgInDegree         1     
AvgIndInDeg         1     
AvgIndOutDeg        2                                      
AvgOutDegree        1                                      
Copy                797                                                                                               
DirectCallEdge      11046            
DirectEdge          48317                                                                                             
DirectRetEdge       2579 
FormalIn            7862
FormalOut           7862
FormalParam         3415
FormalRet           622
Gep                 9207
IndCallEdge         12647
IndRetEdge          5448
IndirectEdge        81231
IndirectEdgeLabels  775993
Load                10524
MSSAPhi             12405
MaxInDegree         396
MaxIndInDeg         396
MaxIndOutDeg        330
MaxOutDegree        1257
PHI                 656
Store               6785
TotalEdge           129548
TotalNode           107278
#######################################################

*********SVFG Statistics***************                                                                      [19/1885]
################ (program : affdicmain.bc)###############
----------------Time and memory stats--------------------
ATNodeTime          0.071
AvgWeight           9.54302
ConnDirEdgeTime     0   
ConnIndEdgeTime     0.161 
OptTime             0     
TLNodeTime          0                                      
[...]

PR #1595, opened 2024-11-13
remove casting to intraicfgnode when allocating branch condition

jumormt:
fix issue #1592


jumormt:
This case is caused by exception handling in c++, eg, ``invoke void %3(ptr noundef nonnull align 8 dereferenceable(8) %1, ptr noundef %2) to label %invoke.cont1 unwind label %lpad''. Here the callsite has two successors with one being the exception handling code.

PR #1638, opened 2025-01-23
remove SVFFileSystem

bjjwwang:


svf-llvm/tools/LLVM2SVF/llvm2svf.cpp
```diff
@@ -61,12 +61,12 @@ int main(int argc, char** argv)
         LLVMModuleSet::preProcessBCs(moduleNameVec);
     }
 
-    SVFModule* svfModule = LLVMModuleSet::buildSVFModule(moduleNameVec);
+    //SVFModule* svfModule = LLVMModuleSet::buildSVFModule(moduleNameVec);
```
yuleisui, line 64: Why comment this?

bjjwwang: because unused var would cause compile error.

svf-llvm/lib/SVFIRBuilder.cpp
```diff
@@ -169,11 +169,6 @@ SVFIR* SVFIRBuilder::build()
         loopAnalysis.build(pag->getICFG());
     }
 
-    // dump SVFIR as JSON
```
yuleisui, line 172: Should comment out rather than removing

bjjwwang: > Should comment out rather than removing

sure

svf-llvm/tools/LLVM2SVF/llvm2svf.cpp
yuleisui: Why unused? Which variable?

Issue #1680, opened 2025-03-26
Extapi getenv handling


mortkever:
Hello, 

What is the intended handling of external functions like getenv(const char *name)? Currently get_env is annotated as "ALLOC_HEAP_RET" in the new method and as "EFT_NOSTRUCT_ALLOC" in the old method which seem inaccurate. This implies getenv returns a pointer to a newly allocated space on the heap/stack while getenv returns a pointer to the same memory when the argument is the same. So the same argument always returns the same pointer but different arguments return different pointers.

How is SVF supposed to handle functions like these? Would it be okay to annotate getenv as static and just consider all possible returns part of a larger struct?
Thanks!

yuleisui:
This is a good observation. Yes, the same string as an argument of `getenv` should always return the same piece of memory. This requires string matching. We will need to add some implementation to support this. However, the input argument may not always be a constant, hence a string analysis (including data-flow of the arguments and all the string operation statements) is needed. The most conservative way is to use the static shared memory (global object).

mortkever:
Great, I’ll look into making a pull request for the conservative approach. 
I do see that there are more static return functions that don't have the static annotation (errno, localeconv, localtime, gmtime...). And that all signs of static returns were [removed ](https://github.com/SVF-tools/SVF/commit/59a840f34dd96ae2e04b40436d677a0533a82053) a few months ago. What is the current intended way to handle this?

yuleisui:
> Great, I’ll look into making a pull request for the conservative approach. I do see that there are more static return functions that don't have the static annotation (errno, localeconv, localtime, gmtime...). And that all signs of static returns were [removed ](https://github.com/SVF-tools/SVF/commit/59a840f34dd96ae2e04b40436d677a0533a82053) a few months ago. What is the current intended way to handle this?

@shuangxiangkan could you help with and reply this issue.

adriaanjacobs:
It seems that SVF has gone through quite significant changes with respect to its handling of functions that return statically-allocated objects. Me and @mortkever have looked into the history here:

In the old JSON format, and even shortly after migrating to the new C-style format, SVF handled them via `annotate("STATIC")` (until SVF 2.7). This was perfectly sound & workable, am not aware of any reported issues here. 

[...]

PR #1688, opened 2025-04-09
Fix statically allocated objects

adriaanjacobs:
Hi

As discussed in #1680, this patch largely reverts #1616, which mistakenly seems to refer to statically-allocated objects as "stack-allocated objects". As far as we can tell, this lead to only 1 bug in SVF, in the SVFIRBuilder, where a regular `isa<Alloca>` was replaced with a function to check for a call to an extapi func returning "stack-allocated storage" (which actually returned an unknown static object). That's obviously wrong. Fixed in [4d7ece4e5b4](https://github.com/SVF-tools/SVF/commit/4d7ece4e5b4b1d8049601fefa6aba2e858e292d9).

However, we (surprisingly!) did not notice any build failures after updating this naming convention and deleting some utility functions. Does that mean that SVF does not check for statically-allocated objects in its handling of external calls? I'm confused here: where should we find & update the code that treats all calls to such functions as aliasing?

I believe there's more work to be done to properly handle these STATIC_RET functions. @mortkever/@shuangxiangkan  any ideas?

### DIFF
--- a/svf-llvm/include/SVF-LLVM/LLVMModule.h
+++ b/svf-llvm/include/SVF-LLVM/LLVMModule.h
@@ -414,7 +414,7 @@ public:
     bool is_arg_alloc(const Function *F);
 
     // Does (F) allocate a new stack object and return it?
-    bool is_alloc_stack_ret(const Function *F);
+    bool is_static_ret(const Function *F);
 
     // Get the position of argument which holds the new object
     s32_t get_alloc_arg_pos(const Function *F);
--- a/svf-llvm/include/SVF-LLVM/LLVMModule.h
+++ b/svf-llvm/include/SVF-LLVM/LLVMModule.h
@@ -413,7 +413,7 @@ public:
     // Does (F) allocate a new object and assign it to one of its arguments?
     bool is_arg_alloc(const Function *F);
 
-    // Does (F) allocate a new stack object and return it?
+    // Does (F) return a pointer to an unknown object with static lifetime?
     bool is_static_ret(const Function *F);
 
     // Get the position of argument which holds the new object
--- a/svf-llvm/include/SVF-LLVM/LLVMUtil.h
+++ b/svf-llvm/include/SVF-LLVM/LLVMUtil.h
@@ -398,18 +398,8 @@ inline bool isHeapAllocExtCall(const Instruction *inst)
     return isHeapAllocExtCallViaRet(inst) || isHeapAllocExtCallViaArg(inst);
 }
 
-bool isStackAllocExtCallViaRet(const Instruction *inst);
-
-inline bool isStackAllocExtCall(const Instruction *inst)
-{
-    return isStackAllocExtCallViaRet(inst);
-}
-
 // Check if a given value represents a heap object.
 bool isHeapObj(const Value* val);
-
-// Check if a given value represents a stack object.
-bool isStackObj(const Value* val);
 
 /// Whether an instruction is a callsite in the application code, excluding llvm intrinsic calls
 bool isNonInstricCallSite(const Instruction* inst);
--- a/svf-llvm/lib/LLVMModule.cpp
+++ b/svf-llvm/lib/LLVMModule.cpp
@@ -1466,9 +1466,9 @@ bool LLVMModuleSet::is_arg_alloc(const Function* F)
     return F && hasExtFuncAnnotation(F, "ALLOC_HEAP_ARG");
 }
 
-bool LLVMModuleSet::is_alloc_stack_ret(const Function* F)
+bool LLVMModuleSet::is_static_ret(const Function* F)
 {
-    return F && hasExtFuncAnnotation(F, "ALLOC_STACK_RET");
+    return F && hasExtFuncAnnotation(F, "STATIC_RET");
 }
 
 // Get the position of argument which holds the new object
--- a/svf-llvm/lib/LLVMUtil.cpp
+++ b/svf-llvm/lib/LLVMUtil.cpp
@@ -659,20 +659,6 @@ bool LLVMUtil::isHeapAllocExtCallViaArg(const Instruction* inst)
     }
 }
 
-bool LLVMUtil::isStackAllocExtCallViaRet(const Instruction *inst)
-{
-    LLVMModuleSet* pSet = LLVMModuleSet::getLLVMModuleSet();
-    bool isPtrTy = inst->getType()->isPointerTy();
-    if (const CallBase* call = SVFUtil::dyn_cast<CallBase>(inst))
-    {
-        const Function* fun = call->getCalledFunction();
-        return fun && isPtrTy &&
-               pSet->is_alloc_stack_ret(fun);
-    }
-    else
-        return false;
-}
-
 /**
  * Check if a given value represents a heap object.
  *
@@ -690,26 +676,6 @@ bool LLVMUtil::isHeapObj(const Value* val)
     // Check if the value is an instruction and if it is a heap allocation external call
     else if (SVFUtil::isa<Instruction>(val) &&
              LLVMUtil::isHeapAllocExtCall(SVFUtil::cast<Instruction>(val)))
-    {
-        return true;
-    }
-    // Return false if none of the above conditions are met
-    return false;
-}
-
-/**
- * @param val The value to check.
- * @return True if the value represents a stack object, false otherwise.
- */
-bool LLVMUtil::isStackObj(const Value* val)
-{
-    if (SVFUtil::isa<AllocaInst>(val))
-    {
-        return true;
-    }
-    // Check if the value is an instruction and if it is a stack allocation external call
-    else if (SVFUtil::isa<Instruction>(val) &&
-             LLVMUtil::isStackAllocExtCall(SVFUtil::cast<Instruction>(val)))
     {
         return true;
     }
--- a/svf-llvm/lib/SVFIRBuilder.cpp
+++ b/svf-llvm/lib/SVFIRBuilder.cpp
@@ -404,7 +404,7 @@ void SVFIRBuilder::initialiseBaseObjVars()
             pag->addHeapObjNode(iter->second, pag->getObjTypeInfo(id), llvmModuleSet()->getSVFType(llvmValue->getType()), icfgNode);
         }
         // Check if the value is an alloca instruction and add a stack object node
-        else if (LLVMUtil::isStackObj(llvmValue))
+        else if (SVFUtil::isa<AllocaInst>(llvmValue))
         {
             NodeID id = llvmModuleSet()->getObjectNode(iter->first);
             pag->addStackObjNode(iter->second, pag->getObjTypeInfo(id), llvmModuleSet()->getSVFType(llvmValue->getType()), icfgNode);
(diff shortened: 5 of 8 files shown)


yuleisui:
@adriaanjacobs I finally had some time to review this pull request. `ALLOC_STACK_RET` (stack lifetime) and `STATIC_RET` (program lifetime) represent two different types of objects. I would suggest retaining both. Are you certain that all `ALLOC_STACK_RET` annotations should be replaced with `STATIC_RET`? We may need to keep both for future annotations and extend external APIs.

I don't think we have to change this line [here](https://github.com/SVF-tools/SVF/pull/1688/files#diff-9602be4b48c87310b723b780876f16174eeda0abff0db15f26e8ab0e9abd0d00R407), as `isStackObj` makes more sense to me if we keep these two types of annotations. 

Please let me know your thoughts and update the patch accordingly.

Thanks.

adriaanjacobs:
Thanks for taking a look.

I don't think I fully understand what the `ALLOC_STACK_RET` is supposed to mean. How can functions return stack-allocated data? Is that not always a stack-use-after-free? E.g.
```c
void* func() {
    char data[SIZE];
    return &data[0];
}
```

I've confirmed that all changed ExtAPI functions were previously (before #1616)  `STATIC_OBJECT` functions, so I'm not sure which function in ExtAPI is an example of a function SVF would like to handle as `ALLOC_STACK_RET`? Do you happen to know what the original intention behind #1616 was, i.e., for which ExtAPI function was it introduced?

yuleisui:
> Thanks for taking a look.
> 
> I don't think I fully understand what the `ALLOC_STACK_RET` is supposed to mean. How can functions return stack-allocated data? Is that not always a stack-use-after-free? E.g.
> 
> ```c
> void* func() {
>     char data[SIZE];
>     return &data[0];
> }
> ```
> 
> I've confirmed that all changed ExtAPI functions were previously (before #1616) `STATIC_OBJECT` functions, so I'm not sure which function in ExtAPI is an example of a function SVF would like to handle as `ALLOC_STACK_RET`? Do you happen to know what the original intention behind #1616 was, i.e., for which ExtAPI function was it introduced?

Some c standard library API like `void* alloca(size_t size);` can return a stack object which is similar to LLVM's alloca instruction. Hence we may need to keep `ALLOC_STACK_RET`, `isStackObj`, `is_static_ret` methods and so on.

I had another look at `extapi.c`, I agree that all of the current `ALLOC_STACK_RET` annotations should be `STATIC_OBJECT`. But we need to handle each API call and allocate for a new global/static object by changing a new condition [here](https://github.com/SVF-tools/SVF/pull/1688/files#diff-9602be4b48c87310b723b780876f16174eeda0abff0db15f26e8ab0e9abd0d00R427):

```
else if (SVFUtil::isa<GlobalValue>(llvmValue) || is_static_ret_obj(llvmValue))
{

}

```

Let me know your thoughts.

shuangxiangkan:
> Thanks for taking a look.
> 
> I don't think I fully understand what the `ALLOC_STACK_RET` is supposed to mean. How can functions return stack-allocated data? Is that not always a stack-use-after-free? E.g.
> 
> ```c
> void* func() {
>     char data[SIZE];
>     return &data[0];
> }
> ```
> 
> I've confirmed that all changed ExtAPI functions were previously (before #1616) `STATIC_OBJECT` functions, so I'm not sure which function in ExtAPI is an example of a function SVF would like to handle as `ALLOC_STACK_RET`? Do you happen to know what the original intention behind #1616 was, i.e., for which ExtAPI function was it introduced?

The original design didn't distinguish between external functions allocating static objects or stack objects—it essentially treated them all as stack objects. So, simply reverting #1616 isn't enough, which is also why you didn't encounter build failures in #1688, as static object allocation for external functions wasn't addressed.

yuleisui:
> > Thanks for taking a look.
> > I don't think I fully understand what the `ALLOC_STACK_RET` is supposed to mean. How can functions return stack-allocated data? Is that not always a stack-use-after-free? E.g.
> > ```c
> > void* func() {
> >     char data[SIZE];
> >     return &data[0];
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I've confirmed that all changed ExtAPI functions were previously (before #1616) `STATIC_OBJECT` functions, so I'm not sure which function in ExtAPI is an example of a function SVF would like to handle as `ALLOC_STACK_RET`? Do you happen to know what the original intention behind #1616 was, i.e., for which ExtAPI function was it introduced?
> 
> Some c standard library API like `void* alloca(size_t size);` can return a stack object which is similar to LLVM's alloca instruction. Hence we may need to keep `ALLOC_STACK_RET`, `isStackObj`, `is_static_ret` methods and so on.
> 
> I had another look at `extapi.c`, I agree that all of the current `ALLOC_STACK_RET` annotations should be `STATIC_OBJECT`. But we need to handle each API call and allocate for a new global/static object by changing a new condition [here](https://github.com/SVF-tools/SVF/pull/1688/files#diff-9602be4b48c87310b723b780876f16174eeda0abff0db15f26e8ab0e9abd0d00R427):
> 
> ```
> else if (SVFUtil::isa<GlobalValue>(llvmValue) || is_static_ret_obj(llvmValue))
> {
> 
> }
> ```
> 
> Let me know your thoughts.

@adriaanjacobs could you check the above the update the pull request and let me know if any questions

mortkever:
> > Thanks for taking a look.
> > I don't think I fully understand what the `ALLOC_STACK_RET` is supposed to mean. How can functions return stack-allocated data? Is that not always a stack-use-after-free? E.g.
> > ```c
> > void* func() {
> >     char data[SIZE];
> >     return &data[0];
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I've confirmed that all changed ExtAPI functions were previously (before #1616) `STATIC_OBJECT` functions, so I'm not sure which function in ExtAPI is an example of a function SVF would like to handle as `ALLOC_STACK_RET`? Do you happen to know what the original intention behind #1616 was, i.e., for which ExtAPI function was it introduced?
> 
> Some c standard library API like `void* alloca(size_t size);` can return a stack object which is similar to LLVM's alloca instruction. Hence we may need to keep `ALLOC_STACK_RET`, `isStackObj`, `is_static_ret` methods and so on.
> 
> I had another look at `extapi.c`, I agree that all of the current `ALLOC_STACK_RET` annotations should be `STATIC_OBJECT`. But we need to handle each API call and allocate for a new global/static object by changing a new condition [here](https://github.com/SVF-tools/SVF/pull/1688/files#diff-9602be4b48c87310b723b780876f16174eeda0abff0db15f26e8ab0e9abd0d00R427):
> 
> ```
> else if (SVFUtil::isa<GlobalValue>(llvmValue) || is_static_ret_obj(llvmValue))
> {
> 
> }
> ```
> 
> Let me know your thoughts.

If i understand correctly, you proposal is to create a new node for each call to a "static return" function?

We think there are two requirements SVF should fulfil to correctly handle these functions.
Firstly SVF should take into account that all pointers returned from the same function should all alias. I wrote tests for this for the SVF testing benchmark and as expected they failed. The solution proposed by us in #1680 did not work either.
Secondly the nodes that end up in the points-to sets should still map to existing llvm values. 
A bit contradictory, but other methods to access all call sites of a function from an SVF node could substitute the second requirement. Perhaps just keeping a map between nodes of static return functions and sets of llvm values of the call site could suffice.

We don't think creating a new node for each call site is the correct approach as it satisfies the second requirement but not the first. All the nodes would still be seen as different nodes and thus not alias, if I'm not mistaken.

yuleisui:
> > > Thanks for taking a look.
> > > I don't think I fully understand what the `ALLOC_STACK_RET` is supposed to mean. How can functions return stack-allocated data? Is that not always a stack-use-after-free? E.g.
> > > ```c
> > > void* func() {
> > >     char data[SIZE];
> > >     return &data[0];
> > > }
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > I've confirmed that all changed ExtAPI functions were previously (before #1616) `STATIC_OBJECT` functions, so I'm not sure which function in ExtAPI is an example of a function SVF would like to handle as `ALLOC_STACK_RET`? Do you happen to know what the original intention behind #1616 was, i.e., for which ExtAPI function was it introduced?
> > 
> > 
> > Some c standard library API like `void* alloca(size_t size);` can return a stack object which is similar to LLVM's alloca instruction. Hence we may need to keep `ALLOC_STACK_RET`, `isStackObj`, `is_static_ret` methods and so on.
> > I had another look at `extapi.c`, I agree that all of the current `ALLOC_STACK_RET` annotations should be `STATIC_OBJECT`. But we need to handle each API call and allocate for a new global/static object by changing a new condition [here](https://github.com/SVF-tools/SVF/pull/1688/files#diff-9602be4b48c87310b723b780876f16174eeda0abff0db15f26e8ab0e9abd0d00R427):
> > ```
> > else if (SVFUtil::isa<GlobalValue>(llvmValue) || is_static_ret_obj(llvmValue))
> > {
> > 
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Let me know your thoughts.
> 
> If i understand correctly, you proposal is to create a new node for each call to a "static return" function?
> 
> We think there are two requirements SVF should fulfil to correctly handle these functions. Firstly SVF should take into account that all pointers returned from the same function should all alias. I wrote tests for this for the SVF testing benchmark and as expected they failed. The solution proposed by us in #1680 did not work either. Secondly the nodes that end up in the points-to sets should still map to existing llvm values. A bit contradictory, but other methods to access all call sites of a function from an SVF node could substitute the second requirement. Perhaps just keeping a map between nodes of static return functions and sets of llvm values of the call site could suffice.
> 
> We don't think creating a new node for each call site is the correct approach as it satisfies the second requirement but not the first. All the nodes would still be seen as different nodes and thus not alias, if I'm not mistaken.

yuleisui:
@adriaanjacobs 

Very good points! We may need two types annotations for distinguish whether a extAPI returns a shared or an individually created object node. SHARED_STATIC_OBJECT or INDIVIDUAL_STATIC_OBJECT

To satisfy your Requirement 1:  SHARED_STATIC_OBJECT should be used for each ext function (one llvm function value to one shared static/global object). To satisfy your Requirement 2: for mapping a node back to llvm value, I think we need each SHARED_STATIC_OBJECT to be an individually defined global variable in the extapi.c.

Let me know your thoughts.

I am unsure which test case you mentioned failed. But Requirement 1 does introduce conservative and spurious results.

shuangxiangkan:
In [#1693](https://github.com/SVF-tools/SVF/pull/1693), we removed the "ALLOC_STACK_RET" annotation for APIs that return statically allocated objects. Instead, we now handle these cases by introducing a global variable for each such API, which allows us to model shared static objects. We also retain the "ALLOC_STACK_RET" annotation for APIs that are expected to return a newly allocated object on each call.

@mortkever and @adriaanjacobs, does this work for you? or any other thoughts or ideas?

PR #1697, opened 2025-05-02
Build, CI, Formatting, API & SVFG Builder Improvements

Johanmyst:
## 📝 Overview
This PR consolidates six orthogonal areas of work into one release:

- **Build System**: make installs relocatable, expose ABI flags, support in- and out-of-tree usage  
- **CI/Workflow**: harden coverage removal flags in GitHub Actions  
- **CMake Formatting**: add `cmake-format` config for consistent styling & formatting  
- **API Additions**: add missing `has…()` checks alongside existing getters  
- **WPA Backends**: unify public interface for polymorphic use  
- **SVFG Builder**: fix read/write order for optimized graphs  

_For full details on each change, see the linked commits below._

If you'd like, I could add a section to the [setup guide wiki page](https://github.com/svf-tools/SVF/wiki/Setup-Guide#getting-started) to explain installing & using SVF's CMake packages. Additionally, I could also add a new page detailing how to use SVF from within an out-of-tree LLVM pass directly from the default clang pass-pipeline (i.e., analyse & possibly modify a module using information from SVF's analyses during regular compilation (by simply adding a `--Wl,--load-pass-plugin=libSVFUser.so`)?

---

## 🚀 Highlights

| Area                  | Summary                                                  | Commit                         |
|-----------------------|----------------------------------------------------------|-------------------------------|
| **Build System**      | Relocatable installs, ABI-flag interface, in-/out-of-tree mode    | [583bff0](https://github.com/SVF-tools/SVF/commit/583bff052ba8fd480d5809d46630a4b74ca08b19) |
| **CI/Workflow**       | Uniform `--ignore-errors unused` on all `lcov --remove`  | [96c021b](https://github.com/SVF-tools/SVF/commit/96c021bef0dabe09df966039b9eeff4bd7716618) |
| **CMake Formatting**  | Add `cmake-format` spec for automatic styling            | [0bc1472](https://github.com/SVF-tools/SVF/commit/0bc1472586cf31258baf769060b90f924a4e5332) |
| **API Existence**     | New `has…()` functions for previously unchecked getters  | [ce288cf](https://github.com/SVF-tools/SVF/commit/ce288cf013171c4b0eba9b40d0fa8c18cba8ad6e) |
| **WPA Polymorphism**  | Publicise ctor/read/dump APIs in `VersionedFlowSensitive`| [93446f6](https://github.com/SVF-tools/SVF/commit/93446f684ad0600854714359434256aed2c9bbed) |
| **SVFG Builder**      | Write unoptimised graph before optimise; explicit flag  | [def7c04](https://github.com/SVF-tools/SVF/commit/def7c043fe32b6bee0b94c68d2de95b3f0b7d2df) |

------

### 1️⃣ Build System ([583bff0])
- **Directory layout**: moved all CMake templates into `cmake/`  
- **Relocatable install**: properly support relocation correctly (`cmake --install … --install-prefix=<prefix>`)  
- **ABI flags**: expose minimum C++ standard, PIC, etc., via an (invisible) interface library  
- **pkgconfig export**: generate `.pc` with core flags (less complete than the CMake package)  
- **In-tree usage**: `add_subdirectory(SVF)` now works transparently  
- **LTO guard**: error out if compiler doesn’t support LTO  
- **Dependency flags**: propagate Z3/LLVM ABI flags on `SVF::Core` & `SVF::LLVM`  
- **Conditional version export**: only set `SVF_VERSION` in parent scope when parent CMake project exists  

### 2️⃣ CI/Workflow Fix ([96c021b])
- Always pass `--ignore-errors unused` to **all** `lcov --remove …` calls  
- Prevent workflow failures when no coverage data exists under `/usr/*` or other paths  
- Ensured error suppression is applied consistently  

### 3️⃣ Automated CMake Formatting ([0bc1472])
- Add `.cmake-format.py` so `cmake-format` can lint & format all CMake files  
- Ensures consistent styling across `CMakeLists.txt`, `*.cmake`, etc.  

### 4️⃣ Existence-Check APIs ([ce288cf])
- Introduce `hasLLVMValue()`, `hasLHSTopLevelPtr()`, etc., for getters that assert on missing entries  
- Avoids breaking existing API: getters still assert, but you can pre-check  
- Getters that return nullptr on missing entries are unmodified  

### 5️⃣ WPA Backend Polymorphism ([93446f6])
- Mark `VersionedFlowSensitive`’s `printStat()`, `dumpTopLevelPtsTo()`, `solveAndWritePtsToFile()`, `readPtsFromFile()`, etc., as **public**  
- Enables code like:
  ```cpp
  SVF::BVDataPTAImpl *pta = nullptr;
  switch(backend) {
    case Andersen_BASE:   pta = new Andersen(...);            break;
    case VFS_WPA:         pta = new VersionedFlowSensitive(...); break;
    default:              /*…*/;                              break;
  }

### 6️⃣ Optimised SVFG Read/Write Fix ([def7c04])
- Add builder argument to build optimised SVFG without command-line option (to support using SVF as a library)  
- Write unoptimised SVFG before optimisation  
- Read unoptimised SVFG, then apply optimisation  
- Resolves crashes when reading a post-optimised file


yuleisui:
Thanks for the contribution. Could you split this pull request into two?

1. Build system
2. All others 

I had a quick look at 2, all look good, and should be ready to merge.

1 looks a bit complicated and I will have to go through it separately.

Cheers

Johanmyst:
> Thanks for the contribution. Could you split this pull request into two?

Sure, give me a bit to figure out what the easiest way to do this would be!

> I had a quick look at 2, all look good, and should be ready to merge.

Great, happy to hear!

> 1 looks a bit complicated and I will have to go through it separately.

Makes sense. I did make sure to not break any backwards compatibility; all current ways of using SVF should continue to work unchanged. The only difference is that the exported CMake package now properly handles relocatability and dependency inheritance.

For example, the current CMake package doesn't properly expose dependencies and necessary ABI flags on the actual targets. As a result, linking against SVF's package after finding system instances (e.g., with `find_package(SVF)`) can still experience issues with this.

For example, say I want to use system-wide installations of Z3 & LLVM to build & link SVF against. I could install Z3 & LLVM like this:

```shell
# Z3 from official Ubuntu repositories
> sudo apt install z3 libz3-4 libz3-dev

# Ubuntu 24.04 (Noble) has LLVM-16 in the official repos as well
> sudo apt install llvm-16 llvm-16-dev llvm-16-tools \
    clang-16 libclang-16-dev libclang-rt-16-dev \
    lld-16 liblld-16 liblld-16-dev \
    ...
```

This would install these libraries in `/usr/...` and into the default paths such that any `CMakeLists.txt` script could do `find_package(Z3)` and `find_package(LLVM)` to find their CMake packages. To link against them --- and automatically inherit all required dependencies, ABI flags, etc. --- one can simply call `target_link_libraries(myLib PRIVATE libz3)`.

These changes support using SVF in the same way. For example, I can build & install SVF globally like this:

```shell
> git clone https://github.com/SVF-tools/SVF.git ./SVF
> cd SVF
> cmake -S ./ \
    -G "Ninja" \
    -B Release-build \
    -DCMAKE_BUILD_TYPE=Release \
    -DCMAKE_INSTALL_PREFIX=/usr/local \
    -DSVF_ENABLE_ASSERTIONS=ON \
    -DSVF_EXPORT_DYNAMIC=ON \
    -DBUILD_SHARED_LIBS=ON \
    -DSVF_DEBUG_INFO=ON \
    -DSVF_USE_LLD=ON
> cmake --build Release-build
> cmake --install Release-build [--install-prefix=/some/other/prefix]
```

This would install SVF's headers, CMake package, etc. to the chosen prefix so that the tools end up in places like `/usr/local/bin/wpa` and `/usr/local/lib/libSvfCore.so`. If I now wanted to use this global SVF instance in another project, I could simply find & link it in the same way SVF finds LLVM and Z3:

`// File: CMakeLists.txt:`
```cmake
cmake_minimum_required(VERSION 3.28)
set(CMAKE_EXPORT_COMPILE_COMMANDS ON)

project(Foo CXX)
add_executable(foo foo.cpp)
find_package(SVF 3.0 REQUIRED CONFIG)
target_link_libraries(foo PRIVATE SVF::SvfCore SVF::SvfLLVM)
install(TARGETS foo RUNTIME DESTINATION bin)
```

`// File: foo.cpp:`
```cpp
#include <MSSA/SVFGBuilder.h>
#include <SVF/Graphs/SVFG.h>
#include <SVF/SVF-LLVM/LLVMModule.h>
#include <SVF/SVF-LLVM/SVFIRBuilder.h>
#include <SVF/Util/CommandLine.h>
#include <SVF/Util/Options.h>
#include <SVF/WPA/Andersen.h>

#include <cstdlib>
#include <string>
#include <vector>

using namespace llvm;
using namespace std;
using namespace SVF;

  std::vector<std::string> modNameVec = OptionBase::parseOptions(
      argc, argv, "bitcodes", "[options] <bitcode...>");

  if (Options::WriteAnder() == "ir_annotator")
    LLVMModuleSet::preProcessBCs(modNameVec);
  LLVMModuleSet::buildSVFModule(modNameVec);

  SVFIR *pag = SVFIRBuilder().build();
  Andersen *ander = AndersenWaveDiff::createAndersenWaveDiff(pag);
  SVFG *svfg = SVFGBuilder().buildFullSVFG(ander);

  svfg->dump("input");

  AndersenWaveDiff::releaseAndersenWaveDiff();
  SVF::LLVMModuleSet::releaseLLVMModuleSet();
  SVFIR::releaseSVFIR();
  llvm::llvm_shutdown();

  return EXIT_SUCCESS;
}
```

I could then build, install, and run my tool like this:

```shell
> cmake -G "Ninja" -B build
> cmake --build build
> cmake --install build --prefix $(pwd)/install
> ./install/bin/foo some/example/bitcode.bc
```

The `find_package(SVF)` snippet finds the SVF CMake package. Internally, this package defines its dependencies, but also exposes its dependencies' dependencies (e.g., ABI flags required by the specific LLVM instance used, like `-fno-rtti`). Consequently, to use SVF, I don't need to also find LLVM and/or Z3 and link against them — CMake resolves that based on SVF's package. This also works for non-system wide installations of SVF (e.g., installed to `~/SVF/install/...`) by setting the environment variable `$SVF_DIR` or passing it as a search path to the `find_package(SVF HINTS <path>)` invocation.

Of course, since the `Release-build` tree still exists as it did before, finding & linking against SVF how most test cases and examples do currently is still supported. Moreover, while headers are now cleanly installed into `<prefix>/include/SVF/...`, using the headers without the prefix (e.g., `#include <WPA/Andersen.h>`) is still possible (this is also exposed by the CMake package implicitly).

Moreover, after the `find_package(SVF)` call, its configuration variables become available, so a user can directly use CMake variables like `${SVF_USE_PIC}` to check whether SVF was built with/without PIC, `${SVF_BUILD_TYPE}` to check the used build type, or install paths (even when the installation prefix is changed after CMake configuration (i.e., during installation with `cmake --install <build_dir> --prefix=some/new/install/prefix`) like `${SVF_INSTALL_BINDIR}` or `${SVF_INSTALL_INCLUDEDIR}`.

Moreover, this keeps the dependencies "separate", ensuring only the `SvfCore` library depends on LLVM. In the future, this could easily be modified to ensure LLVM needn't be found if someone only wants to use & link against `SvfCore` (but not `SvfLLVM`).

I hope this clarifies the changes & use-case a bit more! As I said, I'd be happy to update the Wiki page to describe how to build, install, and use SVF and its dependencies like this, as well as write a new Wiki page describing how to use SVF within the default LLVM clang pass-pipeline! Would you still like me to split this pull request (no problem), or should I leave it combined if this makes enough sense to merge as-is?

Johanmyst:
Rebased from the current upstream's head to make merging easier. @yuleisui, let me know if you'd still like me to separate this pull request or if my previous comment sufficiently clarified my intent/use-case and you want to merge it as-is!

yuleisui:
> Rebased from the current upstream's head to make merging easier. @yuleisui, let me know if you'd still like me to separate this pull request or if my previous comment sufficiently clarified my intent/use-case and you want to merge it as-is!

Thanks, and could you separate the pull request? I can merge your "non-Build System" one first.

PR #1713, opened 2025-05-23
Changed types of SVF install prefix configuration parameters to STRING to avoid absolute path prefixes

Johanmyst:
After my previous pull request #1703 was merged, commit a745cdc changed the default installation prefix for public headers from `include/SVF` to just `include`. As discussed in [my last comment](https://github.com/SVF-tools/SVF/issues/1703#issuecomment-2901506766_) on pull request #1703, installing into a dedicated project prefix can often be desirable for clarity and maintenance. To make this behaviour optional, I included user-configurable build parameters like `SVF_INSTALL_INCLUDEDIR` and `SVF_INSTALL_BINDIR` to override the default installation prefixes.

However, I made a mistake in defining the type of these parameters; they are currently defined as `PATH` types instead of `STRING` types. As these defaulted to their `${CMAKE_INSTALL_<...>}` counterparts (which are always correctly interpreted as **relative to the installation prefix**, this would ensure the final installation destination would be correct (i.e., `<install_prefix>/include/SVF/...`, `<install_prefix>/bin/...`, etc.). However, defining these types as `PATH` types would cause them to be incorrectly resolved into **absolute paths**. Consequently, the current working directory would be internally prepended to these installation prefix parameters, causing the final installation destinations to look like `<pwd>/<user_config_prefix>` (e.g., `$(pwd)/include/SVF`). Consequently, the source/install paths to contain their own prefixes, causing the build configuration to fail.

This small fix simply uses the `STRING` type for these installation prefix parameters (e.g., `SVF_INSTALL_INCLUDEDIR`) instead, ensuring they are always correctly interpreted as **relative paths** (relative to the overall installation prefix).

NB: I would still argue defaulting to the installation prefix for public header files being `include/SVF` instead of the bare `include`, and/or properly describing the possible build parameters in the setup guide. Should I add that to the [Wiki page for that](https://github.com/svf-tools/SVF/wiki/Setup-Guide#getting-started)? If so, is my understanding correct that anyone can modify the wiki without requiring explicit permissions?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -117,27 +117,27 @@ endif()
 # Configure top-level SVF variables (used by CMake for configuring installed SVF package)
 set(SVF_INSTALL_BINDIR
     ${CMAKE_INSTALL_BINDIR}
-    CACHE PATH "Set binaries install dir"
+    CACHE STRING "Set binaries install dir"
 )
 set(SVF_INSTALL_LIBDIR
     ${CMAKE_INSTALL_LIBDIR}
-    CACHE PATH "Set libraries install dir"
+    CACHE STRING "Set libraries install dir"
 )
 set(SVF_INSTALL_EXTAPIDIR
     ${CMAKE_INSTALL_LIBDIR}
-    CACHE PATH "Set extapi.bc install dir"
+    CACHE STRING "Set extapi.bc install dir"
 )
 set(SVF_INSTALL_INCLUDEDIR
     ${CMAKE_INSTALL_INCLUDEDIR}
-    CACHE PATH "Set public headers install dir"
+    CACHE STRING "Set public headers install dir"
 )
 set(SVF_INSTALL_PKGCONFDIR
     ${CMAKE_INSTALL_LIBDIR}/pkgconfig
-    CACHE PATH "Override pkgconfig install dir"
+    CACHE STRING "Override pkgconfig install dir"
 )
 set(SVF_INSTALL_CMAKECONFIGDIR
     ${CMAKE_INSTALL_LIBDIR}/cmake/SVF
-    CACHE PATH "Set CMake package install dir"
+    CACHE STRING "Set CMake package install dir"
 )
 
 # Set location of extapi.bc (installed)


yuleisui:
> After my previous pull request #1703 was merged, commit [a745cdc](https://github.com/SVF-tools/SVF/commit/a745cdc88af37f4395566d3475fb23d718874930) changed the default installation prefix for public headers from `include/SVF` to just `include`. As discussed in [my last comment](https://github.com/SVF-tools/SVF/issues/1703#issuecomment-2901506766_) on pull request #1703, installing into a dedicated project prefix can often be desirable for clarity and maintenance. To make this behaviour optional, I included user-configurable build parameters like `SVF_INSTALL_INCLUDEDIR` and `SVF_INSTALL_BINDIR` to override the default installation prefixes.
> 
> However, I made a mistake in defining the type of these parameters; they are currently defined as `PATH` types instead of `STRING` types. As these defaulted to their `${CMAKE_INSTALL_<...>}` counterparts (which are always correctly interpreted as **relative to the installation prefix**, this would ensure the final installation destination would be correct (i.e., `<install_prefix>/include/SVF/...`, `<install_prefix>/bin/...`, etc.). However, defining these types as `PATH` types would cause them to be incorrectly resolved into **absolute paths**. Consequently, the current working directory would be internally prepended to these installation prefix parameters, causing the final installation destinations to look like `<pwd>/<user_config_prefix>` (e.g., `$(pwd)/include/SVF`). Consequently, the source/install paths to contain their own prefixes, causing the build configuration to fail.
> 
> This small fix simply uses the `STRING` type for these installation prefix parameters (e.g., `SVF_INSTALL_INCLUDEDIR`) instead, ensuring they are always correctly interpreted as **relative paths** (relative to the overall installation prefix).
> 
> NB: I would still argue defaulting to the installation prefix for public header files being `include/SVF` instead of the bare `include`, and/or properly describing the possible build parameters in the setup guide. Should I add that to the [Wiki page for that](https://github.com/svf-tools/SVF/wiki/Setup-Guide#getting-started)? If so, is my understanding correct that anyone can modify the wiki without requiring explicit permissions?

Thanks for this. It looks to me that `include` is better. If you want `include/svf`, we should also have `include/svf-llvm`. I didn't get what parameters to add in the wiki link?

yuleisui:
> > After my previous pull request #1703 was merged, commit [a745cdc](https://github.com/SVF-tools/SVF/commit/a745cdc88af37f4395566d3475fb23d718874930) changed the default installation prefix for public headers from `include/SVF` to just `include`. As discussed in [my last comment](https://github.com/SVF-tools/SVF/issues/1703#issuecomment-2901506766_) on pull request #1703, installing into a dedicated project prefix can often be desirable for clarity and maintenance. To make this behaviour optional, I included user-configurable build parameters like `SVF_INSTALL_INCLUDEDIR` and `SVF_INSTALL_BINDIR` to override the default installation prefixes.
> > However, I made a mistake in defining the type of these parameters; they are currently defined as `PATH` types instead of `STRING` types. As these defaulted to their `${CMAKE_INSTALL_<...>}` counterparts (which are always correctly interpreted as **relative to the installation prefix**, this would ensure the final installation destination would be correct (i.e., `<install_prefix>/include/SVF/...`, `<install_prefix>/bin/...`, etc.). However, defining these types as `PATH` types would cause them to be incorrectly resolved into **absolute paths**. Consequently, the current working directory would be internally prepended to these installation prefix parameters, causing the final installation destinations to look like `<pwd>/<user_config_prefix>` (e.g., `$(pwd)/include/SVF`). Consequently, the source/install paths to contain their own prefixes, causing the build configuration to fail.
> > This small fix simply uses the `STRING` type for these installation prefix parameters (e.g., `SVF_INSTALL_INCLUDEDIR`) instead, ensuring they are always correctly interpreted as **relative paths** (relative to the overall installation prefix).
> > NB: I would still argue defaulting to the installation prefix for public header files being `include/SVF` instead of the bare `include`, and/or properly describing the possible build parameters in the setup guide. Should I add that to the [Wiki page for that](https://github.com/svf-tools/SVF/wiki/Setup-Guide#getting-started)? If so, is my understanding correct that anyone can modify the wiki without requiring explicit permissions?
> 
> Thanks for this. It looks to me that `include` is better. If you want `include/svf`, we should also have `include/svf-llvm`. I didn't get what parameters to add in the wiki link?

If we want `include/svf`, many header inclusions should be changed, like [this](https://github.com/SVF-tools/SVF/blob/89d302a00f0b766dd3de51d8e2fe742dee971614/svf/lib/AE/Core/AbstractState.cpp#L30) from `#include "AE/Core/AbstractState.h"` to  `#include "SVF/AE/Core/AbstractState.h"` to be correct.

Johanmyst:
> > After my previous pull request #1703 was merged, commit [a745cdc](https://github.com/SVF-tools/SVF/commit/a745cdc88af37f4395566d3475fb23d718874930) changed the default installation prefix for public headers from `include/SVF` to just `include`. As discussed in [my last comment](https://github.com/SVF-tools/SVF/issues/1703#issuecomment-2901506766_) on pull request #1703, installing into a dedicated project prefix can often be desirable for clarity and maintenance. To make this behaviour optional, I included user-configurable build parameters like `SVF_INSTALL_INCLUDEDIR` and `SVF_INSTALL_BINDIR` to override the default installation prefixes.
> > However, I made a mistake in defining the type of these parameters; they are currently defined as `PATH` types instead of `STRING` types. As these defaulted to their `${CMAKE_INSTALL_<...>}` counterparts (which are always correctly interpreted as **relative to the installation prefix**, this would ensure the final installation destination would be correct (i.e., `<install_prefix>/include/SVF/...`, `<install_prefix>/bin/...`, etc.). However, defining these types as `PATH` types would cause them to be incorrectly resolved into **absolute paths**. Consequently, the current working directory would be internally prepended to these installation prefix parameters, causing the final installation destinations to look like `<pwd>/<user_config_prefix>` (e.g., `$(pwd)/include/SVF`). Consequently, the source/install paths to contain their own prefixes, causing the build configuration to fail.
> > This small fix simply uses the `STRING` type for these installation prefix parameters (e.g., `SVF_INSTALL_INCLUDEDIR`) instead, ensuring they are always correctly interpreted as **relative paths** (relative to the overall installation prefix).
> > NB: I would still argue defaulting to the installation prefix for public header files being `include/SVF` instead of the bare `include`, and/or properly describing the possible build parameters in the setup guide. Should I add that to the [Wiki page for that](https://github.com/svf-tools/SVF/wiki/Setup-Guide#getting-started)? If so, is my understanding correct that anyone can modify the wiki without requiring explicit permissions?
> 
> Thanks for this. It looks to me that `include` is better. If you want `include/svf`, we should also have `include/svf-llvm`.

I'm not sure I follow. Currently, public headers are installed to the bare `include` prefix; meaning the install tree looks like this:

```
<install_prefix>
├── bin
│   └── ...
├── include
│   ├── AE
│   │   ├── Core
│   │   │   └── ...
│   │   └── Svfexe
│   │       └── ...
│   ├── CFL
│   │   └── ...
│   ├── DDA
│   │   └── ...
│   ├── FastCluster
│   │   └── ...
│   ├── Graphs
│   │   └── ...
│   ├── MemoryModel
│   │   └── ...
│   ├── MSSA
│   │   └── ...
│   ├── MTA
│   │   └── ...
│   ├── SABER
│   │   └── ...
│   ├── SVF-LLVM
│   │   └── ...
│   ├── SVFIR
│   │   └── ...
│   ├── Util
│   │   └── ...
│   └── WPA
│       └── ...
└── lib
    ├── cmake
    │   └── SVF
    │       ├── Modules
    │       │   └── ...
    │       └── ...
    ├── pkgconfig
    │   └── ...
    └── ...
```

Note that the public headers defined in `<SVF_ROOT>/svf-llvm/include/SVF-LLVM` (the sources for these) are installed to the `SVF-LLVM` directory inside of the `<install_prefix>/include` directory. Installing to somewhere like `<SVF_ROOT>/install` is pretty clean, but say I wanted to install to something like `/usr` or `/usr/local`, then all of these directories (e.g., `MTA`, `MSSA`, etc.) would end up directly in the `/usr/include` or `/usr/local/include`. As these directories already contain numerous other files and directories, this could make it very vague which headers belong to SVF.

To avoid this cluttering, projects often install public headers to a separate prefix: e.g., LLVM installs its headers to `<install_prefix>/include/llvm/...` instead of directly into `<install_prefix>/include`, the same goes for numerous other projects as — from what I know — this is often considered to be the "best practice". While the CMake build tree contains the `install_manifest.txt` file (listing all files installed by the `cmake --install ...` step), if this file is lost, uninstalling SVF would require a user to manually erase directories like `MSSA`, `MemoryModel`, `Graphs`, etc. Since these directories don't necessarily have very clear names (e.g., `Util` or `MemoryModel` is very generic and doesn't suggest they are part of SVF), this might be difficult to properly do. If the `SVF` prefix were used, a user would only have to erase `include/SVF`.

> If we want include/svf, many header inclusions should be changed, like [this](https://github.com/SVF-tools/SVF/blob/89d302a00f0b766dd3de51d8e2fe742dee971614/svf/lib/AE/Core/AbstractState.cpp#L30) from #include "AE/Core/AbstractState.h" to #include "SVF/AE/Core/AbstractState.h" to be correct.

No, someone **using SVF as a library** would have to include SVF's headers like that (i.e., `#include <SVF/AE/Core/AbstractState.h>`), but **in-tree users of SVF** have access to the `include` tree in the current build directory.

```cmake
# Expose headers in build-tree only to in-tree users (not as system headers; in-tree users see warnings)
target_include_directories(
  SvfFlags
  INTERFACE $<BUILD_INTERFACE:${SVF_BINARY_DIR}>
  INTERFACE $<BUILD_INTERFACE:${SVF_BINARY_DIR}/include>
  INTERFACE $<BUILD_INTERFACE:${SVF_BINARY_DIR}/include/SVF>
)
```

The above snippet ensures that, **during the build of SVF** (which includes building SVF itself but also any projects that use SVF in-tree (i.e., by having SVF's cloned into `<project_root>/SVF` and adding `add_subdirectory(SVF)` to the project's CMakeLists.txt definition) or extend SVF (i.e., by adding new subprojects to the SVF source tree), users have access to the build directory's include tree. Similarly, the `SvfCore` and `SvfLLVM` libraries declare their source-tree include directories publicly for any build-time users:

```cmake
### <svf_root>/svf/CMakeLists.txt:
target_include_directories(SvfCore PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_LIST_DIR}/include>)

### <svf_root>/svf-llvm/CMakeLists.txt:
target_include_directories(SvfLLVM PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_LIST_DIR}/include>)
```

This ensures that SVF's sources can continue to use the plain includes like `#include "AE/Core/AbstractState.h"`, and this will not cause any issues nor require any changes to SVF's sources or to how SVF is built.

Also, while ultimately it would probably be better to only support using SVF with `#include <SVF/...>`  for **people using SVF as a library**, I also added this snippet:

```cmake
# Expose installed headers publicly (as system headers to suppress internal warnings to end-users)
target_include_directories(
  SvfFlags
  SYSTEM
  INTERFACE $<INSTALL_INTERFACE:${SVF_INSTALL_INCLUDEDIR}>
  INTERFACE $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>
)
```

These directories are defined on the `INSTALL_INTERFACE`, meaning anyone that uses SVF's CMake package (i.e., `find_package(SVF ...)`) would automatically inherit these directories whenever they link against SVF (i.e., `target_link_libraries(Foo SVF::SvfCore SVF::SvfLLVM)`). This means a minimal CMakeLists.txt file to use SVF could look like this:

```cmake
project(Foo)
add_executable(Bar bar.cpp)
find_package(SVF REQUIRED)
target_link_libraries(Bar SVF::SvfCore SVF::SvfLLVM)
install(TARGETS Bar RUNTIME DESTINATION bin)
```

The directories added to SVF's CMake package's install interface are automatically inherited so only the `target_link_libraries()` call also adds required compiler flags like `-I<install_prefix>` for whatever installation prefix was used to install SVF. Note that because of the `SYSTEM` flag in the `target_include_directories(SvfFlags SYSTEM ...)` call, the `-isystem <install_prefix>` compiler flag will be used instead, which ensures end-users don't run into internal warnings from the SVF headers (e.g., if a variable would be unused in one of the SVF header files but the user compiles with flags to show warnings for this).

Note that I also added this include directory for backwards compatibility & maintaining current expected behaviours: `INTERFACE $<INSTALL_INTERFACE:${SVF_INSTALL_INCLUDEDIR}>`. This means that if this parameter was set to `include/SVF`, the build flag `-isystem <install_prefix>/include/SVF` will also be added for end users. This is kind of a cheeky trick to ensure end-users can actually also include SVF headers with `#include <AE/Core/AbstractState.h>`, though ideally, end-users would be using `#include <SVF/AE/Core/AbstractState.h>`.

This all ensures that I can build & install SVF like this on a clean Ubuntu (24.0.4 (24.0.4.2 LTS)) system:

```shell
sudo apt update
sudo apt install llvm-16 llvm-16-tools llvm-16-dev \
    clang-16 libclang-16-dev libclang-rt-16-dev \
    libc++-16-dev libc++abi-16-dev \
    lld-16 liblld-16 liblld-16-dev
sudo apt install z3 libz3-dev

git clone https://github.com/SVF-tools/SVF.git
cd SVF
cmake -S -B Release-build -G "Ninja" \
      -DBUILD_SHARED_LIBS=ON \
      -DCMAKE_BUILD_TYPE=Release \
      -DCMAKE_INSTALL_PREFIX=/usr/local \
      -DSVF_INSTALL_INCLUDEDIR=include/SVF \
      -DSVF_ENABLE_ASSERTIONS=ON \
      -DSVF_EXPORT_DYNAMIC=ON \
      -DSVF_USE_LLD=ON
cmake --build Release-build --verbose
sudo cmake --install Release-build --verbose
```

As the installation prefix `/usr/local` is already part of most default search-paths, this after the `cmake --install Release-build` step, SVF can be found and used as a library in any other CMake project with `find_package(SVF)` and `target_link_libraries(MyProject SVF::SvfCore SVF::SvfLLVM)`. Hopefully this clarifies why I structured the defaults the way I did!

> I didn't get what parameters to add in the wiki link?

The current `build.sh` script only checks `$LLVM_DIR` and `$LLVMHome`, and if neither of them are set, it will always download LLVM; the same goes for Z3. However, if someone installed LLVM from the default Ubuntu repositories, or manually built & installed it elsewhere, this is unexpected and undesirable as the script's name (`build.sh`) doesn't suggest it would do this. Moreover, while it supports some configuration options, it always configures & builds SVF like this:

```shell
cmake -D CMAKE_BUILD_TYPE:STRING="${BUILD_TYPE}"   \
    -DSVF_ENABLE_ASSERTIONS:BOOL=true              \
    -DSVF_SANITIZE="${SVF_SANITIZER}"              \
    -DBUILD_SHARED_LIBS=${BUILD_DYN_LIB}            \
    -S "${SVFHOME}" -B "${BUILD_DIR}"
cmake --build "${BUILD_DIR}" -j ${jobs}
```

Not only does this not actually install SVF (meaning the SVF CMake package cannot be conveniently/properly used), but it also doesn't easily let the user configure any of SVF's build parameters/options. Hence, I would suggest moving the part that pulls LLVM & Z3 into dedicated `get_llvm.sh` and `get_z3.sh` scripts. The setup guide Wiki page could then just mention thee scripts in a "SVF Dependencies" section, and document the available build options in the main "Building & Installing SVF" section. Another "Using SVF" section could then show how the SVF package can be found with `find_package()` and linked directly. I was offering to add this writing to the setup Wiki page, if you'd like!

yuleisui:
> > > After my previous pull request #1703 was merged, commit [a745cdc](https://github.com/SVF-tools/SVF/commit/a745cdc88af37f4395566d3475fb23d718874930) changed the default installation prefix for public headers from `include/SVF` to just `include`. As discussed in [my last comment](https://github.com/SVF-tools/SVF/issues/1703#issuecomment-2901506766_) on pull request #1703, installing into a dedicated project prefix can often be desirable for clarity and maintenance. To make this behaviour optional, I included user-configurable build parameters like `SVF_INSTALL_INCLUDEDIR` and `SVF_INSTALL_BINDIR` to override the default installation prefixes.
> > > However, I made a mistake in defining the type of these parameters; they are currently defined as `PATH` types instead of `STRING` types. As these defaulted to their `${CMAKE_INSTALL_<...>}` counterparts (which are always correctly interpreted as **relative to the installation prefix**, this would ensure the final installation destination would be correct (i.e., `<install_prefix>/include/SVF/...`, `<install_prefix>/bin/...`, etc.). However, defining these types as `PATH` types would cause them to be incorrectly resolved into **absolute paths**. Consequently, the current working directory would be internally prepended to these installation prefix parameters, causing the final installation destinations to look like `<pwd>/<user_config_prefix>` (e.g., `$(pwd)/include/SVF`). Consequently, the source/install paths to contain their own prefixes, causing the build configuration to fail.
> > > This small fix simply uses the `STRING` type for these installation prefix parameters (e.g., `SVF_INSTALL_INCLUDEDIR`) instead, ensuring they are always correctly interpreted as **relative paths** (relative to the overall installation prefix).
> > > NB: I would still argue defaulting to the installation prefix for public header files being `include/SVF` instead of the bare `include`, and/or properly describing the possible build parameters in the setup guide. Should I add that to the [Wiki page for that](https://github.com/svf-tools/SVF/wiki/Setup-Guide#getting-started)? If so, is my understanding correct that anyone can modify the wiki without requiring explicit permissions?
> > 
> > 
> > Thanks for this. It looks to me that `include` is better. If you want `include/svf`, we should also have `include/svf-llvm`.
> 
> I'm not sure I follow. Currently, public headers are installed to the bare `include` prefix; meaning the install tree looks like this:
> 
> ```
> <install_prefix>
> ├── bin
> │   └── ...
> ├── include
> │   ├── AE
> │   │   ├── Core
> │   │   │   └── ...
> │   │   └── Svfexe
> │   │       └── ...
> │   ├── CFL
> │   │   └── ...
> │   ├── DDA
> │   │   └── ...
> │   ├── FastCluster
> │   │   └── ...
> │   ├── Graphs
> │   │   └── ...
> │   ├── MemoryModel
> │   │   └── ...
> │   ├── MSSA
> │   │   └── ...
> │   ├── MTA
> │   │   └── ...
> │   ├── SABER
> │   │   └── ...
> │   ├── SVF-LLVM
> │   │   └── ...
> │   ├── SVFIR
> │   │   └── ...
> │   ├── Util
> │   │   └── ...
> │   └── WPA
> │       └── ...
> └── lib
>     ├── cmake
>     │   └── SVF
>     │       ├── Modules
>     │       │   └── ...
>     │       └── ...
>     ├── pkgconfig
>     │   └── ...
>     └── ...
> ```
> 
> Note that the public headers defined in `<SVF_ROOT>/svf-llvm/include/SVF-LLVM` (the sources for these) are installed to the `SVF-LLVM` directory inside of the `<install_prefix>/include` directory. Installing to somewhere like `<SVF_ROOT>/install` is pretty clean, but say I wanted to install to something like `/usr` or `/usr/local`, then all of these directories (e.g., `MTA`, `MSSA`, etc.) would end up directly in the `/usr/include` or `/usr/local/include`. As these directories already contain numerous other files and directories, this could make it very vague which headers belong to SVF.
> 
> To avoid this cluttering, projects often install public headers to a separate prefix: e.g., LLVM installs its headers to `<install_prefix>/include/llvm/...` instead of directly into `<install_prefix>/include`, the same goes for numerous other projects as — from what I know — this is often considered to be the "best practice". While the CMake build tree contains the `install_manifest.txt` file (listing all files installed by the `cmake --install ...` step), if this file is lost, uninstalling SVF would require a user to manually erase directories like `MSSA`, `MemoryModel`, `Graphs`, etc. Since these directories don't necessarily have very clear names (e.g., `Util` or `MemoryModel` is very generic and doesn't suggest they are part of SVF), this might be difficult to properly do. If the `SVF` prefix were used, a user would only have to erase `include/SVF`.
> 
> > If we want include/svf, many header inclusions should be changed, like [this](https://github.com/SVF-tools/SVF/blob/89d302a00f0b766dd3de51d8e2fe742dee971614/svf/lib/AE/Core/AbstractState.cpp#L30) from #include "AE/Core/AbstractState.h" to #include "SVF/AE/Core/AbstractState.h" to be correct.
> 
> No, someone **using SVF as a library** would have to include SVF's headers like that (i.e., `#include <SVF/AE/Core/AbstractState.h>`), but **in-tree users of SVF** have access to the `include` tree in the current build directory.
> 
> ```cmake
> # Expose headers in build-tree only to in-tree users (not as system headers; in-tree users see warnings)
> target_include_directories(
>   SvfFlags
>   INTERFACE $<BUILD_INTERFACE:${SVF_BINARY_DIR}>
>   INTERFACE $<BUILD_INTERFACE:${SVF_BINARY_DIR}/include>
>   INTERFACE $<BUILD_INTERFACE:${SVF_BINARY_DIR}/include/SVF>
> )
> ```
> 
> The above snippet ensures that, **during the build of SVF** (which includes building SVF itself but also any projects that use SVF in-tree (i.e., by having SVF's cloned into `<project_root>/SVF` and adding `add_subdirectory(SVF)` to the project's CMakeLists.txt definition) or extend SVF (i.e., by adding new subprojects to the SVF source tree), users have access to the build directory's include tree. Similarly, the `SvfCore` and `SvfLLVM` libraries declare their source-tree include directories publicly for any build-time users:
> 
> ```cmake
> ### <svf_root>/svf/CMakeLists.txt:
> target_include_directories(SvfCore PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_LIST_DIR}/include>)
> 
> ### <svf_root>/svf-llvm/CMakeLists.txt:
> target_include_directories(SvfLLVM PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_LIST_DIR}/include>)
> ```
> 
> This ensures that SVF's sources can continue to use the plain includes like `#include "AE/Core/AbstractState.h"`, and this will not cause any issues nor require any changes to SVF's sources or to how SVF is built.
> 
> Also, while ultimately it would probably be better to only support using SVF with `#include <SVF/...>` for **people using SVF as a library**, I also added this snippet:
> 
> ```cmake
> # Expose installed headers publicly (as system headers to suppress internal warnings to end-users)
> target_include_directories(
>   SvfFlags
>   SYSTEM
>   INTERFACE $<INSTALL_INTERFACE:${SVF_INSTALL_INCLUDEDIR}>
>   INTERFACE $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>
> )
> ```
> 
> These directories are defined on the `INSTALL_INTERFACE`, meaning anyone that uses SVF's CMake package (i.e., `find_package(SVF ...)`) would automatically inherit these directories whenever they link against SVF (i.e., `target_link_libraries(Foo SVF::SvfCore SVF::SvfLLVM)`). This means a minimal CMakeLists.txt file to use SVF could look like this:
> 
> ```cmake
> project(Foo)
> add_executable(Bar bar.cpp)
> find_package(SVF REQUIRED)
> target_link_libraries(Bar SVF::SvfCore SVF::SvfLLVM)
> install(TARGETS Bar RUNTIME DESTINATION bin)
> ```
> 
> The directories added to SVF's CMake package's install interface are automatically inherited so only the `target_link_libraries()` call also adds required compiler flags like `-I<install_prefix>` for whatever installation prefix was used to install SVF. Note that because of the `SYSTEM` flag in the `target_include_directories(SvfFlags SYSTEM ...)` call, the `-isystem <install_prefix>` compiler flag will be used instead, which ensures end-users don't run into internal warnings from the SVF headers (e.g., if a variable would be unused in one of the SVF header files but the user compiles with flags to show warnings for this).
> 
> Note that I also added this include directory for backwards compatibility & maintaining current expected behaviours: `INTERFACE $<INSTALL_INTERFACE:${SVF_INSTALL_INCLUDEDIR}>`. This means that if this parameter was set to `include/SVF`, the build flag `-isystem <install_prefix>/include/SVF` will also be added for end users. This is kind of a cheeky trick to ensure end-users can actually also include SVF headers with `#include <AE/Core/AbstractState.h>`, though ideally, end-users would be using `#include <SVF/AE/Core/AbstractState.h>`.
> 
> This all ensures that I can build & install SVF like this on a clean Ubuntu (24.0.4 (24.0.4.2 LTS)) system:
> 
> ```shell
> sudo apt update
> sudo apt install llvm-16 llvm-16-tools llvm-16-dev \
>     clang-16 libclang-16-dev libclang-rt-16-dev \
>     libc++-16-dev libc++abi-16-dev \
>     lld-16 liblld-16 liblld-16-dev
> sudo apt install z3 libz3-dev
> 
> git clone https://github.com/SVF-tools/SVF.git
> cd SVF
> cmake -S -B Release-build -G "Ninja" \
>       -DBUILD_SHARED_LIBS=ON \
>       -DCMAKE_BUILD_TYPE=Release \
>       -DCMAKE_INSTALL_PREFIX=/usr/local \
>       -DSVF_INSTALL_INCLUDEDIR=include/SVF \
>       -DSVF_ENABLE_ASSERTIONS=ON \
>       -DSVF_EXPORT_DYNAMIC=ON \
>       -DSVF_USE_LLD=ON
> cmake --build Release-build --verbose
> sudo cmake --install Release-build --verbose
> ```
> 
> As the installation prefix `/usr/local` is already part of most default search-paths, this after the `cmake --install Release-build` step, SVF can be found and used as a library in any other CMake project with `find_package(SVF)` and `target_link_libraries(MyProject SVF::SvfCore SVF::SvfLLVM)`. Hopefully this clarifies why I structured the defaults the way I did!
> 
> > I didn't get what parameters to add in the wiki link?
> 
> The current `build.sh` script only checks `$LLVM_DIR` and `$LLVMHome`, and if neither of them are set, it will always download LLVM; the same goes for Z3. However, if someone installed LLVM from the default Ubuntu repositories, or manually built & installed it elsewhere, this is unexpected and undesirable as the script's name (`build.sh`) doesn't suggest it would do this. Moreover, while it supports some configuration options, it always configures & builds SVF like this:
> 
> ```shell
> cmake -D CMAKE_BUILD_TYPE:STRING="${BUILD_TYPE}"   \
>     -DSVF_ENABLE_ASSERTIONS:BOOL=true              \
>     -DSVF_SANITIZE="${SVF_SANITIZER}"              \
>     -DBUILD_SHARED_LIBS=${BUILD_DYN_LIB}            \
>     -S "${SVFHOME}" -B "${BUILD_DIR}"
> cmake --build "${BUILD_DIR}" -j ${jobs}
> ```
> 
> Not only does this not actually install SVF (meaning the SVF CMake package cannot be conveniently/properly used), but it also doesn't easily let the user configure any of SVF's build parameters/options. Hence, I would suggest moving the part that pulls LLVM & Z3 into dedicated `get_llvm.sh` and `get_z3.sh` scripts. The setup guide Wiki page could then just mention thee scripts in a "SVF Dependencies" section, and document the available build options in the main "Building & Installing SVF" section. Another "Using SVF" section could then show how the SVF package can be found with `find_package()` and linked directly. I was offering to add this writing to the setup Wiki page, if you'd like!

Thanks for the detailed explanation!

I agree that we should adopt `include/SVF` to allow make install to be cleaner. Please submit a patch for this.

It is also quite good to have `get_llvm.sh` and `get_z3.sh` scripts as per your suggestions. I would say that we still keep a `build.sh` file to make it one-stop shop for all users. Please go ahead and make it a separate patch if you want. You can  modify or edit the wiki or edit that in your forked repo for us to copy it back.

yuleisui:
I will merge this pull request first and we could continue the discussion or open another pull request for those changes.

PR #1748, opened 2025-08-25
Fix abort in ObjTypeInference::inferObjType

Khosro-Moeini:
# Problem
I encountered the abort message in this [line](https://github.com/SVF-tools/SVF/blob/12a270f04a1890d337f418a58b208f2a965a4cae/svf-llvm/lib/ObjTypeInference.cpp#L167) when using a valid IR.

In my case, the source of the problem was that `var` was the first argument of an `iconv` call.

# Fix
Return `res` if `var` is the first argument of an `iconv`call.


yuleisui:
Thanks for the fix.

PR #1785, opened 2026-01-18
Refactor of AE before adding Sparse Features

bjjwwang:
Summary of Changes                                         
                                                             
  Phase 1: Created IAbstractState.h interface                
                                                             
  - File: SVF/svf/include/AE/Core/IAbstractState.h           
  - Created new abstract interface with pure virtual methods 
  for state operations                                       
                                                             
  Phase 2: Modified AbstractState to implement interface     
                                                             
  - Files: AbstractState.h, AbstractState.cpp                
  - Added override specifiers to all interface methods       
  - Added clone() and getStateName() methods                 
  - Changed widening/narrowing to return                     
  std::unique_ptr<IAbstractState>                            
  - Added wideningDense/narrowingDense for internal          
  backward-compatible usage                                  
  - Updated joinWith/meetWith to accept IAbstractState&      
  - Fixed missing include guards in AbstractValue.h          
                                                             
  Phase 3: Decoupled AEDetector classes                      
                                                             
  - Files: AEDetector.h, AEDetector.cpp                      
  - Changed all AbstractState& parameters to IAbstractState& 
  - Added override specifiers to virtual methods             
  - Added dynamic_cast to dense state where needed for       
  AbsExtAPI compatibility                                    
                                                             
  Phase 4: Decoupled AbstractInterpretation engine           
                                                             
  - Files: AbstractInterpretation.h,                         
  AbstractInterpretation.cpp                                 
  - getAbsStateFromTrace() now returns IAbstractState&       
  - Added getDenseAbsStateFromTrace() for internal use       
  - Added createState() factory method (placeholder for      
  sparse state)                                              
  - Updated internal usages to use dense accessors           
                                                             
  Phase 5: Added use-sparse CLI options  (TODO)                               
                                                             
  - Files: Options.h, Options.cpp                            
  - Added -use-sparse CLI option (defaults to false)         
  - Factory method will assert if sparse state is requested  
  (not yet implemented)

### DIFF
--- a/svf/include/AE/Svfexe/AEDetector.h
+++ b/svf/include/AE/Svfexe/AEDetector.h
@@ -152,9 +152,6 @@ class BufOverflowDetector : public AEDetector
 {
     friend class AbstractInterpretation;
 
-private:
-    Map<const GepObjVar*, IntervalValue> gepObjOffsetFromBase; ///< Maps GEP objects to their offsets from the base.
-
 public:
     /**
      * @brief Constructor initializes the detector kind to BUF_OVERFLOW and sets up external API buffer overflow rules.
@@ -354,6 +351,7 @@ private:
     bool detectStrcpy(AbstractState& as, const CallICFGNode *call);
 
 private:
+    Map<const GepObjVar*, IntervalValue> gepObjOffsetFromBase; ///< Maps GEP objects to their offsets from their base.
     Map<std::string, std::vector<std::pair<u32_t, u32_t>>> extAPIBufOverflowCheckRules; ///< Rules for checking buffer overflows in external APIs.
     Set<std::string> bugLoc; ///< Set of locations where bugs have been reported.
     SVFBugReport recoder; ///< Recorder for abstract execution bugs.
--- a/svf/include/AE/Svfexe/AbstractInterpretation.h
+++ b/svf/include/AE/Svfexe/AbstractInterpretation.h
@@ -178,8 +178,14 @@ public:
         }
     }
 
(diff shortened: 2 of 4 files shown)


yuleisui:
Phase 1:
If the AbstractState interface will be used extensively, including in other repositories, it would be better to keep the name AbstractState for the interface and rename the current AbstractState to AbstractStateImpl.

Phase 2:
Why do we need std::unique_ptr wrapping for widening and narrowing?
Do we still need backward compatibility for wideningDense and narrowingDense? We should aim to simplify the current implementations as much as possible for AE.

Phases 3 and 4:
Could we try to simplify the code by removing parts that are no longer used or are unnecessarily complex?

svf/lib/AE/Svfexe/AbsExtAPI.cpp
```diff
@@ -670,7 +670,7 @@ void AbsExtAPI::handleMemcpy(AbstractState& as, const SVF::SVFVar *dst, const SV
     }
 }
 
-void AbsExtAPI::handleMemset(AbstractState& as, const SVF::SVFVar *dst, IntervalValue elem, IntervalValue len)
+void AbsExtAPI::handleMemset(AbstractStateImpl& as, const SVF::SVFVar *dst, IntervalValue elem, IntervalValue len)
```
yuleisui: It looks that we can't use AbstractState anymore? Every place has to be replaced with AbstractStateImpl?

If that is the case, we will have to use AbstractStateInterface and keep AbstractState as the implementation.

bjjwwang:
ok I close it.

PR #1804, opened 2026-03-19
Set correct ICFGNode for ValVar subclasses during SVFIR construction

bjjwwang:
### Problem

Several `ValVar` subclasses were constructed with `nullptr` as their `ICFGNode` because their underlying LLVM values are not `Instruction`s (e.g., `Argument`, `Function`, `GlobalVariable`). The existing code in `SVFIRBuilder` tries `dyn_cast<Instruction>(llvmValue)` to obtain the ICFGNode, which silently produces `nullptr` for all non-instruction values.

This makes it impossible for downstream analyses to determine where a ValVar is "defined" in the ICFG, which is required for sparse analysis optimizations (e.g., semi-sparse abstract interpretation that looks up ValVar values at their definition sites).

### Background: Function Call Parameter Passing in SVF

To understand which ValVars need fixing, it helps to see the full picture of how SVF models function calls:

**Internal function call** `int ret = foo(x)`:

```
CALLER side                                    CALLEE side

                  CallPE (src → dst)
  x (ValVar)    ──────────────────────────►   a (ArgValVar)
  "actual arg"                                 "formal param"
  ICFGNode: instruction that defined x         ICFGNode: FunEntryICFGNode ← FIXED

                                               ... foo body executes ...

                  RetPE (src → dst)
  ret (ValVar)  ◄──────────────────────────   RetValPN
  "actual ret"                                 "formal ret"
  ICFGNode: CallICFGNode (call instruction)    ICFGNode: FunExitICFGNode ← FIXED
```

- `CallPE` copies the actual argument to the formal parameter at `FunEntryICFGNode`.
- `RetPE` copies the formal return value to the actual return at `RetICFGNode`.
- The actual arg and actual ret are regular ValVars from instructions — they already have correct ICFGNodes.
- The **formal param** (`ArgValVar`) and **formal ret** (`RetValPN`) previously had `nullptr` because `Argument` and `Function*` are not `Instruction`.

**External function call** `int ret = ext_func(x)`:

External functions (declarations without bodies) have **no `FunEntryICFGNode` or `FunExitICFGNode`** in the ICFG. More importantly, `handleDirectCall` (which creates `CallPE`/`RetPE`) is only invoked for non-external calls. Therefore:

- **No `CallPE`** is created — the formal param `ArgValVar` is never written to.
- **No `RetPE`** is created — the formal ret `RetValPN` is never read from.
- `AbsExtAPI` handles external calls entirely at the `CallICFGNode`, using only caller-side variables (`callNode->getArgument(i)` for actual args, `callNode->getRetICFGNode()->getActualRet()` for actual ret).

Since external functions' `ArgValVar`/`RetValPN`/`VarArgValPN` are never assigned or queried by any statement, their `nullptr` ICFGNode is safe and correct.

### Changes

**`SVFIRBuilder.cpp`** — set correct ICFGNodes for 5 ValVar types:

| ValVar type | Before | After |
|-------------|--------|-------|
| `ArgValVar` (formal param) | `nullptr` | `FunEntryICFGNode` (defined functions) / `nullptr` (external declarations) |
| `RetValPN` (formal return) | `nullptr` | `FunExitICFGNode` (defined functions) / `nullptr` (external declarations) |
| `VarArgValPN` (variadic param) | `nullptr` | `FunEntryICFGNode` (defined functions) / `nullptr` (external declarations) |
| `GlobalValVar` | `nullptr` | `GlobalICFGNode` |
| `GepValVar` from `GlobalVariable` | `nullptr` | `GlobalICFGNode` |

The previous code for `RetValPN` and `VarArgValPN` had dead logic: it attempted `dyn_cast<Instruction>(llvmValue)` on a `Function*`, which always fails. This dead code is replaced with the direct semantic assignment.

**`SVFVariables.h` / `SVFVariables.cpp`** — add constructor assertions:

- `GlobalValVar`: `assert(icn && "GlobalValVar must have a valid ICFGNode")`
- `GepValVar`: `assert(node && "GepValVar must have a valid ICFGNode")`

These are the only two subclasses where ICFGNode is guaranteed non-null after the fix. Other subclasses (`ArgValVar`, `RetValPN`, `VarArgValPN`) legitimately have `nullptr` for external function declarations.

### DIFF
--- a/svf-llvm/lib/SVFIRBuilder.cpp
+++ b/svf-llvm/lib/SVFIRBuilder.cpp
@@ -509,7 +509,7 @@ void SVFIRBuilder::initialiseValVars()
             pag->addGlobalValNode(iter->second, pag->getICFG()->getGlobalICFGNode(),
                                   llvmModuleSet()->getSVFType(llvmValue->getType()));
         }
-        else if (SVFUtil::isa<ConstantData, MetadataAsValue, BlockAddress>(llvmValue))
+        else if (SVFUtil::isa<ConstantData, ConstantExpr, MetadataAsValue, BlockAddress>(llvmValue))
         {
             pag->addConstantDataValNode(iter->second, icfgNode, llvmModuleSet()->getSVFType(llvmValue->getType()));
         }
@@ -519,7 +519,6 @@ void SVFIRBuilder::initialiseValVars()
         }
         else
         {
-            // Add value node to PAG
             pag->addValNode(iter->second, llvmModuleSet()->getSVFType(llvmValue->getType()), icfgNode);
         }
         llvmModuleSet()->addToSVFVar2LLVMValueMap(llvmValue,
--- a/svf-llvm/lib/SVFIRBuilder.cpp
+++ b/svf-llvm/lib/SVFIRBuilder.cpp
@@ -41,8 +41,6 @@
 #include "Graphs/CallGraph.h"
 #include "Util/Options.h"
 #include "Util/SVFUtil.h"
-#include <llvm/IR/InlineAsm.h>
-#include <llvm/IR/Constants.h>
 
 using namespace std;
 using namespace SVF;
@@ -515,9 +513,9 @@ void SVFIRBuilder::initialiseValVars()
         {
             pag->addBasicBlockValNode(iter->second, llvmModuleSet()->getSVFType(llvmValue->getType()));
         }
-        else if (SVFUtil::isa<llvm::InlineAsm>(llvmValue) ||
-                 SVFUtil::isa<llvm::DSOLocalEquivalent>(llvmValue) ||
-                 SVFUtil::isa<llvm::NoCFIValue>(llvmValue))
+        else if (SVFUtil::isa<InlineAsm>(llvmValue) ||
+                 SVFUtil::isa<DSOLocalEquivalent>(llvmValue) ||
+                 SVFUtil::isa<NoCFIValue>(llvmValue))
         {
             pag->addAsmPCValNode(iter->second, llvmModuleSet()->getSVFType(llvmValue->getType()));
         }
--- a/svf-llvm/lib/SVFIRBuilder.cpp
+++ b/svf-llvm/lib/SVFIRBuilder.cpp
@@ -428,7 +428,7 @@ void SVFIRBuilder::initialiseBaseObjVars()
             NodeID id = llvmModuleSet()->getObjectNode(iter->first);
             pag->addGlobalObjNode(iter->second, pag->getObjTypeInfo(id), icfgNode);
         }
-        else if (SVFUtil::isa<ConstantData, ConstantExpr, MetadataAsValue, BlockAddress>(llvmValue))
+        else if (SVFUtil::isa<ConstantData, MetadataAsValue, BlockAddress>(llvmValue))
         {
             NodeID id = llvmModuleSet()->getObjectNode(iter->first);
             pag->addConstantDataObjNode(iter->second, pag->getObjTypeInfo(id), icfgNode);
@@ -509,7 +509,7 @@ void SVFIRBuilder::initialiseValVars()
             pag->addGlobalValNode(iter->second, pag->getICFG()->getGlobalICFGNode(),
                                   llvmModuleSet()->getSVFType(llvmValue->getType()));
         }
-        else if (SVFUtil::isa<ConstantData, ConstantExpr, MetadataAsValue, BlockAddress>(llvmValue))
+        else if (SVFUtil::isa<ConstantData, MetadataAsValue, BlockAddress>(llvmValue))
         {
             pag->addConstantDataValNode(iter->second, icfgNode, llvmModuleSet()->getSVFType(llvmValue->getType()));
         }
@@ -517,13 +517,10 @@ void SVFIRBuilder::initialiseValVars()
         {
             pag->addConstantAggValNode(iter->second, icfgNode, llvmModuleSet()->getSVFType(llvmValue->getType()));
         }
-        else if (icfgNode)
-        {
-            pag->addValNode(iter->second, llvmModuleSet()->getSVFType(llvmValue->getType()), icfgNode);
-        }
         else
         {
-            pag->addDummyValNode(iter->second, nullptr);
+            // Add value node to PAG
+            pag->addValNode(iter->second, llvmModuleSet()->getSVFType(llvmValue->getType()), icfgNode);
         }
         llvmModuleSet()->addToSVFVar2LLVMValueMap(llvmValue,
                 pag->getGNode(iter->second));
--- a/svf-llvm/lib/SVFIRBuilder.cpp
+++ b/svf-llvm/lib/SVFIRBuilder.cpp
@@ -41,6 +41,8 @@
 #include "Graphs/CallGraph.h"
 #include "Util/Options.h"
 #include "Util/SVFUtil.h"
+#include <llvm/IR/InlineAsm.h>
+#include <llvm/IR/Constants.h>
 
 using namespace std;
 using namespace SVF;
--- a/svf-llvm/lib/SVFIRBuilder.cpp
+++ b/svf-llvm/lib/SVFIRBuilder.cpp
@@ -481,6 +481,9 @@ void SVFIRBuilder::initialiseValVars()
         }
         else if (auto argval = SVFUtil::dyn_cast<Argument>(llvmValue))
         {
+            // Argument is not an Instruction so icfgNode above is nullptr.
+            // Formal params are defined at FunEntryICFGNode (where CallPE copies actual args).
+            // External (declaration-only) functions have no entry node, so keep nullptr.
             const FunObjVar* funObj = llvmModuleSet()->getFunObjVar(argval->getParent());
             const ICFGNode* entryNode = funObj->isDeclaration() ? nullptr : pag->getICFG()->getFunEntryICFGNode(funObj);
             pag->addArgValNode(
@@ -503,6 +506,8 @@ void SVFIRBuilder::initialiseValVars()
         }
         else if (SVFUtil::isa<GlobalValue>(llvmValue))
         {
+            // GlobalValue is not an Instruction so icfgNode above is nullptr.
+            // Global variables are defined at the global ICFG node.
             pag->addGlobalValNode(iter->second, pag->getICFG()->getGlobalICFGNode(),
                                   llvmModuleSet()->getSVFType(llvmValue->getType()));
         }
@@ -546,6 +551,9 @@ void SVFIRBuilder::initialiseNodes()
             ++iter)
     {
         const Value* llvmValue = iter->first;
+        // retSyms keys are Function*, not Instruction, so dyn_cast<Instruction> always fails.
+        // RetValPN represents the callee's return value, defined at FunExitICFGNode.
+        // External functions have no exit node, so keep nullptr.
         const FunObjVar* funObjVar = llvmModuleSet()->getFunObjVar(SVFUtil::cast<Function>(llvmValue));
         const ICFGNode* icfgNode = funObjVar->isDeclaration() ? nullptr : pag->getICFG()->getFunExitICFGNode(funObjVar);
         DBOUT(DPAGBuild, outs() << "add ret node " << iter->second << "\n");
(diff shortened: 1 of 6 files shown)


svf-llvm/lib/SVFIRBuilder.cpp
```diff
@@ -477,14 +477,18 @@ void SVFIRBuilder::initialiseValVars()
         // Check if the value is a function and get its call graph node
         if (const Function* func = SVFUtil::dyn_cast<Function>(llvmValue))
         {
-            // add value node representing the function
             pag->addFunValNode(iter->second, icfgNode, llvmModuleSet()->getFunObjVar(func), llvmModuleSet()->getSVFType(llvmValue->getType()));
         }
         else if (auto argval = SVFUtil::dyn_cast<Argument>(llvmValue))
         {
+            // Argument is not an Instruction so icfgNode above is nullptr.
```
yuleisui: Could you remove this comment? It does not look quite right to me

svf-llvm/lib/SVFIRBuilder.cpp
```diff
@@ -502,7 +506,9 @@ void SVFIRBuilder::initialiseValVars()
         }
         else if (SVFUtil::isa<GlobalValue>(llvmValue))
         {
-            pag->addGlobalValNode(iter->second, icfgNode,
+            // GlobalValue is not an Instruction so icfgNode above is nullptr.
```
yuleisui: Same as above. Please remove this line.

svf-llvm/lib/SVFIRBuilder.cpp
bjjwwang: OK

svf/include/SVFIR/SVFVariables.h
```diff
@@ -1357,6 +1357,7 @@ class GlobalValVar : public ValVar
     GlobalValVar(NodeID i, const ICFGNode* icn, const SVFType* svfType)
         : ValVar(i, svfType, icn, GlobalValNode)
     {
+        assert(icn && "GlobalValVar must have a valid ICFGNode");
```
yuleisui: Could we add more assertions for the subclasses of ValVar?

svf-llvm/lib/SVFIRBuilder.cpp
```diff
@@ -481,9 +481,6 @@ void SVFIRBuilder::initialiseValVars()
         }
         else if (auto argval = SVFUtil::dyn_cast<Argument>(llvmValue))
         {
-            // Argument is not an Instruction so icfgNode above is nullptr.
-            // Formal params are defined at FunEntryICFGNode (where CallPE copies actual args).
-            // External (declaration-only) functions have no entry node, so keep nullptr.
```
yuleisui: Could only remove this single line:             // Argument is not an Instruction so icfgNode above is nullptr.

The below two lines are good

svf-llvm/lib/SVFIRBuilder.cpp
```diff
@@ -506,8 +503,6 @@ void SVFIRBuilder::initialiseValVars()
         }
         else if (SVFUtil::isa<GlobalValue>(llvmValue))
         {
-            // GlobalValue is not an Instruction so icfgNode above is nullptr.
-            // Global variables are defined at the global ICFG node.
```
yuleisui: This second line is good

svf-llvm/lib/SVFIRBuilder.cpp
```diff
@@ -551,9 +546,6 @@ void SVFIRBuilder::initialiseNodes()
             ++iter)
     {
         const Value* llvmValue = iter->first;
-        // retSyms keys are Function*, not Instruction, so dyn_cast<Instruction> always fails.
-        // RetValPN represents the callee's return value, defined at FunExitICFGNode.
-        // External functions have no exit node, so keep nullptr.
```
yuleisui: These are ok

svf-llvm/lib/SVFIRBuilder.cpp
```diff
@@ -569,9 +561,6 @@ void SVFIRBuilder::initialiseNodes()
             iter != llvmModuleSet()->varargSyms().end(); ++iter)
     {
         const Value* llvmValue = iter->first;
-        // varargSyms keys are Function*, not Instruction.
-        // Variadic arguments are received at the function entry point.
-        // External functions have no entry node, so keep nullptr.
```
yuleisui: Same for these comments

svf-llvm/lib/SVFIRBuilder.cpp
```diff
@@ -1692,7 +1681,6 @@ NodeID SVFIRBuilder::getGepValVar(const Value* val, const AccessPath& ap, const
         }
         else if (SVFUtil::isa<GlobalVariable>(curVal))
         {
-            // GEP on a global variable: the resulting GepValVar belongs to the global ICFG node.
```
yuleisui: It is good for this line.

svf/include/SVFIR/SVFVariables.h
yuleisui: Could we add an assertion in ValVar's constructor and remove all subclasses? You could specify which scenarios the assertion always holds?

svf/lib/SVFIR/SVFVariables.cpp
```diff
@@ -82,6 +82,39 @@ void SVFVar::dump() const
...
+    }
+    else if (SVFUtil::isa<GepValVar>(this))
+    {
+        assert(node && "GepValVar must have a valid ICFGNode");
+    }
+    else if (SVFUtil::isa<ArgValVar>(this) ||
+             SVFUtil::isa<RetValPN>(this) ||
+             SVFUtil::isa<VarArgValPN>(this))
+    {
+        // Conditional assert (isDeclaration || icn) is in each subclass constructor.
```
yuleisui, line 100: could promote the asserts from subclasses here and remove the asserts in subclasses

svf/lib/SVFIR/SVFVariables.cpp
```diff
@@ -82,6 +82,39 @@ void SVFVar::dump() const
...
+    {
+        // Constants, function pointers, and dummy nodes don't require an ICFGNode.
+    }
+    else if (ty == ValNode)
+    {
+        // Base ValVar covers values without a dedicated subclass.
+        // Some Instructions are excluded from the ICFG (e.g., llvm.dbg.declare)
+        // and legitimately have nullptr ICFGNode.
+        // TODO: reclassify these or filter them from valSyms to enable assertion.
+    }
```
yuleisui, line 119: we need to have an else branch and assert false.

svf/lib/SVFIR/SVFVariables.cpp
```diff
@@ -82,6 +82,39 @@ void SVFVar::dump() const
...
+    else if (SVFUtil::isa<ConstDataValVar>(this) ||
+             SVFUtil::isa<ConstAggValVar>(this) ||
+             SVFUtil::isa<FunValVar>(this) ||
+             SVFUtil::isa<DummyValVar>(this))
+    {
+        // Constants, function pointers, and dummy nodes don't require an ICFGNode.
+    }
+    else if (ty == ValNode)
+    {
+        // Base ValVar covers values without a dedicated subclass.
```
yuleisui: All these are normal ValVar and should be asserted?

svf/lib/SVFIR/SVFVariables.cpp
bjjwwang: sure

svf-llvm/lib/SVFIRBuilder.cpp
```diff
@@ -509,17 +509,22 @@ void SVFIRBuilder::initialiseValVars()
             pag->addGlobalValNode(iter->second, pag->getICFG()->getGlobalICFGNode(),
                                   llvmModuleSet()->getSVFType(llvmValue->getType()));
         }
-        else if (SVFUtil::isa<ConstantData, MetadataAsValue, BlockAddress>(llvmValue))
+        else if (SVFUtil::isa<ConstantData, ConstantExpr, MetadataAsValue, BlockAddress>(llvmValue))
```
bjjwwang: In the LLVM class hierarchy, `ConstantExpr` and `ConstantData` are siblings, not parent and child:

`Constant`

├── `ConstantData` (`ConstantInt`, `ConstantFP`, `UndefValue`, `ConstantPointerNull`, ...)

├── `ConstantExpr` (`getelementptr`, `bitcast`, ... an expression applied to a constant)

├── `ConstantAggregate`

├── `GlobalValue`

└── `BlockAddress`

`SVFUtil::isa<ConstantData>(llvmValue)` will not match `ConstantExpr`.

The value that slipped through the crack during debugging was a `ConstantExpr`:

`[DEBUG] base ValVar ID=5 ptr getelementptr inbounds ([2 x i32], ptr @a, i64 0, i64 1)`

This is a GEP constant expression on a global array. It doesn't have an `ICFGNode`. If `ConstantExpr` isn't added, it will fall into `else` and become `base ValVar`, triggering `assert(node)`. Therefore, ConstantExpr is not redundant; removing it will cause the program to crash.

svf-llvm/lib/SVFIRBuilder.cpp
```diff
@@ -502,20 +505,34 @@ void SVFIRBuilder::initialiseValVars()
...
+        else if (SVFUtil::isa<BasicBlock>(llvmValue))
+        {
+            pag->addBasicBlockValNode(iter->second, llvmModuleSet()->getSVFType(llvmValue->getType()));
+        }
+        else if (SVFUtil::isa<Instruction>(llvmValue) &&
+                 LLVMUtil::isIntrinsicInst(SVFUtil::cast<Instruction>(llvmValue)))
+        {
+            pag->addIntrinsicValNode(iter->second, llvmModuleSet()->getSVFType(llvmValue->getType()));
+        }
+        else if (!SVFUtil::isa<Instruction>(llvmValue))
```
yuleisui: Could we make this explicit rather than !instruction? 

if(InlineAsm || DSOLocalEquivalent || NoCFIValue)

svf-llvm/lib/SVFIRBuilder.cpp
yuleisui: It would be good to move the below code to line 524-527?

        if (const Instruction* inst =
                    SVFUtil::dyn_cast<Instruction>(llvmValue))
        {
            if (llvmModuleSet()->hasICFGNode(inst))
            {
                icfgNode = llvmModuleSet()->getICFGNode(inst);
            }
        }

yuleisui: and remove this part.

svf-llvm/lib/SVFIRBuilder.cpp
```diff
@@ -502,20 +505,34 @@ void SVFIRBuilder::initialiseValVars()
...
         }
         else if (SVFUtil::isa<ConstantAggregate>(llvmValue))
         {
             pag->addConstantAggValNode(iter->second, icfgNode, llvmModuleSet()->getSVFType(llvmValue->getType()));
         }
+        else if (SVFUtil::isa<BasicBlock>(llvmValue))
+        {
+            pag->addBasicBlockValNode(iter->second, llvmModuleSet()->getSVFType(llvmValue->getType()));
+        }
+        else if (SVFUtil::isa<Instruction>(llvmValue) &&
```
yuleisui: else if (SVFUtil::isa<Instruction>(llvmValue))
{
 if(LLVMUtil::isIntrinsicInst(SVFUtil::cast<Instruction>(llvmValue))))
   pag->addIntrinsicValNode(iter->second, llvmModuleSet()->getSVFType(llvmValue->getType()));
else
{
        assert(llvmModuleSet()->hasICFGNode(inst) && "LLVM instruction is not associated with an ICFGNode")
        icfgNode = llvmModuleSet()->getICFGNode(inst);
        pag->addValNode(iter->second, llvmModuleSet()->getSVFType(llvmValue->getType()), icfgNode);
}
  

}

svf-llvm/lib/SVFIRBuilder.cpp
```diff
@@ -41,6 +41,8 @@
 #include "Graphs/CallGraph.h"
 #include "Util/Options.h"
 #include "Util/SVFUtil.h"
+#include <llvm/IR/InlineAsm.h>
```
yuleisui: Pls put in the BasicTypes header and make typedefs:

https://github.com/SVF-tools/SVF/blob/a5e4ecdafafb2f8e444d1a1e5586b0b2059315de/svf-llvm/include/SVF-LLVM/BasicTypes.h#L45

bjjwwang:
> I will merge it if the crux bcs are ok with no assertions

passed

PR #1685, opened 2025-04-05
Added support for LLVM 20.1.1 and fix .editorconfig glob pattern for Vim

bsdb0y:

PR #1811, opened 2026-04-14
Support LLVM 22.1.0, but 4 CFL cases cannot pass. (maybe no need to pass)

bjjwwang:
Adapt SVF to build against LLVM 22, following ideas from PR #1685.

## TODO: 
1. build.sh wget llvm link
2. 4 CFL cases failed at arraycopy1.c.bc, maybe not necessary to fix.


yuleisui:
It looks quite good. Could you check the cases of field-sensitive Andersen analysis? I remember 1-2 cases will fail if we treat the first field as base. But we could turn off/on this option in the test-suite's cmakelist to pass those tests.

yuleisui:
Let us ignore the 4 CFL cases for now and make sure all other tests can pass.

Please also add a few AE tests for loop/recursion and array handling (in case the codecov decreases)

bjjwwang:
> Let us ignore the 4 CFL cases for now and make sure all other tests can pass.
> 
> Please also add a few AE tests for loop/recursion and array handling (in case the codecov decreases)

sure

PR #1813, opened 2026-04-21
Port SVF to LLVM 21

cjsrxzdyzds:
## Summary

This PR ports SVF to LLVM 21 and restores full test-suite parity on a clean LLVM 21 host.

The changes are intentionally scoped to SVF:
- update `svf-llvm` for LLVM 21 API/build compatibility
- harden `-read-ander` deserialization so analysis-state restoration no longer crashes on malformed or absent node records
- keep the current extapi import path working under LLVM 21 while factoring the logic into dedicated helpers for later LTO-safety follow-up

### LLVM 21 compatibility in `svf-llvm`
- replace removed `StringRef::equals()` calls with `operator==`
- migrate `DataLayout(Module*)` usage to `Module::getDataLayout()`
- guard LLVM APIs and types that are no longer available in newer releases
- guard `llvm_shutdown()` in `svf-ex`
- avoid accidental linkage against in-tree `LLVMSvf*` libraries when building standalone SVF against an LLVM tree that already contains an in-tree SVF integration

### `LLVMModule.cpp` cleanup for the LLVM 21 path
- split extapi import and cloning logic into dedicated helpers
- preserve current behavior while making the import path easier to replace in the planned LLVM 21 LTO-safety follow-up

### Robust `-read-ander` restoration
- skip empty or malformed serialized records
- skip missing base or var nodes where appropriate during restore
- preserve valid restored points-to targets required by later GEP-object restoration
- eliminate the crash path previously hit by standalone read/write regression tests

## Validation

Validated on the clean LLVM 21 host/build:

- build: passed
- Test-Suite: `1528 / 1528` passed

## Commits

- `b8beb7d7` `Update svf-llvm for LLVM 21`
- `6ccf4b43` `Harden read-ander result restoration`

## Notes

This PR is limited to the LLVM 21 port and test-suite correctness. The larger extapi or LTO import redesign can build on the helper seams introduced here.


yuleisui:
Thanks for the contribution. Could you also add a github action to trigger the build and tests?

cjsrxzdyzds:
one important CI concern I want to flag about the Test-Suite:

`SVF#1813` and `SVF-tools/Test-Suite#184` are a matched cross-repo update for the LLVM 21 port. The Test-Suite PR updates committed test fixtures (`test_cases_bc/*.bc`) to LLVM-21-compatible IR. That is why I temporarily pinned the LLVM 21 workflow in the SVF PR to the Test-Suite PR branch: otherwise Actions was testing the LLVM 21 SVF changes against stale default-branch fixtures;

however, if `Test-Suite#184` is merged directly into `SVF-tools/Test-Suite` master right now, the older SVF CI jobs will likely start failing. The reason is that the non-LLVM21 workflows still clone Test-Suite default branch, but they do not run with LLVM 21. Some of the updated `.bc` fixtures now contain LLVM-21-only textual IR syntax (for example `inrange` in constant GEPs), which older LLVM parsers reject.

so the situation is:
  - the Test-Suite changes are needed for correct LLVM 21 validation,
  - but merging them into the shared default fixture set may break the legacy CI jobs unless the workflows are version-isolated first.

yuleisui:
@bjjwwang could you help take a look at this pull request?

cjsrxzdyzds:
Test-Suite PR @https://github.com/SVF-tools/Test-Suite/pull/184 has been updated and should now show CI directly.

relevant commits on the Test-Suite PR branch:
 - [659e2991](https://github.com/SVF-tools/Test-Suite/pull/184/changes/659e2991b84f9acf2b83c467da13f8f979785773): switch CI bitcode generation to LLVM 21.1.0
 - [58f95cbd](https://github.com/SVF-tools/Test-Suite/pull/184/changes/58f95cbd700a005d6e43564c6ac47fe6bded5b73): run Test-Suite CI on pull requests

could you please verify this SVF PR against the updated Test-Suite PR branch

yuleisui:
> Test-Suite PR @[SVF-tools/Test-Suite#184](https://github.com/SVF-tools/Test-Suite/pull/184) has been updated and should now show CI directly.
> 
> relevant commits on the Test-Suite PR branch:
> 
> * [659e2991](https://github.com/SVF-tools/Test-Suite/pull/184/changes/659e2991b84f9acf2b83c467da13f8f979785773): switch CI bitcode generation to LLVM 21.1.0
> * [58f95cbd](https://github.com/SVF-tools/Test-Suite/pull/184/changes/58f95cbd700a005d6e43564c6ac47fe6bded5b73): run Test-Suite CI on pull requests
> 
> could you please verify this SVF PR against the updated Test-Suite PR branch

If test-suite bcs have been updated to llvm-21, we could use SVF's original CIs, and no need for a separate llvm-21.yml?

cjsrxzdyzds:
this PR has been updated to support the latest upstream Test-Suite on LLVM 21, and the branch has been revalidated end to end,
what changed at a high level:
 - updated the local validation target to the latest upstream Test-Suite
 - fixed LLVM 21 text-IR compatibility issues in the loader/normalization path
 - fixed extapi cloning compatibility for the newer LLVM path
 - fixed pointer-array / AE regressions exposed by the newer suite
 - fixed CFL virtual-call handling for the affected C++ cases
 - fixed the AE multi-entry regression for recursive library-style entry SCCs
 - removed the `std::regex`-based IR normalization so GCC 13 no longer fails on a libstdc++ false positive with `-Werror`
validation:
 - full LLVM 21 suite passed locally in `Release-build-llvm21`
 - result: `1959/1959` passed, `0` failed

please review the updated LLVM 21 compatibility changes and the analysis-side fixes. the branch is ready for another pass.

yuleisui:
Thanks for the updates. 

The external API changes look quite big. Could you please explain why those changes are necessary and possibly simplify the current pull request (as it involves 18 files).

cjsrxzdyzds:
Thanks for the review @yuleisui — agreed the diff is too broad. The current PR mixes: 
(a) the actual LLVM 21 API/IR-shape adjustments, 
(b) a LLVMModule::buildFunToFunMap helper extraction I had staged for a follow-up LTO-safety PR, 
and (c) AE/CFL/read-ander correctness fixes that the new Test-Suite fixtures exposed but which are independent of the LLVM 21 port itself. 

I'll restructure as follows over the next few days:
 - Drop the buildFunToFunMap extraction and the LLVMModuleSet ownership-mode rename — both preserve current behavior and were prep for the LTO follow-up. They'll land with that follow-up instead. 
 - Drop the parseIRFileCompat text-IR normalization fallback now that Test-Suite #185 ships LLVM 21-native bitcode. 
 - Move the AE entry-point / AbstractStateManager / CFL / read-ander / extapi-path fixes into a separate, independently reviewable PR. 
 - Squash the CI commits and the three pointer-array commits. 
 
What remains will be the real LLVM 21 surface: StringRef::equals removal, the CloneFunctionInto parent-attachment fix, the byte-offset rewrites in getBaseValueForExtArg / addComplexConsForExt / computeGepOffset (required because LLVM 21 emits opaque-pointer i8-byte GEPs), and the build/CI bumps. 
I'll re-validate against the full Test-Suite at every step and ping you when the slimmed branch is ready.

svf/lib/AE/Svfexe/AbstractInterpretation.cpp
```diff
@@ -94,33 +94,26 @@ AbstractInterpretation::~AbstractInterpretation()
...
-            }
-            else
-            {
-                entryFunctions.push_back(fun);
-            }
+            entryFunctions.push_front(fun);
+        }
+        else
+        {
+            entryFunctions.push_back(fun);
```
JoelYYoung: This step introduces functions that have external callers, meaning functions that should not be treated as analysis entries.

The table funToWTO contains:

1. **All non‑recursive functions** : Functions that do not belong to any recursive SCC in the call graph.

2. **Recursive functions that have at least one external caller**: These are functions inside a recursive SCC but are invoked from outside that SCC. They serve as entry points into recursive components and therefore require their own WTO construction.

cjsrxzdyzds: you're right. Iterating preAnalysis->getFuncToWTO() picks up SCC-entry recursive functions (which have external callers from outside their SCC) and treats them as analysis entries, which is wrong for whole-program AE.
I've reworked this in the latest revision: collectProgEntryFuns now walks the call graph directly and only collects functions with empty in-edges (no callers at all), prioritizing main() at the front of the duque. That matches the program-entry semantics the function name implies, with no SCC contamination.

cjsrxzdyzds: Thanks for the review. Just a heads-up — this file is no longer in the PR diff after the simplification..

The earlier `collectProgEntryFuns` rework turned out to be addressing the wrong layer. The real failure on `safe_ptr_array_access` was that LLVM 21 emits the first element of a pointer-array initialiser as a direct opaque-pointer store (`store ptr %v, ptr %arr`, no GEP), so SVF was writing the array base while later loop GEPs read field object `base_0` under `-ff-eq-base=false`. That's now fixed at the SVFIRBuilder layer in `46e9a345` ("Handle LLVM 21 opaque-pointer array accesses in SVFIRBuilder") via a synthetic field-zero GEP value when the access matches an inferred `[K x ptr]` base. AE entry-point semantics are untouched in this PR.
Your `funToWTO` observation still stands as a separate concern in `AbstractInterpretation.cpp` and would be worth a follow-up issue/PR on master — outside the LLVM 21 port scope.

cjsrxzdyzds:
closing this pr(collateral damage from the rename), PR #1815 is OPEN
@JoelYYoung — flagging that `AbstractInterpretation.cpp` is no longer in the PR diff

PR #1816, opened 2026-04-30
Fix missing push() before pop() in SaberCondAllocator::isSatisfiable

Patrickstarrrrr:
## Bug Description

`SaberCondAllocator::isSatisfiable()` calls `solver.pop()` without a matching `solver.push()`. This causes a Z3 "index out of bounds" exception when the solver's scope stack is empty. Additionally, the added constraint is never popped, permanently polluting the solver state.

## Fix

Add the missing `push()` before `add()`, consistent with the adjacent `isEquivalentBranchCond()` and the usage pattern in `Z3Expr.cpp`.


yuleisui:
Thanks for the fix.

Issue #1817, opened 2026-05-03
fspta fails on global initializer


hiruwaKS:
**demo:**
```cpp
void f() {}
void (*g)() = f;
void (*h)() = g;

int main() {
    h();           // should resolve to f, but failed
    return 0;
}
```
**Params:**: Release-build/bin/wpa -fspta -v-call-cha -merge-pwc -vt-in-ir -ff-eq-base -dump-callgraph

And I found `-ander` works well.

Possible solution: I think this could be related to the function `__cxx_global_var_init` , which is called before `main`.

It feels like a subtle corner case, though. Thank you!

yuleisui:
Thanks for reporting it. Would you like to fix this with a pull request?

yuleisui:
Please close it if this has been fixed

PR #1818, opened 2026-05-05
fix: the entry point should be "svf_main" if using `-svf-main`

hiruwaKS:
**Description:**

This PR fixes a case [issue1817](https://github.com/SVF-tools/SVF/issues/1817) where indirect calls through global function pointers fail to resolve the target due to global initializer.

**Cause:**  
We need a "dummy" `main` to put the functions in `llvm::global_ctor` into call graph, however the original `-svf-main` option's implementation has some bugs. Several parts of the SVF project still consider `main` as the entry even with `-svf-main`  option on.

**Fix:**  
Changed the entry point from `main` to `svf.main` when `-svf-main` is specified.

**Testing:**  
- Tested the provided demo in issue
- Tested with my own test suite (26 c/cpp programs), it does not introduce new bugs.

### DIFF
--- a/svf-llvm/lib/LLVMModule.cpp
+++ b/svf-llvm/lib/LLVMModule.cpp
@@ -597,7 +597,12 @@ void LLVMModuleSet::addSVFMain()
         }
         // return;
         Builder.CreateRetVoid();
-    } else Options::SVFMain.setValue(false); // use SVF::Options to record whether svf.main is added or not
+    }
+    else
+    {
+        // use SVF::Options to record whether svf.main is added or not
+        Options::SVFMain.setValue(false);
+    }
 }
 
 void LLVMModuleSet::collectExtFunAnnotations(const Module* mod)
--- a/svf-llvm/lib/LLVMModule.cpp
+++ b/svf-llvm/lib/LLVMModule.cpp
@@ -542,12 +542,12 @@ void LLVMModuleSet::addSVFMain()
             }
         }
     }
-    // Keep it simple, and reduce checks
-    // // Only create svf.main when the original main function is found, and also
-    // // there are global constructor or destructor functions.
-    // if (orgMain && getModuleNum() > 0 &&
-    //         (ctor_funcs.size() > 0 || dtor_funcs.size() > 0))
-    // {
+
+    // Only create svf.main when the original main function is found, and also
+    // there are global constructor or destructor functions.
+    if (orgMain && getModuleNum() > 0 &&
+            (ctor_funcs.size() > 0 || dtor_funcs.size() > 0))
+    {
         assert(mainMod && "Module with main function not found.");
         Module& M = *mainMod;
         // char **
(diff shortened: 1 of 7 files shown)


svf/lib/AE/Svfexe/AbstractInterpretation.cpp
```diff
@@ -125,7 +113,8 @@ std::deque<const FunObjVar*> AbstractInterpretation::collectProgEntryFuns()
         if (cgNode->getInEdges().empty())
         {
             // If main exists, put it first for priority using deque's push_front
-            if (fun->getName() == "main")
+            const char* main_name=Options::SVFMain() ? "svf.main" : "main";
+            if (fun->getName() == main_name)
```
yuleisui: Could we invoke isProgEntryFunction in SVFUtil.cpp, so that we could remove the options here?

svf/lib/MTA/MHP.cpp
```diff
@@ -567,7 +567,8 @@ bool MHP::isConnectedfromMain(const FunObjVar* fun)
     while (!worklist.empty())
     {
         const CallGraphNode* node = worklist.pop();
-        if ("main" == node->getFunction()->getName())
+        const char* main_name=Options::SVFMain() ? "svf.main" : "main";
```
yuleisui: Similarly, We can use invoke isProgEntryFunction in SVFUtil.cpp, so that we could remove the options here.

svf-llvm/lib/LLVMModule.cpp
```diff
@@ -533,21 +533,21 @@ void LLVMModuleSet::addSVFMain()
...
-    // there are global constructor or destructor functions.
-    if (orgMain && getModuleNum() > 0 &&
-            (ctor_funcs.size() > 0 || dtor_funcs.size() > 0))
-    {
+    // Keep it simple, and reduce checks
+    // // Only create svf.main when the original main function is found, and also
+    // // there are global constructor or destructor functions.
+    // if (orgMain && getModuleNum() > 0 &&
+    //         (ctor_funcs.size() > 0 || dtor_funcs.size() > 0))
+    // {
```
yuleisui: Not quite sure why the code is removed here? Please explain.

svf/lib/MTA/MHP.cpp
hiruwaKS: ok, i'll fix

svf-llvm/lib/LLVMModule.cpp
hiruwaKS: I think it will add a check for whether function `svf.main` is in the module set if `-svf-main` is specified, in usage like `isProgEntryFunction`. It may cause a bit more time to enumerate the function names, so I choose to keep it simple.

hiruwaKS: I figure out a new solution to this, please see the new commit

svf-llvm/lib/LLVMModule.cpp
```diff
@@ -597,7 +597,7 @@ void LLVMModuleSet::addSVFMain()
         }
         // return;
         Builder.CreateRetVoid();
-    }
+    } else Options::SVFMain.setValue(false); // use SVF::Options to record whether svf.main is added or not
```
yuleisui: Please format this line before I merge.

````
else
{
     Options::SVFMain.setValue(false);
}
````

PR #1820, opened 2026-05-06
Full sparse Pull Request 1

bjjwwang:
Summary
=======
Refactors the full-sparse AE merge into a pure SVFG-pull model, breaking the
direct dependency on the base class's ICFG-edge `joinStates` propagation.

What's done
-----------
* `mergeStatesFromPredecessors` is now a single-step SVFG-driven pull.
  For each VFG node attached to the current ICFG node, walk its indirect
  SVFG in-edges; for each obj id in an edge's points-to, pull the value
  from the source SVFGNode's icfg trace (multi-source -> JOIN). No more
  base-class delegation, no copy-then-prune, no liveness precompute.
* `joinStates` reduced to a no-op (full-sparse no longer propagates
  state along ICFG edges).
* GepObj overlay (Phase 2) preserved in `loadValue` for field-precision
  cross-function reads.
* Phase 1 ValVar routing through SVFG def-site preserved
  (with semi-sparse fallback for ValVars without an SVFG def).

What's not yet done (tracked as a TODO inside the new merge)
------------------------------------------------------------
1. `_freedAddrs` ICFG-edge union for the null-deref detector.
   Without this, nullderef precision regresses.
2. Branch refinement (cmp / switch narrowing).
   Under pure SVFG-pull, the cmp icfg's trace doesn't carry the obj that
   the backing load reads, so `isBranchFeasible`'s narrowing silently
   no-ops. Plan: at each conditional intra in-edge, walk the condition
   var back to its backing LoadStmt, pull those obj from the load's
   icfg trace into a local predState, run `isBranchFeasible`, capture
   the narrowed entries into a `branchOverlay` keyed by
   (branch-succ ICFGNode*, obj NodeID); `loadValue` consults it before
   the trace fallback.

Status
------
Builds; not regression-tested in this push. Restoring the two TODO items
should bring sparse precision back to (and likely past) the prior 80% mark
on the AE triage suite.


svf/lib/AE/Svfexe/SparseAbstractInterpretation.cpp
```diff
@@ -122,6 +123,161 @@ const Set<const ICFGNode*> FullSparseAbstractInterpretation::getUseSitesOfObjVar
...
+//  joinStates is a no-op safety net in case base machinery calls it.
+// =====================================================================
+
+void FullSparseAbstractInterpretation::joinStates(
+    AbstractState&, const AbstractState&) // add one more ICFGNODE AS PARAM
+{
+    // No-op: full-sparse doesn't propagate state along ICFG edges.
+}
+
+bool FullSparseAbstractInterpretation::mergeStatesFromPredecessors(
```
yuleisui, line 141: Do we still need to override this method?

svf/lib/AE/Svfexe/SparseAbstractInterpretation.cpp
```diff
@@ -122,6 +123,161 @@ const Set<const ICFGNode*> FullSparseAbstractInterpretation::getUseSitesOfObjVar
...
+//  to [7,11] rather than strong-updating to 11).  Same trade-off as
+//  the 0413 reference design.
+//
+//  TODO: AbsExtAPI's memcpy / memmove / strcpy handlers call
+//  as.store(addr,...) directly, bypassing this override.  If the
+//  destination addr decodes to a GepObjVar, the value won't enter
+//  the overlay — known precision leak; revisit if benchmark exposes.
+// =====================================================================
+
+void FullSparseAbstractInterpretation::storeValue(
```
yuleisui, line 235: Looks no need to override store value and load value methods

svf/lib/AE/Svfexe/SparseAbstractInterpretation.cpp
bjjwwang: no need, I am removing.

PR #1824, opened 2026-05-13
fix: skip callsite if InlineAsm

Khosro-Moeini:
Inline assembly is considered an indirect call site in LLVM IR, but not something SVF can process.

### DIFF
--- a/svf-llvm/lib/SVFIRBuilder.cpp
+++ b/svf-llvm/lib/SVFIRBuilder.cpp
@@ -1256,6 +1256,9 @@ void SVFIRBuilder::visitCallSite(CallBase* cs)
     if(isIntrinsicInst(cs))
         return;
 
+    if(cs->isInlineAsm())
+        return;
+
     DBOUT(DPAGBuild,
           outs() << "process callsite " << svfcall->valueOnlyToString() << "\n");
 


yuleisui:
Thanks for the pull request. Would not skipping it cause a crash? How did you spot it?

Khosro-Moeini:
I spotted such cases while analyzing musl libc. Not skipping them would not cause a crash, but would result in many lines like the following in the -print-fp output:

```
NodeID: 59406
CallSite: CallICFGNode83972 {fun: unlink}
   %3 = tail call i64 asm sideeffect "syscall", "={ax},{ax},{di},~{rcx},~{r11},~{memory},~{dirflag},~{fpsr},~{flags}"(i64 87, i64 %2) #108, !srcloc !8872 CallICFGNode:        Location: CallICFGNode:
        !!!has no targets!!!
```

I don't think there is a point in processing these call sites. Also, including them would make the final stats inaccurate.

yuleisui:
> I spotted such cases while analyzing musl libc. Not skipping them would not cause a crash, but would result in many lines like the following in the -print-fp output:
> 
> ```
> NodeID: 59406
> CallSite: CallICFGNode83972 {fun: unlink}
>    %3 = tail call i64 asm sideeffect "syscall", "={ax},{ax},{di},~{rcx},~{r11},~{memory},~{dirflag},~{fpsr},~{flags}"(i64 87, i64 %2) #108, !srcloc !8872 CallICFGNode:        Location: CallICFGNode:
>         !!!has no targets!!!
> ```
> 
> I don't think there is a point in processing these call sites. Also, including them would make the final stats inaccurate.

Could you let us know what cashes it triggered?

As inlineASM was kept as a value in [SVFIRBuilder](https://github.com/SVF-tools/SVF/blob/167f2673e92bc61a7dd0fc84fb6ec3b1174cc941/svf-llvm/lib/SVFIRBuilder.cpp#L517), and also it is a subclass of CallBase. Though we don't handle it now, it could be handled later when someone is interested in analyzing inline asm.

I would suggest that we fix your crash and retain the inline asm during SVFIR building.

Khosro-Moeini:
As I said, I did not spot a crash. I just thought inlineAsm might have been overlooked, but it seems it was intentionally included. I'll go ahead and close this. Thanks.

PR #1826, opened 2026-05-19
Add AE function entry mode on latest master

bjjwwang:


svf/lib/Util/Options.cpp
```diff
@@ -799,6 +799,20 @@ const OptionMap<u32_t> Options::AESparsity(
         "Sparse abstract execution via SVFG."
     }
 });
+const OptionMap<u32_t> Options::FunEntry(
+    "fun-entry",
```
yuleisui: ae-fun-entry

svf/include/AE/Svfexe/AbstractInterpretation.h
```diff
@@ -94,6 +94,12 @@ class AbstractInterpretation
         WIDEN_NARROW
     };
 
+    enum FunEntryMode
```
yuleisui: AEFunEntryMode

svf/lib/AE/Svfexe/AbstractInterpretation.cpp
```diff
@@ -749,12 +797,23 @@ void AbstractInterpretation::handleFunction(const ICFGNode* funEntry, const Call
...
         else if (const ICFGCycleWTO* cycle = SVFUtil::dyn_cast<ICFGCycleWTO>(comp))
         {
-            if (mergeStatesFromPredecessors(cycle->head()->getICFGNode()))
+            const ICFGNode* node = cycle->head()->getICFGNode();
+            if (node->getInEdges().empty())
+            {
+                updateAbsState(node, getAbsState(globalNode));
+                handleLoopOrRecursion(cycle, caller);
+            }
+            else if (mergeStatesFromPredecessors(node))
```
yuleisui: It is unclear to me why we need to change this?

bjjwwang: > It is unclear to me why we need to change this?

Because GlobalICFGNode just connects to FunEntryNode of Main Function. If there's no main function, FunEntryNode of no-caller Functions should inherit actively from GlobalNode.

svf/include/AE/Svfexe/AbstractInterpretation.h
bjjwwang: > AEFunEntryMode

fixed

svf/lib/Util/Options.cpp
bjjwwang: > ae-fun-entry

fixed

svf/lib/AE/Svfexe/AbstractInterpretation.cpp
yuleisui: It should be handled in this loop at line 176 (entryFunctions should include all no-caller functions):

    // handle Global ICFGNode of SVFModule
    handleGlobalNode();
    for (const FunObjVar* entryFun : entryFunctions)
    {
        const ICFGNode* funEntry = icfg->getFunEntryICFGNode(entryFun);
        updateAbsState(node, getAbsState(globalNode));
        handleFunction(funEntry);
    }

PR #1830, opened 2026-06-01
fix: prevent abort when Value is ConstantData

Khosro-Moeini:
As per https://github.com/llvm/llvm-project/commit/9383fb23e18bb983d0024fb956a0a724ef9eb03d, LLVM 21 no longer permits the inspection of `users` of `ConstantData`. I encountered two crashes due to this change.

I added checks to guard the two calls that caused the crashes.


yuleisui:
Thanks for the fix. It would be good to check any other places of the uses of constantdata.

PR #1841, opened 2026-06-17
Remove BasicBlock Values

mbarbar:
They do not appear to be used anywhere.

### DIFF
--- a/svf-llvm/lib/SVFIRBuilder.cpp
+++ b/svf-llvm/lib/SVFIRBuilder.cpp
@@ -467,19 +467,18 @@ void SVFIRBuilder::initialiseValVars()
         const ICFGNode* icfgNode = nullptr;
         auto llvmValue = iter->first;
 
-        // Check if the value is a function and get its call graph node
-        if (const Function* func = SVFUtil::dyn_cast<Function>(llvmValue))
-        {
-            pag->addFunValNode(iter->second, icfgNode, llvmModuleSet()->getFunObjVar(func), llvmModuleSet()->getSVFType(llvmValue->getType()));
-        }
-        else if (
+        // Intentionally unhandled values.
+        if (
             SVFUtil::isa<BasicBlock>(llvmValue) ||
             SVFUtil::isa<InlineAsm>(llvmValue) ||
             SVFUtil::isa<DSOLocalEquivalent>(llvmValue) ||
             SVFUtil::isa<NoCFIValue>(llvmValue)
-        )
+        ) { continue; }
+
+        // Check if the value is a function and get its call graph node
+        if (const Function* func = SVFUtil::dyn_cast<Function>(llvmValue))
         {
-            // Intentionally unhandled.
+            pag->addFunValNode(iter->second, icfgNode, llvmModuleSet()->getFunObjVar(func), llvmModuleSet()->getSVFType(llvmValue->getType()));
         }
         else if (auto argval = SVFUtil::dyn_cast<Argument>(llvmValue))
         {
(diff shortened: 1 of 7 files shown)


svf-llvm/lib/SVFIRBuilder.cpp
```diff
@@ -510,10 +510,6 @@ void SVFIRBuilder::initialiseValVars()
         {
             pag->addConstantAggValNode(iter->second, icfgNode, llvmModuleSet()->getSVFType(llvmValue->getType()));
         }
-        else if (SVFUtil::isa<BasicBlock>(llvmValue))
-        {
-            pag->addBasicBlockValNode(iter->second, llvmModuleSet()->getSVFType(llvmValue->getType()));
```
yuleisui, line 515: I would still keep this branch but remove "pag->addBasicBlockValNode(iter->second, llvmModuleSet()->getSVFType(llvmValue->getType()));" and put a comment.

It is better to move this branch just below the branch "if (const Function* func = SVFUtil::dyn_cast<Function>(llvmValue))", so it is clearer that we don;t handle basicblock.

mbarbar: Yep okay. I'm planning later to open a PR removing ASMPCValNode. Should I do the same for its branch?

yuleisui: It is good to put it under the same pr.

mbarbar: Pushed its removal.

svf-llvm/lib/SVFIRBuilder.cpp
```diff
@@ -529,6 +521,12 @@ void SVFIRBuilder::initialiseValVars()
                 pag->addValNode(iter->second, llvmModuleSet()->getSVFType(llvmValue->getType()), icfgNode);
             }
         }
+        else
+        {
+            pag->addDummyValNode(iter->second, icfgNode);
```
yuleisui: Why did you add a dummy for any unhandled node? Also good to make a comment here.

yuleisui: Would this cause a new dummy node to be created for any LLVM value not on PAG?

mbarbar: Yes. I'm experimenting a bit but I'm not really sure what we should do. Some observations:

From what I gather from the log, before BB nodes and ASM etc. nodes were added, I believe it was adding a standard Val node.

SVF expects the number of PAG nodes to be same or greater than the number of symbols. To ignore these symbols in the first place, well they actually exist in the IR and are referred to. For the ASM nodes, I found a case where `getCalledOperand()` returns an inline ASM node so it needs a corresponding Val node, but we were never properly analysing any ASM node...

So I'm conflicted. On the one hand these nodes are referred to in the IR, on the other hand our model in SVF has been ignoring them.

yuleisui: I think it is best to ignore these LLVM values and do not create nodes for these values if we don't analyse them. Which ones we may analyse later?

yuleisui: Dummy nodes contains a bit too many types of llvm values and if we keep these llvm values in the form of dummy nodes, it will be a bit hard to maintain later

mbarbar: How should we deal with e.g this
```
void SVFIRBuilder::handleIndCall(CallBase* cs)
{
    const CallICFGNode* cbn = llvmModuleSet()->getCallICFGNode(cs);
    NodeID indFunPtrId = llvmModuleSet()->getValueNode(cs->getCalledOperand());
    const_cast<CallICFGNode*>(cbn)->setIndFunPtr(pag->getGNode(indFunPtrId));
    pag->addIndirectCallsites(cbn,indFunPtrId);
}
```
`cs->getCalledOperand()` returns an `InlineAsm` `Value` which is fed into `getGNode`.

The node has forever existed I believe, first as a generic Val node then as an InlineAsmVal node.

yuleisui: I see, then we may have to keep AsmPCValNode? Let us do a removal of other unused values first and make this as an issue or a separate pull request.

mbarbar: Mm okay, I will try early next week.

yuleisui: Thanks, I will merge it soon. It would be good to make another pull request for AsmPCValNode later.

PR #1842, opened 2026-06-17
Remove llvm var ID from GEP Val node

mbarbar:
It is unused.


mbarbar:
Unless it is used for debugging?

mbarbar:
Never mind, it is used by the graph DB.

Issue #1843, opened 2026-06-18
SVF tries to run NPM even if it is not installed


mbarbar:
In `svf/lib/Util/ExtAPI.cpp`, SVF may try to look in `npm root`. That means actually running `npm` and `npm` may not be installed. In such a case we get a couple of `sh npm not found` type errors printed out.

PR #1844, opened 2026-06-19
fix issue #1843: handling case if npm is not installed

yuleisui:


svf/lib/Util/ExtAPI.cpp
```diff
@@ -71,50 +71,68 @@ bool ExtAPI::setExtBcPath(const std::string& path)
...
+        if (!svfdir)
+            return "";
+
+        bcFilePath = svfdir;
+
         if (!bcFilePath.empty() && bcFilePath.back() != '/')
             bcFilePath.push_back('/');
-        bcFilePath.append(SVF_BUILD_TYPE "-build").append("/lib/extapi.bc");
+
+        bcFilePath.append(SVF_BUILD_TYPE "-build/lib/extapi.bc");
```
mbarbar, line 117: I know this was in the original code before but what even is SVF_BUILD_TYPE? Grepping I only see it defined in CMAKE stuff. And syntactically, the only (?) way this can be okay is if it is a macro evaluating to a string literal. But obviously it's compiling fine, I don't get why.

mbarbar (approved):
Aside from the one comment, LGTM!

svf/lib/Util/ExtAPI.cpp
yuleisui: I can recall where SVF_BUILD_TYPE is defined, I guess it is either release or debug? Could you check?

mbarbar: Ah yep, it is in cmake/SVFConfigHdr.cmake.in
```
/**
 * SVF instance configuration header
 *
 * WARNING: This file is automatically generated; changes will be overwritten!
 *
 */

#ifndef SVF_CONFIG_H_IN
#define SVF_CONFIG_H_IN

// Expose base properties of this SVF instance
#define SVF_VERSION "@SVF_VERSION@"
#define SVF_BUILD_TYPE "@SVF_BUILD_TYPE@"
```

Then I assume that `@` thing comes from CMakeLists.txt
```
set(SVF_BUILD_TYPE ${CMAKE_BUILD_TYPE})
```

PR #1846, opened 2026-06-28
Add more logging to setup.sh

mbarbar:


setup.sh
```diff
@@ -74,3 +74,5 @@ export PATH=$SVF_DIR/$Build/bin:$PATH
 
 # Add compiled library directories to $LD_LIBRARY_PATH
 export LD_LIBRARY_PATH=$SVF_DIR/$Build/svf:$SVF_DIR/$Build/svf-llvm:$LD_LIBRARY_PATH
+
+echo "Added SVF, LLVM, and Z3 to \$PATH, \$LD_LIBRARY_PATH, and \$DYLD_LIBRARY_PATH."
```
yuleisui: Add these three respectively or each to \$PATH, \$LD_LIBRARY_PATH, and \$DYLD_LIBRARY_PATH?

Would be good to make it clear.

mbarbar: Fixed.

PR #1851, opened 2026-06-30
Pull some enums out of major classes

mbarbar:
PTATY out of PointerAnalysis avoids a circular dependency for GraphWriter.h to use Options.h. This, I think, is also better, that places like Options.h have as few dependencies as possible. It also unraveled a lot of dependencies being included transitively (hence the size of this PR). Overall, it would be good to spend some time looking at dependencies etc.


yuleisui:
This is still marked as draft? Let me know once ready

mbarbar:
Oops sorry, yeah it's ready.

PR #1862, opened 2026-07-07
Simplify #includes

mbarbar:
Based on the trouble from PTATY, I decided to make #includes a little simpler by shuffling #includes around and removing some unnecessary #includes. This isn't comprehensive (done manually) but it works. It seems to build a little faster from scratch (could be noise) and in theory, we should also have faster incremental builds.


yuleisui:
Would this affect the Software-Security-Analysis repo before I do the merge?

mbarbar:
Tested and builds fine.

PR #1863, opened 2026-07-08
Fix build against LLVM 16-20: Value::hasUseList() doesn't exist before LLVM 21

Acture:
## Problem

Building SVF's `svf-llvm` target against LLVM 16-20 fails to compile:

```
error: 'const SVF::Value' {aka 'const class llvm::Value'} has no member named 'hasUseList';
did you mean 'llvm::Use* llvm::Value::UseList'? (not accessible from this context)
```

`llvm::Value::hasUseList()` doesn't exist before LLVM 21 (see #1864 for the version history
and why iterating `users()` on a `ConstantData` value without it can abort). `ObjTypeInference.cpp`
calls `hasUseList()` at two call sites as a guard before iterating `->users()`.

## Fix

Version-conditional shim: use the native `hasUseList()` on LLVM 21+, fall back to the
equivalent `!isa<ConstantData>(v)` check on older versions.

```cpp
#if LLVM_VERSION_MAJOR >= 21
static inline bool hasUseList(const Value* v) { return v->hasUseList(); }
#else
static inline bool hasUseList(const Value* v) { return !SVFUtil::isa<ConstantData>(v); }
#endif
```

`isa` has to be qualified as `SVFUtil::isa` here, otherwise it's ambiguous with `llvm::isa`
inside SVF's own namespace.

## Verification

Built the full `svf-llvm` static library against `llvm-18-dev` 18.1.3, `llvm-20-dev` 20.1.x,
and this repo's pinned LLVM 21.1.8. All three compile and link cleanly.

No other call sites in the codebase reference `hasUseList()`.

### DIFF
--- a/svf-llvm/lib/ObjTypeInference.cpp
+++ b/svf-llvm/lib/ObjTypeInference.cpp
@@ -77,6 +77,18 @@ using namespace SVFUtil;
 using namespace LLVMUtil;
 using namespace cppUtil;
 
+// llvm::Value::hasUseList() was added in LLVM 21 alongside the change that
+// stopped instances of ConstantData (e.g. plain integer/float constants)
+// from carrying a use-list at all (see llvm/llvm-project@87f312a, "IR:
+// Remove uselist for constantdata"). On LLVM < 21 the method doesn't exist,
+// but the exact same condition is available through the public
+// isa<ConstantData> check that hasUseList() itself wraps internally.
+#if LLVM_VERSION_MAJOR >= 21
+static inline bool hasUseList(const Value* v) { return v->hasUseList(); }
+#else
+static inline bool hasUseList(const Value* v) { return !SVFUtil::isa<ConstantData>(v); }
+#endif
+
 
 const std::string TYPEMALLOC = "TYPE_MALLOC";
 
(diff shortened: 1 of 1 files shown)


yuleisui:
As the current SVF successfully builds on top of LLVM-21, it is unclear whether your pull request is only for LLVM-18 or why this pr is needed?

Acture:
> As the current SVF successfully builds on top of LLVM-21, it is unclear whether your pull request is only for LLVM-18 or why this pr is needed?

Sorry, the title was misleading. hasUseList() was only added in LLVM 21 (see https://github.com/SVF-tools/SVF/issues/1864), so it's not there on 16 through 20. That's why it builds fine on LLVM-21 for you.

I also changed the fix. Deleting the guard compiled but could bring back an abort that a previous commit (https://github.com/SVF-tools/SVF/commit/e83f846a41a72e7a4137f084b17f58a66873ae14) was meant to prevent. Now it uses the real hasUseList() on 21+ and falls back to an equivalent check below that. Rebuilt and tested against llvm-18-dev, llvm-20-dev, and LLVM 21.1.8, and squashed into one commit.

PR #1867, opened 2026-07-10
Code formatting for pull requests

yuleisui:
Add astyle formatting and clang-format for changed files

### DIFF
--- a/.github/workflows/svf-lib_binaries.yml
+++ b/.github/workflows/svf-lib_binaries.yml
@@ -111,7 +111,7 @@ jobs:
             git push https://yuleisui:${{secrets.yulei_git_key}}@github.com/SVF-tools/Test-Suite.git   HEAD:master
           fi
 
-      # update-svf-lib
+      # update-svf-lib binaries to svf-npm for later publishing
       - name: update-svf-lib
         if: |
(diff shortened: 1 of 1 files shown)


.github/workflows/github-action.yml
```diff
@@ -48,6 +49,8 @@ jobs:
           echo $(pwd)
           if [ "${{matrix.sanitizer}}" != "" ]; then export SVF_SANITIZER="${{matrix.sanitizer}}"; fi
           if [ "$RUNNER_OS" == "Linux" ] && [ "${{matrix.sanitizer}}" == "" ]; then export SVF_COVERAGE=1; fi
+          astyle --style=allman -n -r "*.h" "*.cpp"
+          if [ ! -z $FILES_CHANGED ]; then clang-format -i $FILES_CHANGED; fi
```
mbarbar: What is the purpose of clang-format + astyle rather than just one? Also, could you add a bad style change to some random file just so we could see what CI does?

yuleisui: Good idea, better to have clang only?

mbarbar: Not really familiar with them but that sounds better.

PR #1872, opened 2026-07-15
Remove Constant Aggregate nodes from SVFIR

mbarbar:
These do not appear to be used anywhere.


mbarbar:
I'm not really sure if we want to keep this for future or anything, but I incline to think if we don't need it now, we don't need to have it now.

PR #1873, opened 2026-07-21
Process rep nodes when collapsing an SCC

mbarbar:
When collapsing any cycle, we need to potentially handle stores/loads (postProcessNode) and copies/geps (processNode) if the points-to set has changed or if the rep node gets new edges. I originally thought not so for the latter but SCC detection can happen before all nodes in the SCC get the same points-to set.

Adding to worklist/reanalyzing conditionally upon those two conditions requires a lot of refactoring I believe. I've instead added the rep node unconditionally whenever an SCC is collapsed.

### DIFF
--- a/svf/lib/WPA/Andersen.cpp
+++ b/svf/lib/WPA/Andersen.cpp
@@ -722,6 +722,10 @@ void Andersen::mergeSccCycle()
         const NodeBS& subNodes = getSCCDetector()->subNodes(repNodeId);
         // merge sub nodes to rep node
         mergeSccNodes(repNodeId, subNodes);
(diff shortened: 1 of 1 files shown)


mbarbar:
Interestingly this solves a real problem but the stack-1.bc stuff we were looking at before has some other problems too it turns out. Investigating still (that is probably the access-from-0-offset vs. access-from-base you mentioned but I want to confirm it).

yuleisui:
This is an issue orthogonal to the gep offset. Shall I merge this one first?

mbarbar:
Yes, it should be good to merge.

PR #1874, opened 2026-07-21
Double up GEP edges in graph output

mbarbar:
I have trouble differentiating the purple from blue, especially when zoomed out.
image
image
Left is old, right is new.


yuleisui:
The current colour is a little too light and, unfortunately, appears almost black to me.

Could we make your purple a bit darker, or distinguish the arrow in another way, such as by changing its decoration or shape?

mbarbar:
How about double lines?
image
Dashed and dotted are used for other things, so I thought best not to use that.

PR #1877, opened 2026-08-03
Fix AE pointer comparison and branch feasibility

bjjwwang:
## Summary

A points-to set records alternative targets for one runtime pointer. A known single target such as `{a}` identifies the pointer's exact target. A multi-target set such as `{a, b}` means that either target may be selected at runtime. Consequently, two equal multi-target sets do not prove that the two concrete pointers are equal.

- Normalize pointer operands into known address sets, `{NULL}`, or `{BlackHole}`.
- For `eq` and `ne`, produce a definite equality result only for the same known single target, and a definite inequality result for disjoint known target sets.
- Keep overlapping sets with multiple possible targets and comparisons involving `BlackHole` conservative as `[0, 1]`.
- Keep pointer-ordering predicates conservative because abstract object IDs do not encode runtime address order.
- Use the computed comparison result for branch feasibility, with a conservative fallback when no interval result is available.
- Remove the duplicate predicate-aware pointer block that was unreachable behind the original address fast path.

## Problem

`AbstractInterpretation::updateStateOnCmp` handled every comparison between two address values as if it were equality. Consequently, `icmp ne` on the same known target evaluated to true, while `icmp ne` on disjoint targets evaluated to false. The predicate-aware address switch later in the function never ran because the earlier address branch had already consumed those operands.

The old logic also treated equality of two abstract address sets as concrete pointer equality. This is valid for `{a}` compared with `{a}`, but not for `{a, b}` compared with `{a, b}`: the two runtime pointers may independently select different targets. Similarly, `BlackHole` denotes a statically unresolved target for this comparison and cannot be compared as if it were one concrete address.

The old relational cases compared abstract object IDs. These IDs identify analysis objects; their numeric order does not represent runtime pointer order. In addition, `isCmpBranchEdgeFeasible` bypassed pointer and null comparisons, so even a sound comparison result could not refine branch reachability.

## Semantics

For pointer comparison, the implementation follows these cases:

| Left targets | Right targets | `==` | `!=` |
|---|---|---:|---:|
| `{a}` | `{a}` | `[1,1]` | `[0,0]` |
| `{a}` | `{b}`, where `a != b` | `[0,0]` | `[1,1]` |
| `{a, b}` | `{c}`, where the sets are disjoint | `[0,0]` | `[1,1]` |
| `{a, b}` | `{a}` | `[0,1]` | `[0,1]` |
| `{a, b}` | `{a, b}` | `[0,1]` | `[0,1]` |
| a set containing `BlackHole` | any set | `[0,1]` | `[0,1]` |

`NULL` is a known distinguished target, so two null operands follow the same-known-target case. Pointer-ordering predicates remain `[0,1]`.

## Validation

- The `ae` target builds successfully with LLVM 21.1.8 and Z3 4.16.0.
- Focused pointer/null AE regressions: 28/28 passed across dense, semi-sparse, full-sparse, and no-main modes.
- Targeted comparison tests verify definite results for the same and disjoint known targets.
- A targeted overlapping-target test confirms that both feasible branches are retained.

### DIFF
--- a/svf/lib/AE/Svfexe/AbstractInterpretation.cpp
+++ b/svf/lib/AE/Svfexe/AbstractInterpretation.cpp
@@ -488,30 +488,63 @@ static IntervalValue computeCmpConstraint(s32_t predicate, s64_t succ,
     return result;
 }
 
+/// Evaluate a comparison between abstract pointer values. Object identifiers
+/// are not runtime addresses, so ordering comparisons are only decidable when
+/// both operands denote the same singleton object.
+static IntervalValue compareAddressValues(s32_t predicate,
+                                          const AddressValue& lhs,
+                                          const AddressValue& rhs)
+{
+    const bool bothNull = lhs.empty() && rhs.empty();
+    const bool oneNull = lhs.empty() != rhs.empty();
+    const bool sameSingleton =
+        lhs.size() == 1 && rhs.size() == 1 && *lhs.begin() == *rhs.begin();
+    const bool sameValue = bothNull || sameSingleton;
+    const bool disjoint = !lhs.hasIntersect(rhs);
+
+    switch (predicate)
+    {
+    case CmpStmt::ICMP_EQ:
+        if (sameValue)
+            return IntervalValue(1, 1);
+        if (!oneNull && disjoint)
+            return IntervalValue(0, 0);
+        return IntervalValue(0, 1);
+    case CmpStmt::ICMP_NE:
+        if (sameValue)
+            return IntervalValue(0, 0);
+        if (!oneNull && disjoint)
+            return IntervalValue(1, 1);
+        return IntervalValue(0, 1);
+    case CmpStmt::ICMP_UGT:
+    case CmpStmt::ICMP_SGT:
+    case CmpStmt::ICMP_ULT:
+    case CmpStmt::ICMP_SLT:
+        return sameValue ? IntervalValue(0, 0) : IntervalValue(0, 1);
+    case CmpStmt::ICMP_UGE:
+    case CmpStmt::ICMP_SGE:
+    case CmpStmt::ICMP_ULE:
+    case CmpStmt::ICMP_SLE:
+        return sameValue ? IntervalValue(1, 1) : IntervalValue(0, 1);
+    default:
+        return IntervalValue(0, 1);
+    }
+}
+
 bool AbstractInterpretation::isCmpBranchEdgeFeasible(const IntraCFGEdge* edge,
-        AbstractState& as)
+                                                     AbstractState& as)
 {
     const ICFGNode* pred = edge->getSrcNode();
     s64_t succ = edge->getSuccessorCondValue();
-    const CmpStmt* cmpStmt = SVFUtil::cast<CmpStmt>(
-                                 *edge->getCondition()->getInEdges().begin());
+    const CmpStmt* cmpStmt =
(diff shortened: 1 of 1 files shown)


bjjwwang:
### Formal semantics of conservative pointer comparison

For clarity, begin with how a points-to set represents one runtime pointer. The set contains alternative targets, not simultaneous targets:

- A known single target `{a}` means that the pointer's runtime target is exactly `a`.
- A multi-target set `{a, b}` means that the pointer may select either `a` or `b`.
- Therefore, `{a}` compared with `{a}` is definitely equal, whereas `{a, b}` compared with `{a, b}` may be equal or unequal because the two pointers may independently select different targets.
- A set containing `BlackHole` has no known target for this comparison and remains conservative.

Let `N(v)` normalize an abstract pointer operand into an address set:

$$
N(v) =
\begin{cases}
\{\mathit{NULL}\} & \text{if } v \text{ denotes a null pointer}, \\
A_v & \text{if } v \text{ carries a known abstract address set } A_v, \\
\{\mathit{BlackHole}\} & \text{otherwise.}
\end{cases}
$$

`BlackHole` is SVF's existing sentinel for a statically unresolved pointer target. In the pointer-comparison semantics below, any normalized set containing `BlackHole` is conservatively interpreted as the top of the address domain and may denote any concrete address. This scope does not claim that every existing `AddressValue` operation implements `BlackHole` as an absorbing lattice-top element. Define the comparison concretization for a normalized set $A$ as:

$$
\gamma_{\mathrm{cmp}}(A) =
\begin{cases}
\mathit{Addr} & \text{if } \mathit{BlackHole} \in A, \\
A & \text{otherwise.}
\end{cases}
$$

For equality, the set of possible concrete outcomes is:

$$
O_{eq}(A,B) = \{\,[a=b] \mid a \in \gamma_{\mathrm{cmp}}(A),\ b \in \gamma_{\mathrm{cmp}}(B)\,\}.
$$

We abstract Boolean outcomes as an interval:

$$
\alpha_B(O) = [\min O,\max O],
$$

where `[1,1]` means definitely true, `[0,0]` means definitely false, and `[0,1]` means that both outcomes remain possible.

The same semantics can be implemented directly with two predicates:

```text
hasUnknown(A, B) = BlackHole in A or BlackHole in B

mustEqual(A, B) =
    not hasUnknown(A, B) and |A| = 1 and A = B

mayEqual(A, B) =
    hasUnknown(A, B) or (A intersect B is not empty)

abstract(A == B) = [mustEqual(A, B), mayEqual(A, B)]
abstract(A != B) = [not mayEqual(A, B), not mustEqual(A, B)]
```

This gives the expected results without separate control-flow cases:

| Operands | `==` | `!=` |
|---|---:|---:|
| same known single target | `[1,1]` | `[0,0]` |
| known disjoint sets | `[0,0]` | `[1,1]` |
| overlapping sets with multiple possible targets | `[0,1]` | `[0,1]` |
| either operand unknown | `[0,1]` | `[0,1]` |

Pointer ordering predicates conservatively produce `[0,1]`. SVF object identifiers denote abstract objects, not ordered runtime addresses, so their numeric order cannot justify pruning a concrete program path.

Finally, branch feasibility should consume only the abstract comparison result $R$. For a branch edge labelled by $b \in \{0,1\}$:

$$
\mathit{feasible}(edge_b) \iff R \sqcap [b,b] \neq \bot.
$$

Thus, an edge is removed only when every concrete pair represented by the operands produces the opposite outcome. Null and pointer comparisons do not need additional special cases in branch feasibility.

Issue #1819, opened 2026-05-06
a problem with std::function


hiruwaKS:
Hello, this maybe a bug in implementation.

**demo:**
```cpp
#include <functional>
#include <map>

using namespace std;

void target() {}

void middleFunction(std::function<void()> cb) { cb(); }

int main() {
    std::map<int, std::function<void()>> callbacks={{1, target}};
    middleFunction(target); // resolved
    callbacks[1](); // resolved
    middleFunction(callbacks[1]); // can't resolve...
    return 0;
}
```
**Params:**
```bash
// LLVM14
./Release-build/bin/wpa -ander -ff-eq-base -dump-callgraph
./Release-build/bin/wpa -ander -ff-eq-base -svf-main -dump-callgraph
./Release-build/bin/wpa -ander -merge-pwc -vt-in-ir -ff-eq-base -svf-main -dump-callgraph
./Release-build/bin/wpa -ander -model-consts -model-arrays -merge-pwc -vt-in-ir -ff-eq-base -svf-main -dump-callgraph
./Release-build/bin/wpa -fspta -ff-eq-base -dump-callgraph
./Release-build/bin/wpa -fspta -ff-eq-base -fspta -ff-eq-base -svf-main -dump-callgraph
./Release-build/bin/wpa -fspta -merge-pwc -vt-in-ir -ff-eq-base  -svf-main -dump-callgraph
./Release-build/bin/wpa -fspta -model-consts -model-arrays -fspta  -merge-pwc -vt-in-ir -ff-eq-base -svf-main -dump-callgraph
```

All the params mentioned aboved fails to resolve the indirect call mentioned in demo, which feels odd.

yuleisui:
thanks for reporting this. I don't think SVF is able to treat C++ map as an array fetch here. But It could also be STL or  name mangling issue. Would you help dig this up?

hiruwaKS:
The root cause is small buffer optimization when `std::function` is in `std::map`. Since `std::function` is quite commonly used, I think it's worth paying attention to this problem.

A problem I discovered is in `GepValVar::isPointer` and `GepObjVar::isPointer`. I initially used libstdc++. In libstdc++13, `std::_Rb_tree_node` stores its value in an `unsigned char` array to implement SBO. After field-sensitive GEP expansion in SVF, `isPointer()` returns false, so the constraint propagation is incomplete.

[...]

PR #1884, opened 2026-08-26
Add option: field sensitive soundness for std::function problem

hiruwaKS:
This is the PR, but I don't really want it merged — it's just a trial
please see: https://github.com/SVF-tools/SVF/issues/1819
Thank you!


yuleisui:
> This is the PR, but I don't really want it merged — it's just a trial please see: #1819 Thank you!

Thank you.

PR #1887, opened 2026-09-09
Make AE pointer comparisons conservative

bjjwwang:
## Summary

A points-to set records alternative targets for one runtime pointer, not targets held simultaneously. A known single target such as `{a}` identifies the pointer's exact target. A multi-target set such as `{a, b}` means that either target may be selected at runtime. Consequently, two equal multi-target sets do not by themselves prove that the two concrete pointers are equal.

This PR makes AE pointer comparison conservative and predicate-aware:

- normalize pointer operands into known address sets, `{NULL}`, or `{BlackHole}`;
- produce definite `eq`/`ne` results only when known targets prove the outcome;
- keep overlapping multi-target sets, unresolved targets, and pointer ordering conservative;
- use the resulting Boolean interval directly for branch feasibility; and
- remove the duplicate pointer-comparison implementation that was unreachable behind the original address fast path.

## Problem

`AbstractInterpretation::updateStateOnCmp` previously handled every comparison between two address values as if it were equality. As a result, `icmp ne` on the same known target evaluated to true, while `icmp ne` on disjoint targets evaluated to false. The predicate-aware address switch later in the function never ran because the earlier address branch had already consumed those operands.

The old logic also treated equality of two abstract address sets as concrete pointer equality. This is valid for `{a}` compared with `{a}`, but not for `{a, b}` compared with `{a, b}`: the two runtime pointers may independently select different targets. Similarly, `BlackHole` is a statically unresolved target and must not be compared as if it were one concrete address.

Finally, pointer-ordering cases compared abstract object IDs, whose numeric order does not represent runtime address order. `isCmpBranchEdgeFeasible` also bypassed pointer and null comparisons, preventing a sound comparison result from controlling branch reachability.

## Formal semantics

Let `N(i, v)` normalize operand `v`, whose SVF variable ID is `i`, into an abstract address set:

$$
N(i,v) =
\begin{cases}
\{\mathit{NULL}\} & \text{if } i=\mathit{NullPtr} \text{ or } v=[0,0], \\
A_v & \text{if } v \text{ carries a known address set } A_v, \\
\{\mathit{BlackHole}\} & \text{otherwise.}
\end{cases}
$$

`BlackHole` is SVF's existing sentinel for a statically unresolved pointer target. For pointer comparison only, a set containing `BlackHole` is conservatively interpreted as the top of the address domain:

$$
\gamma_{\mathrm{cmp}}(A) =
\begin{cases}
\mathit{Addr} & \text{if } \mathit{BlackHole} \in A, \\
A & \text{otherwise.}
\end{cases}
$$

This scoped definition does not claim that every existing `AddressValue` operation implements `BlackHole` as an absorbing lattice-top element.

For equality, the possible concrete outcomes are:

$$
O_{eq}(A,B) =
\{\,[a=b] \mid
a \in \gamma_{\mathrm{cmp}}(A),\,
b \in \gamma_{\mathrm{cmp}}(B)\,\}.
$$

The Boolean outcomes are abstracted to an interval:

$$
\alpha_B(O)=[\min O,\max O],
$$

where `[1,1]` is definitely true, `[0,0]` is definitely false, and `[0,1]` retains both outcomes.

Equivalently, the implementation uses:

```text
hasUnknown(A, B) =
    BlackHole in A or BlackHole in B

mustEqual(A, B) =
    not hasUnknown(A, B)
    and |A| = 1
    and A = B

mayEqual(A, B) =
    hasUnknown(A, B)
    or (A intersect B is not empty)

abstract(A == B) = [mustEqual(A, B), mayEqual(A, B)]
abstract(A != B) = [not mayEqual(A, B), not mustEqual(A, B)]
```

This yields:

| Left targets | Right targets | `==` | `!=` |
|---|---|---:|---:|
| `{a}` | `{a}` | `[1,1]` | `[0,0]` |
| `{a}` | `{b}`, where `a != b` | `[0,0]` | `[1,1]` |
| `{a, b}` | `{c}`, where the sets are disjoint | `[0,0]` | `[1,1]` |
| `{a, b}` | `{a}` | `[0,1]` | `[0,1]` |
| `{a, b}` | `{a, b}` | `[0,1]` | `[0,1]` |
| a set containing `BlackHole` | any set | `[0,1]` | `[0,1]` |

`NULL` is a known distinguished target, so two null operands follow the same-known-target case.

Pointer-ordering predicates conservatively produce `[0,1]`. SVF object IDs identify abstract objects; their numeric order cannot justify pruning a concrete program path.

For a branch edge labelled by $b \in \{0,1\}$ and comparison result $R$:

$$
\mathit{feasible}(edge_b)
\iff R \sqcap [b,b] \neq \bot.
$$

Thus, an edge is removed only when every concrete pointer pair represented by the operands produces the opposite outcome. Branch feasibility does not need separate pointer or null cases.

## Implementation

- `normalizePointerAddresses` gives null, known addresses, and unresolved operands one address-set representation.
- `comparePointerValues` implements the cases above as an explicit decision tree.
- `updateStateOnCmp` dispatches pointer operands to that helper and keeps the existing interval comparison for non-pointer operands.
- `isCmpBranchEdgeFeasible` consumes the stored comparison interval and remains conservative when no interval result is available.

## Validation

- The `ae` target builds successfully with LLVM 21.1.8 and Z3 4.16.0.
- Focused pointer/null AE regressions: 28/28 passed across dense, semi-sparse, full-sparse, and no-main modes.
- Targeted tests verify definite results for the same and disjoint known targets.
- An overlapping-target test confirms that both feasible branches are retained.

### DIFF
--- a/svf/lib/AE/Svfexe/AbsExtAPI.cpp
+++ b/svf/lib/AE/Svfexe/AbsExtAPI.cpp
@@ -235,13 +235,9 @@ void AbsExtAPI::initExtFunMap()
         const AbstractValue& ptrVal = ae->getAbsValue(callNode->getArgument(0), callNode);
         for (auto addr: ptrVal.getAddrs())
         {
-            if (AbstractState::isBlackHoleObjAddr(addr))
-            {
-            }
-            else
-            {
+            if (!AbstractState::isBlackHoleObjAddr(addr) &&
+                    !AbstractState::isNullMem(addr))
                 as.addToFreedAddrs(addr);
-            }
         }
     };
     // Add all free-related functions to func_map
@@ -455,6 +451,9 @@ IntervalValue AbsExtAPI::getStrlen(const ValVar *strValue, const ICFGNode* node)
     const AbstractValue& ptrVal = ae->getAbsValue(strValue, node);
     for (const auto& addr : ptrVal.getAddrs())
     {
+        // Null has no backing buffer whose size can be inspected.
+        if (AbstractState::isNullMem(addr))
+            continue;
         NodeID objId = as.getIDFromAddr(addr);
         if (svfir->getBaseObject(objId)->isConstantByteSize())
         {
--- a/svf/lib/AE/Svfexe/AbstractInterpretation.cpp
+++ b/svf/lib/AE/Svfexe/AbstractInterpretation.cpp
@@ -495,23 +495,12 @@ bool AbstractInterpretation::isCmpBranchEdgeFeasible(const IntraCFGEdge* edge,
     s64_t succ = edge->getSuccessorCondValue();
     const CmpStmt* cmpStmt = SVFUtil::cast<CmpStmt>(
                                  *edge->getCondition()->getInEdges().begin());
-
-    if (cmpStmt->getOpVarID(0) == IRGraph::NullPtr ||
-            cmpStmt->getOpVarID(1) == IRGraph::NullPtr)
-        return true;
-
-    AbstractValue opVal[2] =
-    {
-        getAbsValue(cmpStmt->getOpVar(0), pred),
-        getAbsValue(cmpStmt->getOpVar(1), pred)
-    };
-
-    const bool hasIntervalCmp = opVal[0].isInterval() && opVal[1].isInterval();
-    if (!hasIntervalCmp && (opVal[0].isAddr() || opVal[1].isAddr()))
+    const AbstractValue& cmpValue = getAbsValue(cmpStmt->getRes(), pred);
+    if (!cmpValue.isInterval())
         return true;
 
     // Feasibility check: cmp result must be compatible with branch successor
-    IntervalValue resVal = getAbsValue(cmpStmt->getRes(), pred).getInterval();
+    IntervalValue resVal = cmpValue.getInterval();
     resVal.meet_with(IntervalValue((s64_t)succ, succ));
     if (resVal.isBottom())
         return false;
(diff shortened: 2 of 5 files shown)


svf/lib/AE/Svfexe/AbstractInterpretation.cpp
```diff
@@ -1138,229 +1128,136 @@ void AbstractInterpretation::updateStateOnBinary(const BinaryOPStmt *binary)
...
-        }
-        updateAbsValue(cmp->getRes(), resVal, node);
-    }
-    // if op0 or op1 is nullptr, compare abstractValue instead of touching addr or interval
-    else if (op0 == IRGraph::NullPtr || op1 == IRGraph::NullPtr)
+    if (cmp->getOpVar(0)->getType()->isPointerTy())
     {
-        IntervalValue resVal = (op0Val.equals(op1Val)) ? IntervalValue(1, 1) : IntervalValue(0, 0);
-        updateAbsValue(cmp->getRes(), resVal, node);
+        updateAbsValue(cmp->getRes(), comparePointerValues(cmp, op0Val, op1Val),
```
yuleisui, line 1199: Could you explain why we need to compare pointer values here?

svf/lib/AE/Svfexe/AbstractInterpretation.cpp
```diff
@@ -1138,229 +1128,136 @@ void AbstractInterpretation::updateStateOnBinary(const BinaryOPStmt *binary)
...
+                       node);
+        return;
     }
-    else
+
+    if (!op0Val.isInterval() || !op1Val.isInterval())
+        return;
```
yuleisui, line 1205: There are a number of places you used if(...isInterval) return. This is a bit ad hoc and could cause maintenance issues. Could simplify/restructure and avoid such conditional returns in the middle of the code.

Please search all such places and fix them.

svf/lib/AE/Svfexe/AbstractStateManager.cpp
```diff
@@ -324,6 +324,9 @@ AddressValue AbstractInterpretation::getGepObjAddrs(const ValVar* pointer, Inter
         const AbstractValue& addrs = getAbsValue(pointer, node);
         for (const auto& addr : addrs.getAddrs())
         {
+            // Null has no object from which a field address can be derived.
+            if (AbstractState::isNullMem(addr))
```
yuleisui, line 328: There are a number of places you used if(...isInterval) return. This is a bit ad hoc and could cause maintenance issues. Could simplify/restructure and avoid such conditional return in the middle of the code.

Please search all such places (either fix or replace with assertions where necessary).

svf/lib/AE/Svfexe/AbsExtAPI.cpp
```diff
@@ -455,6 +449,10 @@ IntervalValue AbsExtAPI::getStrlen(const ValVar *strValue, const ICFGNode* node)
     const AbstractValue& ptrVal = ae->getAbsValue(strValue, node);
     for (const auto& addr : ptrVal.getAddrs())
     {
+        // Unknown and null pointers have no concrete backing buffer to inspect.
+        if (AbstractState::isBlackHoleObjAddr(addr) ||
```
bjjwwang: blkhole 和 nullptr 
一起 包起来

svf/lib/AE/Svfexe/AbstractStateManager.cpp
yuleisui: Could you do a global search of isNullMem if you want all of them to be replaced with `isNullOrBlackHoleAddr`?

svf/lib/AE/Svfexe/AbstractStateManager.cpp
```diff
@@ -341,6 +347,9 @@ AbstractValue AbstractInterpretation::loadValue(const ValVar* pointer, const ICF
     AbstractValue res;
     for (auto addr : ptrVal.getAddrs())
     {
+        // Null has no memory object from which a value can be loaded.
+        if (AbstractState::isNullMem(addr))
```
yuleisui, line 351: Same here

svf/lib/AE/Svfexe/AbstractStateManager.cpp
```diff
@@ -352,7 +361,12 @@ void AbstractInterpretation::storeValue(const ValVar* pointer, const AbstractVal
     const AbstractValue& ptrVal = getAbsValue(pointer, node);
     AbstractState& as = getAbsState(node);
     for (auto addr : ptrVal.getAddrs())
+    {
+        // Null has no memory object that can be updated.
+        if (AbstractState::isNullMem(addr))
```
yuleisui, line 366: Same here and better do a global search under the AE folder.

bjjwwang: Done.

PR #1888, opened 2026-09-13
Add getMayAliases interface with Andersen implementation

cjsrxzdyzds:
### What

Clients asking for every node that may alias one pointer currently have to call `mayAlias` on every node. This adds `virtual NodeBS PointerAnalysis::getMayAliases(NodeID)` and an Andersen implementation returning exactly the live SVFIR nodes `q` for which `mayAlias(node, q)` holds.

Other analyses can implement the interface later. The base default prints an unsupported diagnostic and calls `std::abort()`, including with assertions disabled. Andersen's implementation is callable through `PointerAnalysis*`.

### How

- Andersen collects candidates from reverse points-to sets for the bases and fields of the query's expanded objects, plus the black hole. It expands reverse keys to their SCC members and confirms live candidates with `mayAlias`.
- It falls back to every live node when reverse sets are unavailable, the query may point to the black hole, or the reverse-entry count exceeds the SVFIR node count. This threshold is a heuristic, not a wall-clock guarantee.
- The query adds no work to solving. The in-tree caller is Andersen's existing alias-test validation, which now calls through the base interface and compares against exhaustive `mayAlias` enumeration.

Two supporting fixes preserve the other SCC members when `cleanConsCG` removes one node, and expose the base `mayAlias` overloads hidden by `BVDataPTAImpl`. The existing `spec-gap.c` alias test catches a regression of the SCC fix.

### Validation

- Latest fetched, unmodified Test-Suite master (`b459ab3d`): **2,575/2,575** tests pass, including **454** interface checks in 114 Andersen alias-test programs. An empty-result mutation makes `spec-gap.c` fail. No Test-Suite changes.
- GCC 11 Release build with assertions and `-Werror`; Clang 14/21 checks of the affected implementations: no warnings or errors.
- Steensgaard, CFLAlias and FlowDDA reject unsupported calls as documented. This also passes with the actual base implementation compiled under `NDEBUG`.
- After the interface change, all nodes of `spec-gap.c` and `unordered_map-1.cpp` agree under nine Andersen configurations: **26,137 queries, zero disagreements**.
- Earlier full-node sweeps over 254 programs found no disagreements in completed runs. `type-while.c` aborts upstream too under `-field-limit=0`; 54 programs hit existing upstream alias-test assertions under each of `-sander` and `-sfrander`.
- Earlier corpus measurements covered 20 C/C++ modules, 10 Rust modules from 7 crates, and three larger modules. Sampled queries agreed with exhaustive enumeration; timings and limitations are in [the measurement report](https://github.com/SVF-tools/SVF/pull/1888#issuecomment-5657932851). These performance measurements were not rerun for the interface-only change.

AI assistance: Claude Code and Codex.


yuleisui:
Thanks for the contributions. Could you clarify the ones below?

1. Could this be applied to all points-to analysis rather than Andersen's alone? If so, please update your patch.
2. As collecting all aliases may be time-consuming, would this patch incur significant time?
3. For the `cleanConsCG` [fix](https://github.com/SVF-tools/SVF/pull/1888/changes#diff-3d3eabe579bc4d70d56def66183fbedf445526ce8b313bace1cdea0144c69e89L184), may I know whether you could supply any test cases?

cjsrxzdyzds:
Thanks for looking at this. Answers below.

### 1. Other points-to analyses

It can be offered for every points-to analysis, with three limits:

- The fast path needs reverse points-to sets and Andersen's SCC representatives. Steensgaard, FlowSensitive and VersionedFlowSensitive keep no reverse sets, so for them `getMayAliases` has to call `mayAlias` on every SVFIR node.
- `CFLAlias::alias` looks both nodes up in the CFL graph, and `getGNode` asserts "Node not found!" for SVFIR nodes the graph does not hold. A loop over SVFIR nodes therefore aborts on CFL analyses.
- `FlowDDA` and `ContextDDA` compute points-to sets on demand; `DDAPass::alias` computes both nodes before comparing them. On the analysis object, a loop over nodes only sees the sets computed so far. An exact answer would need a demand query for every pointer.

The update I would make:

- `PointerAnalysis` gets a virtual `getMayAliases` that calls `mayAlias` on every SVFIR node.
- `Andersen` keeps the fast path and falls back to the base version. Its alias-test check compares the two, so CI covers both.
- `CFLAlias` overrides it to try only the nodes in its graph.
- For `FlowDDA` and `ContextDDA`, the comment says the result covers the points-to sets computed so far.

If this matches what you had in mind, I will update the patch.

### 2. Time

The patch adds no time unless a client calls `getMayAliases`, because it does not change solving. In SVF, the only caller is the new check in Andersen's alias tests.

In CI, `wpa -ander` on the 114 programs of `basic_c_tests` and `basic_cpp_tests` took 17.38 s on master and 17.77 s with this PR, averaged over three runs each, about 2% more. That run checks 454 pointers, and each check also calls `mayAlias` on every node.

One call on real modules under `-nander`, against calling `mayAlias` on every node. The two sample rows give the median over modules of the mean time per call; the other rows are means over 80 calls:

| Modules | SVFIR nodes | `getMayAliases`, per call | `mayAlias` on every node, per call |
|---|---|---|---|
| 20 C/C++ modules from the LLVM test-suite | median 8,520 | 0.021 ms | 2.1 ms |
| 10 Rust modules from 7 crates | median 55,986 | 0.21 ms | 35 ms |
| sqlite3 | 137,933 | 1.08 s | 1.26 s |
| tramp3d-v4 | 297,782 | 12.8 s | 17.0 s |
| redb (Rust) | 410,492 | 0.10 s | 0.46 s |

So a call can be expensive on large programs: 12.8 s on tramp3d-v4, whose whole Andersen solve takes 29.8 s. In these measurements it was never slower than calling `mayAlias` on every node, because the fast path is skipped when the reverse sets to walk hold more entries than there are nodes. For the analyses without the fast path, a call is the loop in the last column.

### 3. Tests for the `cleanConsCG` fix

`basic_c_tests/spec-gap.c`, which `wpa_tests` already runs, covers it.

- With line 184 restored and the rest of this PR kept, the new `getMayAliases` check fails on `spec-gap.c` and its assertion aborts `wpa`. CI builds with `SVF_ENABLE_ASSERTIONS`, so `wpa_tests` fails. No other program of the 114 fails. So CI catches a regression of the fix through this PR, with no Test-Suite change.
- The existing alias tests could not catch it. `cleanConsCG` runs in `normalizePointsTo`, after solving. After that point, only the cycle statistics read the SCC member sets, so the bug changed statistics but no alias result. `getMayAliases` is the first code that needs the member sets to be right.
- The bug is visible on master with `wpa -ander -stat`:

| Program | master: TotalCycleNum / NodesInCycles / MaxNodesInSCC | this PR |
|---|---|---|
| `basic_c_tests/spec-gap.c` | 3 / 9 / 3 | 4 / 17 / 8 |
| `basic_cpp_tests/unordered_map-1.cpp` | 1 / 13 / 13 | 2 / 32 / 19 |

svf/include/MemoryModel/AbstractPointsToDS.h
```diff
@@ -101,6 +101,12 @@ class PTData
         return ptdTy;
     }
 
+    /// Whether reverse points-to sets were requested when this was constructed.
```
mbarbar:
```suggestion
    /// Whether we maintain reverse points-to sets.

File: svf/include/WPA/Andersen.h [hunk 2]
```diff
@@ -252,6 +252,9 @@ class Andersen:  public AndersenBase
         return getPTDataTy()->unionPts(id,ptd);
     }
 
+    /// Collect exactly the SVFIR nodes q for which mayAlias(node, q) holds, clearing aliases first
+    void getMayAliases(NodeID node, NodeBS& aliases);
```
mbarbar: I would much prefer this to construct and return the set of aliases rather than pass in/out as reference, so like
```
NodeBS getMayAliases(NodeID node);
```

(I know this style of interface is present in some places, but I think we should move toward more functional interfaces. The extra references and mutations become hard to reason about sometimes and there should be no performance impact in this case.)

svf/include/WPA/Andersen.h
```diff
@@ -354,6 +357,12 @@ class Andersen:  public AndersenBase
     /// SCC detection
     virtual NodeStack& SCCDetect();
 
+    /// Collect may-alias candidates from reverse points-to sets, if cheaper than trying every node
+    bool getMayAliasCandidates(const PointsTo& expandedPts, NodeBS& candidates);
```
mbarbar: Likewise, I would prefer this to construct and return the candidates set (and not return a bool at all).

---

This function I believe does not have a reason to simply short circuit and leave the long path to the caller as the only caller tries this shortcut, and if it fails, takes the long path. Maybe one day we will have a caller that cannot afford the long path but today is not that day, we can think about it when the time comes.

So this function should just do everything. It should try the shortcut, and if that fails, do it the long way. The docstring can describe this if you so wish, so something like.
```
Collect may-alias candidates. 
Attempts to do so from reverse points-to sets (cheaper), and if not possible, 
falls back to trying every node (costlier).
```

---

I also believe it better if this function took a NodeID and got the expanded points-to set itself rather than being passed said points-to set as an argument. It would be easier to understand for users and it would make implementation simpler (that big comment about expanded FI objects would be unnecessary).

svf/lib/WPA/Andersen.cpp
```diff
@@ -914,6 +917,115 @@ void Andersen::updateNodeRepAndSubs(NodeID nodeId, NodeID newRepId)
...
+    expandFIObjs(getPts(node), expandedPts);
+
+    NodeBS candidates;
+    if (!getMayAliasCandidates(expandedPts, candidates))
+    {
+        for (SVFIR::iterator it = pag->begin(), eit = pag->end(); it != eit; ++it)
+            candidates.set(it->first);
+    }
+
+    // Field nodes removed by normalizePointsTo keep stale points-to entries
```
mbarbar: Is this comment for the `hasGNode` check? If so, can you move it down and clarify that, e.g. "Check candidate still exists because field nodes removed by ..."

mbarbar (commented):
Looks quite good to me however I'd like to see the interfaces changed as I pointed out. I think it will be easier to use/read and also make the implementation simpler.

svf/lib/WPA/Andersen.cpp
```diff
@@ -181,7 +181,10 @@ void AndersenBase:: solveAndwritePtsToFile(const std::string& filename)
 
 void AndersenBase::cleanConsCG(NodeID id)
 {
-    consCG->resetSubs(consCG->getRep(id));
+    // Remove only this node from its representative's members. Erasing the whole
+    // member set would leave every other member mapped to a representative that no
+    // longer lists it, so sccSubNodes would stop being the inverse of sccRepNode.
+    consCG->getSubs(consCG->sccRepNode(id)).reset(id);
```
mbarbar, line 187: Why did the rep node retrieval change from `getRep` to `sccRepNode`?

svf/include/MemoryModel/AbstractPointsToDS.h
cjsrxzdyzds: Applied.

svf/include/WPA/Andersen.h
cjsrxzdyzds: Done: `NodeBS getMayAliases(NodeID node)`. The alias-test check now compares its return value.

svf/include/WPA/Andersen.h
cjsrxzdyzds: Done. `NodeBS getMayAliasCandidates(NodeID node)` expands the node's points-to set itself, tries the reverse points-to sets, and otherwise returns every node, with your docstring. `getMayAliases` now only confirms the candidates with `mayAlias`. I kept a shorter comment on why the bases and all their fields are collected, since that is what makes the candidate set complete.

svf/lib/WPA/Andersen.cpp
cjsrxzdyzds: Yes, it was for the `hasGNode` check. Moved there and reworded as you suggested.

svf/lib/WPA/Andersen.cpp
cjsrxzdyzds: On this path both return the same node. `cleanConsCG` is only called from `normalizePointsTo`, for GEP nodes that `collapseField` has already merged into their base, and merging gives each of them an entry in `nodeToRepMap`. I checked this with a build that logs every call: over 2,400 `wpa` runs (the 254 Test-Suite programs under nine solver and flag settings, plus `wpa -ander` with alias tests on the 114 programs CI checks), `cleanConsCG` was called 387 times, and every node had an entry.

I used `sccRepNode` because it only reads the map. `getRep` is `nodeToRepMap[node]`, which for a node without an entry would insert one and return node 0. `sccRepNode` is also what Andersen uses elsewhere to find a representative, for example in `unionPts`. So the switch is defensive rather than a fix, and I can restore `getRep` if you prefer the smaller diff.

cjsrxzdyzds:
Thanks for the review. 6b0ce8e3 makes the changes: `getMayAliases(NodeID)` and `getMayAliasCandidates(NodeID)` return their sets, `getMayAliasCandidates` expands the points-to set and falls back to every node itself, the stale-entry comment sits at the `hasGNode` check, and `hasReversePts` uses your wording. Results are unchanged: ctest against Test-Suite master passes 2,575 of 2,575 tests, including the 454 `getMayAliases` checks in `wpa_tests`.

svf/lib/WPA/Andersen.cpp
mbarbar: Makes sense, thanks for the explanation.

mbarbar:
Thanks for updating. LGTM! Happy for this to be merged when @yuleisui is happy.

yuleisui:
Thanks @cjsrxzdyzds The patch looks quite good now. I would like `getMayAliases` to be an interface in the `PointerAnalysis` class so that every subclass, not just Andersen's analysis, can implement and use it later, although we currently only use it for Andersen's validation.

cjsrxzdyzds:
Thanks @yuleisui. b6a76312 adds `virtual NodeBS getMayAliases(NodeID)` to `PointerAnalysis`, with Andersen providing the current implementation. The base default reports unsupported and calls `std::abort()`, so other analyses can override the interface later and unsupported calls fail explicitly even with assertions disabled.

Andersen's existing alias-test validator now calls through `PointerAnalysis*` and compares against exhaustive `mayAlias` enumeration. Candidate selection is unchanged.

Validation: all 2,575 tests pass against the latest unmodified Test-Suite master (`b459ab3d`), including 454 interface checks. GCC and Clang 14/21 checks pass; unsupported-default probes also pass with the base implementation compiled under `NDEBUG`. Full-node checks of two regression programs across nine Andersen configurations agree on 26,137 queries. An empty-result mutation is caught by the existing `spec-gap.c` check.

I also updated the PR description to reflect the interface and current signature, and corrected the earlier corpus and sweep qualifications.

PR #1892, opened 2026-09-17
fix(svf-llvm): guard getFlattenedElemIdx against opaque-pointer type mismatch

WizardBornov:
## What

`IRGraph::getFlattenedElemIdx()` asserts that its `SVFType` argument is a struct or array type whenever `Options::ModelArrays()` is disabled. Under LLVM's opaque pointers, the `SVFType` resolved for a GEP base in `collectMemcpyFields()` (SVFIRExtAPI.cpp) can fold to something other than a struct/array `SVFType`, even though the corresponding LLVM aggregate genuinely is a `StructType`/`ArrayType`. That mismatch trips the assert on real, opaque-pointer-compiled code.

## Why this shape of fix

This looks like the same class of issue already fixed in `ObjTypeInference.cpp` (c5083f6b, e83f846a) — opaque pointers letting a GEP base fold to a value whose `SVFType` doesn't match what the surrounding code expects. I followed the same pattern here: guard the call and fall back to a safe default (field/element 0) instead of aborting, rather than trying to correctly re-derive the type. `collectMemcpyFields` already has an explicit best-effort trade-off in this area (see the `MaxFieldLimit` comment just above the new helper), so a degraded offset on this path is consistent with existing behavior, not a new soundness hole.

## Context / caveat

Ran into this integrating SVF against an LLVM 21 `-O0` build for a call-graph research project (PASS Lab, IIT Guwahati). I have **not** compiled this exact patch against a live build in this checkout — my LLVM 21 toolchain is mid-setup on a fresh machine right now. Opening as a **draft** for that reason; happy to convert to ready for review once I've verified it compiles and reproduces the fix against the original crash case, or sooner if a maintainer wants to just review the diff on its merits.


svf-llvm/lib/SVFIRExtAPI.cpp
```diff
@@ -88,6 +88,26 @@ bool hasNondetArgStoreAtExtCall(const CallICFGNode* callICFGNode)
...
+/// back to field/element 0 rather than aborting: memcpy field flattening is
+/// already a best-effort external-memory approximation (see the
+/// MaxFieldLimit comment above), so a degraded offset here is consistent
+/// with the existing soundness trade-off rather than a new one.
+APOffset safeFlattenedElemIdx(IRGraph* pag, const SVFType* svfType, u32_t idx)
+{
+    if (!Options::ModelArrays() &&
+            !SVFUtil::isa<SVFStructType>(svfType) &&
+            !SVFUtil::isa<SVFArrayType>(svfType))
+        return 0;
```
yuleisui: Could we move these four lines to be inlined within `getFlattenedElemIdx`?

WizardBornov:
Good catch, moved into getFlattenedElemIdx() directly in 4523a677 so it protects every caller, not just collectMemcpyFields(). Also dropped the now-redundant assert in the array branch since the guard above already proves it by that point.

svf/lib/Graphs/IRGraph.cpp
```diff
@@ -147,6 +147,15 @@ const SVFType *IRGraph::getOriginalElemType(const SVFType *baseType, u32_t origI
...
 u32_t IRGraph::getFlattenedElemIdx(const SVFType *T, u32_t origId)
 {
+    // Under LLVM's opaque pointers, the SVFType resolved for a GEP base can fold to
+    // something other than a struct/array SVFType even though the corresponding LLVM
+    // aggregate genuinely is a StructType/ArrayType (seen integrating SVF against an
+    // LLVM 21 -O0 build). Fall back to element 0 instead of hitting the assert below.
+    if (!Options::ModelArrays() &&
+            !SVFUtil::isa<SVFStructType>(T) &&
+            !SVFUtil::isa<SVFArrayType>(T))
+        return 0;
```
yuleisui: These few lines should be put under the else branch between lines 165 and 176.

svf/lib/Graphs/IRGraph.cpp
```diff
@@ -164,7 +173,6 @@ u32_t IRGraph::getFlattenedElemIdx(const SVFType *T, u32_t origId)
         else
         {
             /// When Options::ModelArrays is disabled, any element index Array is modeled as the base
-            assert(SVFUtil::isa<SVFArrayType>(T) && "Only accept struct or array type if Options::ModelArrays is disabled!");
```
yuleisui, line 167: Would be good to revise this nested `else` branch:

```
 if(Options::ModelArrays())
{
}
else
{
  if(SVFUtil::isa<SVFStructType>(T))
  {
  }
  else if(SVFUtil::isa<SVFArrayType>(T))
  {
  }
  else // please indicate what other types will fall into this else branch!
  {
  }
}

``

**Comment by WizardBornov** [comment]
Both addressed in 7280d00a: moved the guard entirely inside the !ModelArrays() branch, and split struct/array/other into an explicit if/else if/else. The 'other' case is the opaque-pointer GEP-base-folding scenario, now named and commented at the point it's handled instead of implicit.

File: svf/lib/Graphs/IRGraph.cpp [hunk 4]
```diff
@@ -161,10 +161,17 @@ u32_t IRGraph::getFlattenedElemIdx(const SVFType *T, u32_t origId)
...
+            return 0;
+        }
+        else
+        {
+            // Under LLVM's opaque pointers, the SVFType resolved for a GEP base can fold to
+            // something other than a struct/array SVFType even though the corresponding LLVM
+            // aggregate genuinely is a StructType/ArrayType (seen integrating SVF against an
+            // LLVM 21 -O0 build). Fall back to element 0 instead of asserting.
```
yuleisui, line 174: Could you also provide an example of the SVFType other than Array/Struct here?

yuleisui: Ideally, we shall add an assertion here to make sure only certain type(s) will fall into this `else` branch, so it is under control.

WizardBornov:
Fair ask, and I want to be straight with you rather than guess. I don't have a captured example from a fresh reproduction, but I do have a real one from when I originally hit this: a full backtrace from an actual crash on real libavcodec code (LLVM 21, true -O0 -g, opaque pointers), before I'd written any fix.

Assertion:
```
IRGraph.cpp:163: Assertion `SVFUtil::isa<SVFArrayType>(T) && "Only accept struct or array type if Options::ModelArrays is disabled!"' failed.
```

Call chain (from gdb, Release build so no locals available at the crash frame, but the chain itself is exact):
```
main -> SVFIRBuilder::build() -> handleExtCall() -> addComplexConsForExt()
     -> getMemcpyFields() -> collectMemcpyFields() [recursive] -> IRGraph::getFlattenedElemIdx()
```

So it's `addComplexConsForExt` handling a real memcpy-family external call on real libavcodec code, which matches the reasoning in the patch: a GEP base's SVFType folding to something other than struct/array under opaque pointers, similar to the ObjTypeInference/ConstantData issue already fixed.

I tried to re-derive the concrete type by rebuilding with debug symbols and re-triggering it just now, but libavcodec-scale whole-program analysis needs ~11GB+ RSS and I couldn't get a clean run through in the environment I have right now (this matches your own project's experience too, for what it's worth -- I found a note in an earlier session here mentioning libavcodec-scoped SVF runs needing close to that much memory with very little headroom). I don't want to guess at the type name to fill that gap.

If the mechanism above is convincing on its own, I'm happy to leave the current fallback-to-0 as is. If you'd rather hold for a fully captured repro with the concrete type, I can keep trying to get a clean run through and follow up when I have it -- just wanted to give you an honest state of things rather than sit on it.

yuleisui:
> Fair ask, and I want to be straight with you rather than guess. I don't have a captured example from a fresh reproduction, but I do have a real one from when I originally hit this: a full backtrace from an actual crash on real libavcodec code (LLVM 21, true -O0 -g, opaque pointers), before I'd written any fix.
> 
> Assertion:
> 
> ```
> IRGraph.cpp:163: Assertion `SVFUtil::isa<SVFArrayType>(T) && "Only accept struct or array type if Options::ModelArrays is disabled!"' failed.
> ```
> 
> Call chain (from gdb, Release build so no locals available at the crash frame, but the chain itself is exact):
> 
> ```
> main -> SVFIRBuilder::build() -> handleExtCall() -> addComplexConsForExt()
>      -> getMemcpyFields() -> collectMemcpyFields() [recursive] -> IRGraph::getFlattenedElemIdx()
> ```
> 
> So it's `addComplexConsForExt` handling a real memcpy-family external call on real libavcodec code, which matches the reasoning in the patch: a GEP base's SVFType folding to something other than struct/array under opaque pointers, similar to the ObjTypeInference/ConstantData issue already fixed.
> 
> I tried to re-derive the concrete type by rebuilding with debug symbols and re-triggering it just now, but libavcodec-scale whole-program analysis needs ~11GB+ RSS and I couldn't get a clean run through in the environment I have right now (this matches your own project's experience too, for what it's worth -- I found a note in an earlier session here mentioning libavcodec-scoped SVF runs needing close to that much memory with very little headroom). I don't want to guess at the type name to fill that gap.
> 
> If the mechanism above is convincing on its own, I'm happy to leave the current fallback-to-0 as is. If you'd rather hold for a fully captured repro with the concrete type, I can keep trying to get a clean run through and follow up when I have it -- just wanted to give you an honest state of things rather than sit on it.

Thanks for getting back. I would suggest to add an assert so that we know which type would fail into this branch if the type is neither array or struct.

WizardBornov:
Added in c47aa39b: prints the concrete type via `SVFUtil::errs()` before falling back (so it's visible even in release builds where the assert compiles out), plus an `assert(false, ...)` to fail loudly if this fires under assertions-enabled builds. That way if it's ever hit again, we get the exact type instead of guessing.

svf/lib/Graphs/IRGraph.cpp
```diff
@@ -161,10 +161,23 @@ u32_t IRGraph::getFlattenedElemIdx(const SVFType *T, u32_t origId)
...
+            // Under LLVM's opaque pointers, the SVFType resolved for a GEP base can fold to
+            // something other than a struct/array SVFType even though the corresponding LLVM
+            // aggregate genuinely is a StructType/ArrayType (seen integrating SVF against an
+            // LLVM 21 -O0 build). Report the concrete type so a future hit of this branch is
+            // diagnosable rather than guessed at, then fall back to element 0 instead of
+            // crashing (the printed type survives into release builds even when the assert
+            // below is compiled out).
+            SVFUtil::errs() << "getFlattenedElemIdx: type is neither struct nor array: "
+                            << T->toString() << "\n";
+            assert(false && "getFlattenedElemIdx: type is neither struct nor array (see printed type above)");
```
yuleisui, line 180: Would your test case (opaque pointer type) cause the assertion failure here?

WizardBornov:
Yes — in any assertions-enabled build (no `NDEBUG`), hitting this branch still trips `assert(false)` immediately, same as before this commit. The `SVFUtil::errs()` print happens right before it, so the concrete type is visible in the crash output rather than having to be guessed at. It's only in an NDEBUG/release-style build that the assert compiles out and it falls through to the `return 0` fallback instead.

So to be precise about what this commit does and doesn't do: it doesn't eliminate the crash universally, it makes it diagnosable where it still happens (assertions on) and silent-safe where it doesn't (assertions off). That matches what you asked for specifically, but flagging the distinction in case you'd rather it never abort regardless of build config — happy to drop the assert and keep just the print if so.

yuleisui:
> Good catch, moved into getFlattenedElemIdx() directly in [4523a67](https://github.com/SVF-tools/SVF/commit/4523a677c60efb73a69313db7cfceb01c394fe4d) so it protects every caller, not just collectMemcpyFields(). Also dropped the now-redundant assert in the array branch since the guard above already proves it by that point.

> Yes — in any assertions-enabled build (no `NDEBUG`), hitting this branch still trips `assert(false)` immediately, same as before this commit. The `SVFUtil::errs()` print happens right before it, so the concrete type is visible in the crash output rather than having to be guessed at. It's only in an NDEBUG/release-style build that the assert compiles out and it falls through to the `return 0` fallback instead.
> 
> So to be precise about what this commit does and doesn't do: it doesn't eliminate the crash universally, it makes it diagnosable where it still happens (assertions on) and silent-safe where it doesn't (assertions off). That matches what you asked for specifically, but flagging the distinction in case you'd rather it never abort regardless of build config — happy to drop the assert and keep just the print if so.

OK, just merged and `assert(false)` will catch the error, but if we want to pass your tests, it would be good to change it to `assert(type)`, where `type` is the one that triggered your assertion.

WizardBornov:
Want to make sure I implement the right thing rather than guess. Could you clarify what condition you'd want `assert(type)` to check?

The current `T` in that branch is already a non-null `SVFType*` (we dereference it via `T->toString()` right above), so a plain `assert(T)` wouldn't actually catch anything, it's always true by that point. If you mean something narrower, like asserting that `T` is specifically one of the concrete SVFType kinds you'd expect to legitimately land here (so an unrelated/unexpected type still trips loudly, but a known-benign one doesn't), I'm happy to add that, just want to confirm which kind(s) you have in mind before I do.

yuleisui:
(1) `assert(false)`, as it is now, then it will always make your case crash. 
(2) narrower the assert down to a certain type based on your case and still output the error (as a warning).

(1) is harsher but (2) looks that can solve your orignial issue

yuleisui:
> (1) `assert(false)`, as it is now, then it will always make your case crash. (2) narrower the assert down to a certain type based on your case and still output the error (as a warning).
> 
> (1) is harsher but (2) looks that can solve your orignial issue

You could try to submit another pull request based on (2).

WizardBornov:
Appreciate the concrete direction. Being straight about where I actually am: I'm short on both time and machine resources right now (the FSE deadline is eating most of both), and pinning down the exact type needs a real repro at scale, which is the part that's been blocked. I'll make a real attempt to nail down the concrete type by this weekend and open a follow-up PR based on (2) once I have it, rather than guess at it now.

yuleisui:
> Appreciate the concrete direction. Being straight about where I actually am: I'm short on both time and machine resources right now (the FSE deadline is eating most of both), and pinning down the exact type needs a real repro at scale, which is the part that's been blocked. I'll make a real attempt to nail down the concrete type by this weekend and open a follow-up PR based on (2) once I have it, rather than guess at it now.

Thank you.

PR #1893, opened 2026-09-19
Free the old backing when assigning to a PointsTo

cjsrxzdyzds:
## What is wrong

`PointsTo` stores its bit vector in a union and rebuilds it with placement new on assignment, because the stored type can change. Neither assignment operator ends the lifetime of the value already held, so whatever that value owned is lost:

```cpp
PointsTo &PointsTo::operator=(const PointsTo &rhs)
{
    if (this == &rhs)
        return *this;
    this->type = rhs.type;
    ...
    // Placement new because if type has changed, we have
    // not constructed the new type yet.
    if (type == SBV) new (&sbv) SparseBitVector<>(rhs.sbv);   // old sbv never destroyed
    ...
}
```

`SparseBitVector` owns a linked list of elements, so assigning into a set that already holds elements leaks all of them.

## Why it does not show today

The hot caller, `BVDataPTAImpl::alias`, constructs two fresh empty values per call and assigns into each exactly once through `expandFIObjs`. Placement new over an empty `SparseBitVector` loses nothing. Every assignment site in `svf/lib` assigns into a freshly constructed value, so the defect is latent.

It surfaces as soon as a caller reuses one buffer across a loop, which is a natural thing to write when avoiding a per-iteration allocation. That is how it was found, while working on `Andersen::getMayAliases`.

## The change

Both assignment operators destroy the current backing first, through the type currently stored, which may differ from the source's. The destructor's body is factored into `destroyBacking()` and reused. The move assignment operator also gains the self-assignment guard the copy operator already had; without it, self-assignment would now destroy the value before moving from it.

## Measurement

A standalone program assigns a 20,000 element set into one already populated target 20,000 times and reads `VmRSS` before and after:

| | Resident set growth |
|---|---|
| Before this change | 1,025,688 KB |
| After this change | 116 KB |

## Testing

- `wpa -ander` over the 114 basic C and C++ programs in Test-Suite: passes with and without the change.
- Abort counts for `wpa -fspta`, `-steens` and `-cfl` over `basic_c_tests` are identical either way, at 11, 31 and 62. Those are pre-existing and unrelated to this change.
- GCC 11 Release build with assertions on and `-Wall -Werror`: no warnings.

No behaviour change is expected or observed.

## Related, not changed here

`PointsTo::PointsToIterator`'s two assignment operators repeat the same placement-new-without-destroy pattern. The iterators appear not to own heap storage, so that is a correctness tidy-up rather than a leak, and it is left out to keep this change to the defect.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


yuleisui:
@mbarbar could you help review this one?

mbarbar (approved):
Thanks LGTM 👍 

I left a comment at some point, maybe when I wrote this class, to switch the union to std::variant once we use C++17 (and now we are on C++17). My first intuition looking at this code again is that this would've avoided the problem identified here but it'd need more thought. Something we should do in the future.

PR #1894, opened 2026-09-20
Accelerate Andersen getMayAliases with reverse points-to sets

cjsrxzdyzds:
### What

`Andersen::getMayAliases` can spend minutes repeatedly expanding points-to sets and confirming reverse-index candidates with full `mayAlias` calls. This follow-up to #1888 answers the same exact query directly from Andersen's reverse points-to sets. On the stable LLVM Values shared with upstream, query time improves by 4.78x to 43.64x across the measured modules.

### How

- Expand the query points-to set once and derive the raw objects whose expansion can intersect it.
- Visit those objects' reverse points-to sets, confirm each possibly stale posting with a current points-to bit test, and expand SCC representatives to their members.
- Accumulate representatives and answers in temporary dense bitsets. This makes arbitrary-order insertion constant time and emits the final sparse `NodeBS` in increasing ID order.
- Include the black-hole reverse set, so pointers that may target the black hole still alias the query.
- Keep exhaustive enumeration when reverse points-to sets are unavailable or the query itself may target the black hole. The fallback reuses the query expansion and expands each candidate once.

The two temporary dense sets use about `2 * node_bound / 8` bytes per query (approximately 78 KB for the measured 310,670-node tramp3d run).

### Validation

- Latest fetched, unmodified Test-Suite master (`8fdb0f07`): **2,575/2,575** registered tests pass, including **454** exact `getMayAliases` comparisons in 114 Andersen alias-test programs. No Test-Suite changes.
- GCC 11 Release build with assertions, `-Wall`, and `-Werror`: clean.
- A same-process exhaustive oracle compared the optimized result with per-node `mayAlias` for 28 tramp3d nodes that showed answer-count drift across independent runs: about **8.7 million comparisons, zero mismatches**.

### Performance

The driver selects queries by a fixed hash of stable LLVM Value identity, then joins independent runs by that identity rather than by run-specific SVF node IDs. Every variant used the same instrumentation, upstream base `1ac5af35`, a 2,000-Value cap, and a 900-second per-module query budget. Times sum only the stable Values completed by every variant.

| Module | Shared Values | Upstream | This PR | Speedup |
| --- | ---: | ---: | ---: | ---: |
| spec-gap | 213 | 0.001045 s | 0.000219 s | **4.78x** |
| redb-mt | 2,000 | 0.130263 s | 0.019220 s | **6.78x** |
| sqlite3 | 1,455 | 901.830667 s | 119.641196 s | **7.54x** |
| tramp3d-v4 | 267 | 910.210344 s | 20.856300 s | **43.64x** |

This PR also completed all 2,000 sampled sqlite3 Values in 167.0 seconds and all 2,000 tramp3d Values in 277.1 seconds. Upstream reached only 1,456 and 267 Values, respectively, before crossing the 900-second budget.

Recorded points-to and expanded-points-to sizes matched for every shared Value. Answer counts matched for spec-gap, redb-mt, and sqlite3. Tramp3d had 28 answer-count differences across independent analysis runs despite matching PAG size and NodeIDs; the same-process exhaustive check above is the correctness evidence for those nodes. This separation avoids treating cross-run state equality as an oracle or filtering timings by a post-run outcome.

AI assistance: Claude Code and OpenAI Codex.


svf/include/WPA/Andersen.h
```diff
@@ -357,9 +357,9 @@ class Andersen:  public AndersenBase
     /// SCC detection
     virtual NodeStack& SCCDetect();
 
-    /// Collect may-alias candidates. Attempts to do so from reverse points-to sets (cheaper),
-    /// and if not possible, falls back to trying every node (costlier).
-    NodeBS getMayAliasCandidates(NodeID node);
+    /// Answer a may-alias query from the reverse points-to sets. Returns false when
+    /// those sets cannot answer it, leaving the caller to try every node instead.
+    bool collectMayAliasesFromIndex(const PointsTo& expandedPts, NodeBS& aliases);
```
mbarbar: I'd like for us to avoid passing in the return value as much as possible. Would much prefer returning a `std::optional<NodeBS>` if we want to leave the fail case to the caller.

svf/lib/WPA/Andersen.cpp
```diff
@@ -917,74 +921,153 @@ void Andersen::updateNodeRepAndSubs(NodeID nodeId, NodeID newRepId)
...
- * Collect every SVFIR node that may alias the given node. Each candidate is confirmed
- * with mayAlias, so the result is exactly the nodes q for which mayAlias(node, q) holds.
+ * A set of node ids held as one bit per id.
+ *
+ * SparseBitVector keeps its non-empty words in a sorted linked list. Inserting
+ * ids from reverse points-to sets in arbitrary order repeatedly walks that list.
+ * This temporary dense form makes insertion constant time, then drains ids in
+ * increasing order so the final SparseBitVector is cheap to build.
+ */
+class DenseNodeSet
```
mbarbar: I believe just using `CoreBitVector` would suffice here.

svf/lib/WPA/Andersen.cpp
```diff
@@ -917,74 +921,153 @@ void Andersen::updateNodeRepAndSubs(NodeID nodeId, NodeID newRepId)
...
-        for (NodeID obj : objs)
-            entries += getRevPts(obj).size();
-        if (entries <= pag->getTotalNodeNum())
+        objects.set(object);
+        const NodeID base = pag->getBaseObjVarID(object);
+        objects.set(base);
+        if (isFieldInsensitive(base))
+            objects |= pag->getAllFieldsObjVars(base);
+    }
+    objects.set(pag->getBlackHoleNode());
```
mbarbar, line 987: Can this be moved to right after the declaration of `objects`

svf/lib/WPA/Andersen.cpp
```diff
@@ -917,74 +921,153 @@ void Andersen::updateNodeRepAndSubs(NodeID nodeId, NodeID newRepId)
...
+        for (NodeID subNode : sccSubNodes(representative))
+            answer.set(subNode);
+    });
+
+    // normalizePointsTo can remove field nodes while their ids remain in
+    // points-to sets. Do not return those stale nodes.
+    answer.forEach([&](NodeID id)
+    {
+        if (pag->hasGNode(id))
+            aliases.set(id);
```
mbarbar, line 1021: This loop can be merged in the one above without performance concern I believe and would allow us to do away with `answer`.

svf/include/WPA/Andersen.h
cjsrxzdyzds: Agreed. An empty `NodeBS` is a valid successful answer here, so `std::optional<NodeBS>` expresses the unavailable-index case more clearly than a Boolean plus output parameter. I will update the helper accordingly.

svf/lib/WPA/Andersen.cpp
cjsrxzdyzds: Agreed; I missed that `CoreBitVector` already provides the dense representation and ordered iteration needed here. I will replace the local wrapper with a `CoreBitVector` pre-sized to `(nodeBound + CoreBitVector::WordSize - 1) / CoreBitVector::WordSize` words, so arbitrary insertions do not repeatedly grow it.

svf/lib/WPA/Andersen.cpp
cjsrxzdyzds: Yes, I will move `nodeBound` immediately after `objects` is declared and use it when constructing the temporary bit vectors.

svf/lib/WPA/Andersen.cpp
cjsrxzdyzds: The second dense set is intended to preserve globally increasing insertion into the final `NodeBS`. Representatives are visited in order and each SCC's subnodes are ordered, but concatenating those per-SCC sequences is not guaranteed to be globally ordered; inserting that sequence directly can reintroduce the `SparseBitVector` traversal cost this change is avoiding. I will benchmark the merged form: if it is neutral I will simplify this as suggested; if it regresses the large cases I will keep the second `CoreBitVector` and make this rationale explicit in the comment.

cjsrxzdyzds: I benchmarked this after replacing the local wrapper with pre-sized `CoreBitVector`s. On stable 250-Value samples from sqlite3 and tramp3d, a same-process alternating A/B built both answers from the exact same `reps`; all 500 result sets were identical. Merging the loops made aggregate materialization 8-10% slower on sqlite3 and 2-3% slower on tramp3d. Repeated end-to-end pairs also favored the two-stage path. Empty and small answers do benefit from direct insertion, but their total cost is negligible; the large answers dominate and the concatenated SCC subnode sequences contain backward sparse-element transitions. I also tested an adaptive form that writes directly until the first backward transition and then restarts through the dense set; it was 1.0% slower on sqlite3 and 8.8% slower on tramp3d. I will therefore retain the second `CoreBitVector` and document the ordering rationale in the code.

yuleisui:
@cjsrxzdyzds @mbarbar I suspect there are other places we need to fix memory leaks wrt points-to data structure (e.g., standard, corebitvector and sparsebitvector)

mbarbar:
>@cjsrxzdyzds @mbarbar I suspect there are other places we need to fix memory leaks wrt points-to data structure (e.g., standard, corebitvector and sparsebitvector)

I checked those yesterday and they should be fine I believe.

(Wrong PR by the way.)

yuleisui:
> > @cjsrxzdyzds @mbarbar I suspect there are other places we need to fix memory leaks wrt points-to data structure (e.g., standard, corebitvector and sparsebitvector)
> 
> I checked those yesterday and they should be fine I believe.
> 
> (Wrong PR by the way.)

I see. That one has been merged. Thanks.

svf/lib/WPA/Andersen.cpp
yuleisui: Would be good to remove this duplicate data structure

cjsrxzdyzds: Agreed. Commit 7162b963 removes DenseNodeSet and uses pre-sized CoreBitVector for both temporary sets. I kept the second set for ordered insertion into the sparse result, following the measurements in the thread above. The GCC 11 assertions build with -Wall -Werror and all 2,575 Test-Suite tests pass.

cjsrxzdyzds:
I squashed the branch into a single commit, 190a1c41, so the history no longer adds `DenseNodeSet` in one commit and removes it in the next. The final tree differs from the previously reviewed 7162b963 by one change:

```diff
-            if (getPts(representative).test(object))
+            if (!representatives.test(representative) &&
+                    getPts(representative).test(object))
```

A representative that points to several objects in the query's set appears in several reverse sets. After the first posting confirms it, later postings cannot change the answer, so the points-to lookup is skipped.

**Measurement.** The driver selected 2,000 LLVM Values per module by a fixed hash of stable Value identity. For each Value, one process ran every variant in rotating order, and all results matched: 6,000 queries, zero mismatches. The table gives total query time over the 2,000 Values:

| Module (PAG nodes) | 7162b963 | This commit | Speedup |
| --- | ---: | ---: | ---: |
| sqlite3 (138,181) | 162.1 s | 118.5 s | **1.37x** |
| tramp3d-v4 (310,670) | 292.7 s | 229.4 s | **1.28x** |
| redb-mt (53,147) | 0.017 s | 0.017 s | 1.00x |

The same run also profiled the two temporary `CoreBitVector`s, which are allocated per query. Their fixed cost is about 2 µs per query on sqlite3 and 5 µs on tramp3d. Over every Value in the module, that is about 0.004% of the total query time. A variant that reuses scratch buffers across queries saved only 1–2% in total, so I did not pursue it. Replacing the dense sets with sorted vectors was 1.4–1.9x slower on large queries. The reverse-set walk takes 92–97% of the remaining time.

Validation: the Release build with assertions, `-Wall` and `-Werror` is clean, and all 2,575 Test-Suite tests pass on this tree.

yuleisui:
@mbarbar could you check again before I merge.

mbarbar:
Thanks LGTM

